PR #19, opened 2016-09-05
Don't truncate to 64k in mb_str_replace

Synchro:
I ran into this is a non-PhpSpreadsheet context, but using this code. The idea behind this function is that it's a multibyte-aware version of `str_replace`, but unlike that function, it imposes a 64k limit on strings, and it will unexpectedly truncate when handling strings longer than that. All that's required to fix it is remove the length limit by setting it to `null` instead, which is all this PR does. In case this limit was imposed for memory reasons, it could perhaps instead be raised to a larger value, but that would still break technical compatibility with str_replace.


Synchro:
The build failure is a code coverage timout in Clover, don't think it's to do with this change.

PowerKiKi:
This line goes all the way back to the very first commit of this project back in 2010. So it's hard to know the reason why the limit was hardcoded. But it seems safe to remove it. Thanks for the PR!

PR #29, opened 2016-10-18
Update composer.json

glensc:
fill description for recommend tags. the value is not dependency, but informational text:
https://getcomposer.org/doc/04-schema.md#suggest

originally submitted here: https://github.com/PHPOffice/PHPExcel/pull/544

but @PowerKiKi [pointed out](https://github.com/PHPOffice/PHPExcel/pull/544#issuecomment-254551315) this repo as well

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,8 +36,8 @@
         "friendsofphp/php-cs-fixer": "^1.11"
     },
     "suggest": {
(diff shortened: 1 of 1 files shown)


PowerKiKi:
I was too quick to dismiss it the first time, thank you for not letting it go that easily

PR #32, opened 2016-10-23
Remove non-obvious BaseReader's dependency

kamazee:
[BaseReader calls `isValidFormat`](https://github.com/kamazee/PhpSpreadsheet/blob/fb786e91b0177785ae7964be62ba45c73cdf8a1d/src/PhpSpreadsheet/Reader/BaseReader.php#L263) despite it doesn't require it to be implemented. A stub is added to mitigate it. Not really a good solution but an ad-hoc hack that preserves backward compatibility. Should be rethought and refactored, I guess.

### DIFF
--- a/src/PhpSpreadsheet/Reader/BaseReader.php
+++ b/src/PhpSpreadsheet/Reader/BaseReader.php
@@ -292,4 +292,19 @@ abstract class BaseReader implements IReader
     {
         return $this->securityScan(file_get_contents($filestream));
     }
+
+    /**
+     * Checks if the file to be read is of a valid format.
+     * Redefined in some readers that inherit from this one.
+     *
+     * It's kind of against OOP and it's only added here because base reader
+     * actually calls this method despite it might be not implemented in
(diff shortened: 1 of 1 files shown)


Quix0r:
I would make the method abstract (no body) and let the reader decide on its own rather than giving a true (which implies valid format which should NOT be the default).

kamazee:
@Quix0r I agree, and actually that was my initial intent. This would be a breaking change, though: any thingparty reader that extended `BaseReader` would become broken as it might not have an `isValidFormat` method — that's why I decided to return true by default which seems to be fully backwards compatible.

PowerKiKi:
Since PhpSpreadsheet does break many things, I think it would be a good time to implement @Quix0r suggestion. Would any of you submit a PR for that ?

Quix0r:
I think @kamazee should make a new one and this should be closed to keep things clear.

Quix0r:
Opps, only a merger (rebase -m).

PR #33, opened 2016-10-23
Require GD for saving images in xls

kamazee:
GD is implicitly required for saving images anyway, this exception when it's not loaded simply gives more information on what to do if the execution flow reaches a places that depends on a function from GD.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Xls.php
+++ b/src/PhpSpreadsheet/Writer/Xls.php
@@ -454,6 +454,9 @@ class Xls extends BaseWriter implements IWriter
         // the BSE's (all the images)
         foreach ($this->spreadsheet->getAllsheets() as $sheet) {
             foreach ($sheet->getDrawingCollection() as $drawing) {
+                if (!extension_loaded('gd')) {
+                    throw new \RuntimeException('Saving images in xls requires gd extension');
+                }
                 if ($drawing instanceof \PhpOffice\PhpSpreadsheet\Worksheet\Drawing) {
                     $filename = $drawing->getPath();
 


kamazee:
@Quix0r at a glance, GD is not really required by the whole class -- which is why I put the check specifically where its functions are used.

Better long-term alternative would be to use an image manipulation library, e.g. Imagine, to concentrate on saving sheet instead of dealing with pictures.

Quix0r:
Okay. :)

Progi1984:
@PowerKiKi Don't you think that can be put in composer.json require section ?

kamazee:
@Progi1984 it's listed there as the suggestion.

PowerKiKi:
@Progi1984 I tend to agree with @kamazee, that we shouldn't force dependencies for people who don't use images at all. But at the same time it would be easier for us to handle. I am willing to hear more opinions about it and maybe revisit this decision later.

Quix0r:
Currently the `extension_loaded()` call is done on every single image. This is not a so good design (I was not aware of that when I made my first review). In my view the check should be (at least) moved out of the loop as it does return on all iterations the same value, making more (useless) calls.

I would do this policy:

- only do things in a loop if it a) depends on the iterated item or b) if it differs in each iteration (which may imply a) )
- if it returns a "static" (same result each call) value, do it outside any iterations (as no change will ever happen)

kamazee:
@Quix0r I tend to agree. It's possible to move it out of the inner loop and perform a check if a collection of images is not empty. I'll take another look when I have a minute.

PR #34, opened 2016-10-23
Make detecting HTML a bit more strict

kamazee:
This is a naive attempt to fix PHPOffice/PHPExcel#806.
The whole thing seems to be fragile as supporting invalid use cases (improperly named files) has been favored over supporting valid use cases (treating files according to extension of a file name); however, the fix seems to address the specific issue I stumbled upon and might be sufficient for simple cases until another massive refactoring touches this area.


PowerKiKi:
thanks !

PR #35, opened 2016-10-25
Avoid array_key_exists()

Quix0r:
Compared to isset() it is much slower (there are tons of benchmarks out there) and should be avoided where posible (due to the slow performance). Only if you have keys that have NULL as value and you need to check if they are there (with NULL value) then and only then you need array_key_exists().


Quix0r:
Thank you. :-)

Issue #41, opened 2016-11-15
invalid value for $this->cellXfCollection[$pIndex];


kampfq:
in Spreadsheet class there will be a "null" as argument "$pIndex" when applying styles.

this causes a null return value when executing $this->cellXfCollection[$pIndex] on line 900;

PowerKiKi:
Please provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) of code that exhibits this issue. And don't forget to mention what version of this project you are using.

oleibman:
According to the linked PR(?), the problem was in PhpSpreadsheet::getCellXfByIndex. In the current code, the parameter for that function is now typed as `int`. According to Php's coercion rules for non-strict, null will be coerced to 0 in this situation. So the problem is now resolved. Removing stale and help wanted labels, but leaving issue closed.

PR #42, opened 2016-11-15
Fix for #41

kampfq:
fix error with null as argument


PowerKiKi:
This PR does not adhere to our coding style. Also it would probably be best to identify why the parameter is `null` in the first place and fix the cause, rather than the symptom. Maybe there is a more serious bug hidden behind that `null`.

Closing for now, but feel free to re-open if you have a better suggestion.

PR #46, opened 2016-11-24
Calculate column auto width for merged cells (if horizontal, 1-column…

tomaszsita:
… wide merge only)

This is to fix the situation when cells merged vertically are ignored in columns auto size calculations.

I also removed $calculateMergeCells param as it wasn't used and I believe the only situation when you want to calculate the width of merged columns has been covered here (and there's no reason to not calculate it)


PowerKiKi:
Thank you !

If possible, next time try to include unit tests.

PR #47, opened 2016-11-30
Fix CS and use use statement for consistance

ad3n:
Fix CS and use use statement for consistance


ad3n:
This job is not finish yet, i will finish after you accept this proposal :-)

PowerKiKi:
This won't be merged because it goes against our CS as [mentioned here](https://github.com/PHPOffice/PhpSpreadsheet#want-to-contribute) and clearly defined in `.php_cs`.

If you feel like a rule should be added and/or changed, it's probably best to talk about it first. And it would be best for everyone if it can be enforced by PHP-CS-Fixer.

PR #54, opened 2016-12-14
Excel 2003 XML working pattern property

Arthur-min:
Corrected the pattern property value.


PowerKiKi:
Thank you again for rebasing ! however it was already merged a few minutes ago in https://github.com/PHPOffice/PhpSpreadsheet/commit/ab7aa68300057032467370a822776c7de4568f18

Arthur-min:
I made a mistake on the previous PR, I fix it here with strtolower() ;)

Issue #53, opened 2016-12-13
Does not work drawing


johndoejdg:
Image  do not appear in spreadsheet after this code but in PhpExcel this works.
```
        $signaturePath = \Yii::$app->basePath.'/storage/for_documents/sign.png';
        $objDrawing = new Drawing();    //create object for Worksheet drawing
        $objDrawing->setName('Sign');
        $objDrawing->setDescription('Sign');
        $objDrawing->setPath($signaturePath);
        $objDrawing->setWidth(60);
        $objDrawing->setWorksheet($this->sheet);
        
        $objWriter = IOFactory::createWriter($saleDocumentInExcel, "Excel5");
```

PowerKiKi:
I confirm that the following code wil output an xlsx with an image, but an xls without an image (assuming a `test.jpg` beside the script):

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$spreadsheet = new PhpOffice\PhpSpreadsheet\Spreadsheet();

$drawing = new PhpOffice\PhpSpreadsheet\Worksheet\Drawing(); 
$drawing->setPath('test.jpg');
$drawing->setWorksheet($spreadsheet->getActiveSheet());

$writer = new PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet);
$writer->save('test.xlsx');

$writer = new PhpOffice\PhpSpreadsheet\Writer\Xls($spreadsheet);
$writer->save('test.xls');
```

Unfortunately I won't have time to investigate further. But if you could determine exactly what commit broke the feature, it would be very helpful.

johndoejdg:
I found strange place when debug this problem:
![image](https://cloud.githubusercontent.com/assets/2064465/21461918/f2371f92-c95e-11e6-8fd7-680c65c0c3a0.png)

Switch never execute any switch block because expected that class will start with "\\" in but class name start from latin character. I guess it can be reason of this issue but not check it.

johndoejdg:
I check it, drawing now attached.
Submit PR with fix: https://github.com/PHPOffice/PhpSpreadsheet/pull/61

atiqShafi:
My need is same, I have tried from view with html img tag but it does not worked. then I have tried below methods :
 
    public static function beforeSheet(BeforeSheet $event) {
        
        $drawing = new Drawing();
        $drawing->setPath(public_path('sent_files/35/100_1521622088/35_final_card.png'));
[...]

PR #61, opened 2016-12-23
Fix attaching drawing for xls

johndoejdg:
Fix according to this issue: https://github.com/PHPOffice/PhpSpreadsheet/issues/53


hienning:
What about to use ::class notation instead of writing namespace in the string? Php will always checks the namespace path for correctness and always returns the correct path, while a namespace writing in string will not.

johndoejdg:
@hienning: good point. Updated PR according you suggestion.

PowerKiKi:
Thank you ! I squashed, reverted minor code style mistakes and merged it.

PR #66, opened 2017-01-04
Properly format non-integer font sizes

c960657:
The OOXML spec defines the font sizes (the `sz` element) as a double, so the value needs to be wrapped in `PHPExcel_Shared_String::FormatNumber()` to avoid breaking in non-English locales.


PowerKiKi:
Thank you for taking the time to re-submit your PR here. I confirmed that the failing Travis is unrelated to this change, so I merged it as is.

PR #69, opened 2017-01-09
Use 'use' sections instead of fully qualified class names

odbc:
Propose use 'use' sections instead of fully qualified class names


PowerKiKi:
Thanks for the contribution, please fix the [code style issues](https://travis-ci.org/PHPOffice/PhpSpreadsheet/jobs/190129743#L197) and I'll merge it. You can do that via `./vendor/bin/php-cs-fixer fix`

PowerKiKi:
Thanks !

Quix0r:
Nice work. I would also recommend this way.

PR #73, opened 2017-01-13
Add failing test for getting settings globally affecting external xml reading

amcsi:
It appears that when `getLibXmlLoaderOptions()` is called, it globally affects external xml reading.
This is bad, because just reading excel files with this library has a global side effect, and it breaks Soap.

If you remove the row with `\PhpOffice\PhpSpreadsheet\Settings::getLibXmlLoaderOptions();`, you can see that the test passes

Otherwise in my PR I only added a failing test. I don't have a fix.


PowerKiKi:
While the use-case might be valid this PR won't be merged because it relies on a third party server (http://ec.europa.eu) and it doesn't fix anything but actually break tests.

But we can still keep the discussion going until a solution is found.

amcsi:
Hey, I just heard that PRs speak louder than issues

Not sure how to reproduce this without using a 3rd party URL though

PR #75, opened 2017-01-17
Add redis cache support

titanrat:
Add redis cache support using MemCache type as example


PowerKiKi:
Thanks for the contribution. You can fix the code style via `./vendor/bin/php-cs-fixer fix`.

However there are talk to refactor entirely the caching system in #3 (and also a [PR for Redis against PHPExcel](https://github.com/PHPOffice/PHPExcel/pull/772)). Please join in the discussion and make your opinion known whether PSR-16 support would be an interesting feature. I'll wait until we reach a agreement with @MarkBaker on the cache system future to merge or reject this PR...

titanrat:
Code style fixed. I'll try to join discussion later.

Quix0r:
I would like to see `use` (which is like `impot` in Java) instead of having static (full-qualified) class references everywhere:
````
use \Vendor\Project\Type\Foo;

$foo = new Foo();
````
Now if `\Vendor\Project\Type\Foo` has changed to e.g. `\Vendor\Project\Type\Fooier\Foo` then you have to update all references, with `use` only in the header in each file.

titanrat:
@Quix0r You want me to fix that in my new file CachedObjectStorage\Redis.php ?

Quix0r:
@titanrat I'm no dev here, maybe @PowerKiKi or @MarkBaker can say something about it?

titanrat:
@Quix0r fixed. But only in file which i created.

Quix0r:
@titanrat fine with me. :-) And maybe the people here won't like your PR then.

titanrat:
@Quix0r Let's see what they say.

PowerKiKi:
@Quix0r like I [said earlier](https://github.com/PHPOffice/PhpSpreadsheet/pull/75#issuecomment-273227658), I won't merge this until we get a decision in #3 about moving to [PSR-16](http://www.php-fig.org/psr/psr-16/). There are chances that this PR will become useless.

PowerKiKi:
This made obsolete by the introduction of PSR-16, thanks for the contribution anyway.

PR #80, opened 2017-01-26
Set column data type

strati72:
You can set column data type before import it.

How to use it:

```php
$objReader = new PHPExcel_Reader_CSV();
$objReader->setColDataType($columnLetter,PHPExcel_Cell_DataType::TYPE_STRING);
$objPHPExcel = $objReader->load("sample.csv");
````

### DIFF
--- a/src/PhpSpreadsheet/Reader/Csv.php
+++ b/src/PhpSpreadsheet/Reader/Csv.php
@@ -71,6 +71,14 @@ class Csv extends BaseReader implements IReader
     private $contiguousRow = -1;
 
     /**
+	 * Col Data Type 
+	 *
+	 * @access private
+	 * @array PHPExcel_Cell_DataType
+	*/
+	private $_colDataType = array();
+
+    /**
      * Create a new CSV Reader instance.
      */
     public function __construct()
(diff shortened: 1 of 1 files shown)


titanrat:
You can fix the code style via ./vendor/bin/php-cs-fixer fix.

strati72:
Done! All works

PowerKiKi:
Could you please explain why you need this feature ? What is the use case ? What kind of problem does it solve ?

I couldn't find this information in the commit message, not in the discussion here.

strati72:
Example1
ID custoner1 = "05724"
ID customer2 = "5724"
Example 2
Telephon number customer1 = "0166156423"
Telephon number customer2 = "166156423"
Example 3
Italian code flour type: "0" 
Italiani code flour type "00" 

In these cases, the zero (or zeros) initials are important. Without specifying the column type, the zeros will be lost. 
Example 2: in the second case, I get confused a free phone number with a overly expensive
Example 3:, an expensive get confused flour with an economic

PowerKiKi:
I had another look at this PR and see a few issues with it:

- there should not be any `_` prefix for variables/properties
- "column" should not be abbreviated "col"
- the default value of setColDataType is something that does not exist at all
- there is no unit tests

But more importantly I don't think it's the proper way to solve this use-case because it should be possible to do exactly what you want by using the more flexible `IValueBinder`. Have a look at [this](http://phpspreadsheet.readthedocs.io/en/develop/#using-value-binders-to-facilitate-data-entry) and [that](http://phpspreadsheet.readthedocs.io/en/develop/topics/reading-files/#a-brief-word-about-the-advanced-value-binder).

Issue #81, opened 2017-01-31
Avoid using slow in_array() when no need for NULL values


Quix0r:
The PHP function `in_array()` is slow in performance compared to `isset()`. This has the fundamental reason that `in_array()` loops through the whole array an compares each key. `isset()` in contrast is only checking if the given key is set but it doesn't find `NULL` values because it cannot do this (see documentation of each functions).

I would go and rewrite PhpSpreadsheet so it uses isset() as I see no use for checking NULL values e.g. with `$filterTypes` in `src/PhpSpreadsheet/Worksheet/AutoFilter/Column.php`. Also e.g. `src/PhpSpreadsheet/Writer/Xls/Worksheet.php` uses it but you can do it otherwise:

````
if (!isset($lockingArray[$someLock])) {
    // Do some stuff
    // Stuff is done
    $lockingArray[$someLock] = TRUE;
}
````

MarkBaker:
[in_array()](http://www.php.net/manual/en/function.in-array.php) doesn't check keys at all, it checks values in the array; so it performs a totally different function to [isset()](http://www.php.net/manual/en/function.isset.php)

Quix0r:
Yes, sorry. That is what i mean. :-) I will make a PR where `in_array()` can be easily replace by isset() without breaking anything.

PowerKiKi:
Closing after #82

PR #82, opened 2017-01-31
Rewrites: in_array() versus isset()

Quix0r:
As mentioned in #81 in_array() performs slow as it searches the whole array for a value. If you can use isset() because the value is unimportant, only the fact that `$foo` is in `$foos`, then you should use `isset()`. If you need to check on NULL values, use `array_key_exists()` (which we don't need/do here).

So isset() is IMHO really fine and better here.

### DIFF
--- a/src/PhpSpreadsheet/CachedObjectStorageFactory.php
+++ b/src/PhpSpreadsheet/CachedObjectStorageFactory.php
@@ -58,17 +58,17 @@ class CachedObjectStorageFactory
      * @var string[]
      */
     private static $storageMethods = [
-        self::CACHE_IN_MEMORY,
-        self::CACHE_IN_MEMORY_GZIP,
-        self::CACHE_IN_MEMORY_SERIALIZED,
-        self::CACHE_IGBINARY,
-        self::CACHE_TO_PHPTEMP,
-        self::CACHE_TO_DISCISAM,
-        self::CACHE_TO_APC,
-        self::CACHE_TO_MEMCACHE,
-        self::CACHE_TO_WINCACHE,
-        self::CACHE_TO_SQLITE,
-        self::CACHE_TO_SQLITE3,
+        self::CACHE_IN_MEMORY => true,
+        self::CACHE_IN_MEMORY_GZIP => true,
+        self::CACHE_IN_MEMORY_SERIALIZED => true,
+        self::CACHE_IGBINARY => true,
+        self::CACHE_TO_PHPTEMP => true,
+        self::CACHE_TO_DISCISAM => true,
+        self::CACHE_TO_APC => true,
+        self::CACHE_TO_MEMCACHE => true,
+        self::CACHE_TO_WINCACHE => true,
+        self::CACHE_TO_SQLITE => true,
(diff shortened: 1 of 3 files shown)


src/PhpSpreadsheet/CachedObjectStorageFactory.php
```diff
@@ -136,7 +136,7 @@ public static function getCacheStorageClass()
      **/
     public static function getAllCacheStorageMethods()
     {
-        return self::$storageMethods;
+        return array_keys(self::$storageMethods);
```
Quix0r: This returns the same array as expected (no change of "public method behaviour").

src/PhpSpreadsheet/Writer/Xls/Worksheet.php
```diff
@@ -518,8 +518,10 @@ public function close()
                 foreach ($conditionalStyles as $conditional) {
                     if ($conditional->getConditionType() == \PhpOffice\PhpSpreadsheet\Style\Conditional::CONDITION_EXPRESSION
                         || $conditional->getConditionType() == \PhpOffice\PhpSpreadsheet\Style\Conditional::CONDITION_CELLIS) {
-                        if (!in_array($conditional->getHashCode(), $arrConditional)) {
-                            $arrConditional[] = $conditional->getHashCode();
+                        if (!isset($arrConditional[$conditional->getHashCode()])) {
+                            // This hash code has been handled
+                            $arrConditional[$conditional->getHashCode()] = true;
```
Quix0r: This (old) code should make sure that $condinal is only written once, as I understand the code.

Quix0r:
Hmm a lot merges against upstream/develop, maybe I should rebase (which will close this PR)?

PowerKiKi:
Yes you should rebase. And it will not close the PR, but you may have to force push (which is fine for this).

Quix0r:
Waahh, tons of merge conflicts to resolve ... Any idea how to make it faster?

`git rebase -m upstream/develop`

I use this commonly.

PowerKiKi:
Since your patch is not that big, it may be faster to start from scratch.

But more importantly, did you benchmark those changes? At least the cache part seems to make no significant difference while making the code more complex to follow. I'll have to double check once I'm on my laptop, but I'm not sure I would merge it as is, unless you can show a significant speed improvement for normal-ish use-cases.

Quix0r:
Redone and force-pushed ...

PowerKiKi:
> No need to benchmark

In the case of `CachedObjectStorageFactory` you tried to increase speed of `initialize()`, at the cost of code readability of overall class, and a possible decrease in speed for `getAllCacheStorageMethods()` and `getCacheStorageMethods()`. Both of those methods are public and it is hard to predict how they are used in the wild.

So it seems you tried to improve 1 thing by possibly worsening 3 things, without a strong argument in favor of this change. To me it doesn't look like an improvement. Especially because the name `initialize()` suggest that it should probably not be used very often, hence does not have a significative impact in the first place.

For those reasons, I only merged the two other classes.

Quix0r:
I may come back with more changes where `isset()` can be safely used.

PR #84, opened 2017-02-01
fix redunant cell

titanrat:
extendRowsForChartsAndImages function in Html writer creates redunant cell in the end of table. Fixed it


PowerKiKi:
I think it would be more readable to return `$html` early instead of nesting the whole block inside the `if`.

Also could you add unit tests to prevent regression?

titanrat:
Fixed nested if.
Unfortunately, I can't add the unit tests, I haven't enough experience in writing them.

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -596,6 +596,12 @@ private function extendRowsForChartsAndImages(\PhpOffice\PhpSpreadsheet\Workshee
         }
         $html = '';
         ++$colMax;
+
+        //Return $html if $rowMax didn't change
+        if ($row !== $rowMax) {
+            return $html;
```
Quix0r: `$html` didn't change between initialization (line 597) and first usage. Maybe return hard-coded empty string instead?

titanrat: fixed

Quix0r: Okay, I then should not mark so quickly for approval. Better reviewing first. :-)

PowerKiKi:
Thank you for the contribution.

As a sidenote you may want to edit your git settings to use the same email as the one you use in GitHub so the commit are correctly linked to your GitHub account ([read more here](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/)).

Issue #94, opened 2017-02-14
xlsm parsing issue: $imageData Null


alexmilde:
Parsing some xslm might end in an $imageData null Error
Reader/Xlsx.php (line 1480)
Proposed fix: wrap result in if statement.
I will fork and propose the fix.

PR #95, opened 2017-02-14
issue #94 fix xslm parsing issue (null object for imageData)

alexmilde:


PowerKiKi:
I think it would be more readable to `continue;` early instead of nesting the whole block inside the `if`.

Also did you identify why the data was missing in the first place ?

alexmilde:
`$shape->xpath('//v:imagedata');`
seems to return null as far as i recall.

my testfile which caused the error is rather complex - so i cant really pin it down.

so you prefer:
`                                                if ($imageData) {
                                                  continue;
                                                }`

PowerKiKi:
yeap, `if ($imageData) { continue; }` [would be nice](http://softwareengineering.stackexchange.com/a/18459/85361) IMHO.

alexmilde:
well its opinion based, but i will change it ;)

PowerKiKi:
Much easier to review, thank you !

alexmilde:
your welcome. 
btw: this was my first PR ;) 
do you delete the branch or shall i?

PowerKiKi:
You should do it if you don't need anymore.

Issue #85, opened 2017-02-07
Ods reader skips subsequent spaces


agopaul:
Basically the reader class ignores the `<text:s>` tag, which is used when you have 2 or more consecutive spaces.

[See  the OpenDocument specs](http://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part1.html#element-text_s)

This is something that it's not implemented in the PHPExcel library either.

I worked on a [quick fix](https://github.com/redokun/PHPExcel/commit/a588c611bc5c51ab2dfa0019d124671ca3bc592e) (sorry for the code style change) which I'm still testing, but it is more of a dirty patch ATM, since the original implementation uses SimpleXML which to my understanding is insufficient for the purpose because of a missing option to traverse the XML tree down to the Node level. You can't access to mixed contexts where you have both Node and Element, and text nodes are not counted as children by SimpleXML.

This means that the whole class should be rewritten to use another XML parsing lib, like DOMDocument.

If the maintainers are willing to consider this change, I can work on a PR which would include: writing at least a few basic unit tests for the Ods reader class, migrating to DOMDocument and finally implementing a parser that takes `<text:s>` tags into account.

Edit: grammar

PowerKiKi:
Your offer to help us much appreciated. Do you have argument in favor or against a migration to DOMDocument  in mind ? I am not particularly familiar with their differences and their availability/popularity...

agopaul:
I looked at the class more carefully and I realized that a complete rewrite is not necessary. The only part affected by the refactoring would be a part of the `loadIntoExisting` method.

DOMDocument and SimpleXML share the same underlying C library (libxml2) and it seems that [they have similar performance](http://stackoverflow.com/questions/9367069/simplexml-vs-domdocument-performance). Having the same dependency (libxml2), they are both bundled with PHP when complied with the libxml option (no PECL necessary).

The main point against SimpleXML is that it's insufficient for reading the XML tree because it ignores the text nodes and therefore it cannot be used to parse XML parts with mixed text nodes and Elements.

See the comparison below:

**SimpleXMLElement**

    <?php  
    $xml = '<?xml version="1.0" encoding="UTF-8"?>
    <test>There are <s/>multiple spaces here <s/></test>';
    
    $s = new SimpleXMLElement($xml);
    
    foreach ($s as $item) {
    	/** @var $item \SimpleXMLElement */
    	var_dump($item->getName());
    }

    /*
     * Output:
     *
[...]

PR #98, opened 2017-02-20
Ods reader: various fixes and unit tests added (bug #85)

agopaul:
While the PR was intended to fix only a problem with the parsing of subsequent spaces (https://github.com/PHPOffice/PhpSpreadsheet/issues/85), I actually ended up fixing a few more bugs while rewriting the content parsing logic to use DOMDocument.

Fixes and changes made:

- Added a few basic unit tests for the Ods reader
- The parsing lib for the content parser is now DOMDocument rather than SimpleXML
- Subsequent spaces are now parsed correctly. See the initial bug report https://github.com/PHPOffice/PhpSpreadsheet/issues/85
- The sheet count is now correct. Before this change an empty sheet was added at the end of the Worksheet collection.
- Fixed the hyperlinks parsing
- Fixed parsing of rich text. Strings containing styles like bold and italic inside (eg. "This **is** a *string*") were badly truncated. Now all the text is parsed, discarding the inline style. I don't know if this lib should parse the inline styles too or discard them, but now at least the content is preserved. Maybe a future PR will enable the reader to parse the inline font styles.
- Lastly, I fixed a few docblocks

As a side note: I can't run the whole test suite for some reason. It hangs after about 88% or so. Should I exclude some test group or run phpunit with additional flags/configuration?


PowerKiKi:
Great contribution, thank you very much !

PR #100, opened 2017-02-22
Decode comment text in xls reader (#99)

### LINKED ISSUE #99: Comments in xls format should be decoded
Currently, xls reader does not decode comment properly.

The [attachment](https://github.com/PHPOffice/PhpSpreadsheet/files/792964/unicode_comment.zip) is a xls file with two comments in A1 and B1.

Reader:
```php
require_once __DIR__.'/src/Bootstrap.php';

$filename = './unicode_comment.xls';
$spreadsheet = \PhpOffice\PhpSpreadsheet\IOFactory::load($filename);
$notes = $worksheet->getComments();
foreach ($notes as $key => $note) {
	$text = $note->getText()->getPlainText();
	echo "$key: $text\n";
}
```
Output
```
A1: ROSA:
SU ACOMPANTE ES OSCAR PUCH Y NO LO TENEMOS REFLEJADO
B1: `O
```

Expected Output
```
A1: ROSA:
SU ACOMPAÑANTE ES OSCAR PUCH Y NO LO TENEMOS REFLEJADO
B1: 你好
```

I would make a pull request for this.

ciffel:
It would resolve #99.

References:
* [The Microsoft Excel File Format](https://www.openoffice.org/sc/excelfileformat.pdf): From BIFF8 on, strings are always stored using UTF-16LE text encoding. The character array is a sequence of 16-bit values. Additionally it is possible to use a compressed format, which omits the high bytes of all characters, if they are all zero
* [Apache POI](http://grepcode.com/file/repo1.maven.org/maven2/org.apache.poi/poi/3.7/org/apache/poi/util/StringUtil.java#StringUtil.readUnicodeString%28org.apache.poi.util.LittleEndianInput%29): is16BitFlag

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xls.php
+++ b/src/PhpSpreadsheet/Reader/Xls.php
@@ -1629,8 +1629,20 @@ class Xls extends BaseReader implements IReader
         $cbRuns = self::getInt2d($recordData, 12);
         $text = $this->getSplicedRecordData();
 
+        $textByte = $text["spliceOffsets"][1] - $text["spliceOffsets"][0] - 1;
+        $textStr = substr($text["recordData"], $text["spliceOffsets"][0] + 1, $textByte);
+        // get 1 byte
+        $is16Bit = ord($text["recordData"][0]);
+        // Additionally it is possible to use a compressed format,
+        // which omits the high bytes of all characters, if they are all zero
+        if (($is16Bit & 0x01) === 0) {
+            $textStr = \PhpOffice\PhpSpreadsheet\Shared\StringHelper::ConvertEncoding($textStr, 'UTF-8', 'ISO-8859-1');
+        } else {
+            $textStr = $this->decodeCodepage($textStr);
+        }
+
         $this->textObjects[$this->textObjRef] = [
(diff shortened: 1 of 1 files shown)


PowerKiKi:
Thanks for the contribution, the code style can be fixed via `./vendor/bin/php-cs-fixer fix` (or manually if you prefer). Also would you be able to include a unit tests covering this issue, preferably without depending on a pre-existing .xls file ?

ciffel:
XLS is binary format. Should I put the xls file in tests/data/Reader/XLS? Do you have any suggestion about how to create testing for binary format .

PowerKiKi:
I don't think we have any test do that yet, but I was thinking writing a temp file with PhpSpreadsheet and reading it. That way that avoids cluttering the repository with to many binaries, that would kinda test read and write and hopefully that should still be fast enough. What do you think?

ciffel:
The code style is fixed.

Using a temp file which is read and written with PhpSpreadsheet to create a testing is a great idea. Unfortunately, comment writer for xls format has not implemented yet.

PowerKiKi:
I guess I'll merge it as is for now. Thanks for the contribution !

If you ever need to write comments, another PR would be welcome ;-)

Issue #96, opened 2017-02-17
PHP 7.1 compatibility


PowerKiKi:
There are currently [5 failing tests](https://travis-ci.org/PHPOffice/PhpSpreadsheet/jobs/202110926#L277) that prevent us to declare PHP 7.1 compatibility. Any help to solve those would be very much appreciated.

ankitm123:
I see no one is assigned yet, I would like to jump on it, and see if I can help.

PowerKiKi:
Any help with this is welcome, thank you for offering your help. You could submit separate PR for each tests or all at once, up to you.

ankitm123:
3 issues were fixed, but the remaining 2 issues are due to the dompdf package. I changed composer.json file to require 0.8.0v for Dompdf, it still gives the same issue. I will look into it a bit more.
Check this link: https://github.com/dompdf/dompdf/releases/tag/v0.8.0

ankitm123:
Updating to v0.8.0 for dompdf does fix the remaining issues. Take a look at the [passing tests](https://travis-ci.org/ankitm123/PhpSpreadsheet/jobs/204905445) for php 7.1.Now that the build completed for php 7.1, Do we need to keep php 7.1 in the allowed failure list? Guess not.

PR #102, opened 2017-02-23
PHP7.1 compatibility for issue #96

ankitm123:
Fix for PHP 7.1 compatibility #96. Typecasting the variables fixed the issue. Refer to this page for more details:
https://wiki.php.net/rfc/invalid_strings_in_arithmetic


.travis.yml
```diff
@@ -23,7 +22,8 @@ before_script:
   # Deactivate xdebug if we don't do code coverage
   - if [ -z "$COVERAGE" ]; then rm -rfv /home/travis/.phpenv/versions/$(phpenv version-name)/etc/conf.d/xdebug.ini ; fi
   - composer install
-
+  - composer update
```
PowerKiKi: This should not be here, instead composer.lock file should be included in this PR.

test
```diff
@@ -0,0 +1 @@
+running composer update after composer install to fetch v0.8.0
```
PowerKiKi: This file should not be here at all

ankitm123:
@PowerKiKi, I added the composer.lock file, and removed the other unnecessary code. Thanks for the suggestion.

PowerKiKi:
Thanks !

PR #109, opened 2017-03-05
Simplify mbStrReplace()

c960657:
By design, UTF-8 allows any byte-oriented substring searching algorithm, since the sequence of bytes for a character cannot occur anywhere else ([source](https://en.wikipedia.org/wiki/UTF-8#Advantages_3)).

So `str_replace()` also works for UTF-8-encoded strings, assuming that the input strings are valid UTF-8 strings. AFAICT the previous implementation of mbStrReplace() did nothing to detect invalid strings.

Also, `str_replace()` does not support [Unicode equivalence](https://en.wikipedia.org/wiki/Unicode_equivalence), but nor do the other `mb_string` functions, and nor does `=SUBSTITUTE()` in Excel (tested on Excel for Mac version 15.19.1).

### DIFF
--- a/src/PhpSpreadsheet/Shared/StringHelper.php
+++ b/src/PhpSpreadsheet/Shared/StringHelper.php
@@ -553,9 +553,6 @@ class StringHelper
     /**
      * Replace into multi-bytes string.
      *
-     * Strangely, PHP doesn't have a mb_str_replace multibyte function
-     * As we'll only ever use this function with UTF-8 characters, we can simply "hard-code" the character set
-     *
      * @param string|string[] $search
      * @param string|string[] $replace
      * @param string $subject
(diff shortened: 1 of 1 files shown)


PowerKiKi:
Could you let me know how you tested it in Excel, so I can reproduce on latest Excel on Windows, please ?

c960657:
First I used this PHP code, `var_dump("\u{0061}\u{030A}", "\u{00E5}")`, to generate the strings `å` and `å` (I'm not sure if the two variants survive Github's formatting). The first letter is composed of an `a` with a combing ring above ` ̊`, and the last is the same letter represented as a single character.

I copy-pasted the two letters into a spreadsheet and then used the formula `=SUBSTITUTE(A1;"å";"X")` (containing the single-character variant) to see if each variant matched. In my experiment, only one of them matched, even though the two letters looked identical to the naked eye.

PowerKiKi:
Thanks ! I was able to reproduce and confirm everything you said. I merged with even further simplification and added unit tests, but still crediting you.

Issue #115, opened 2017-03-14
Ability to add value to cell instead of just setting


murilozilli:
Today I saw that the project only has setValue methods(or setCellValue for Worksheet class). I think it would help to have addValue methods as well, giving the chance for the user to break the line inside the cell if he wishes.

That would help users that want to put values inside the same cell on a loop for instance.

PowerKiKi:
Closing for the reasons explained in https://github.com/PHPOffice/PhpSpreadsheet/pull/116#issuecomment-286624055

PR #116, opened 2017-03-14
Feature 115

murilozilli:
Added methods on PhpSpreadsheet/Cell and PhpSpreadsheet/Worksheet to add value to a cell instead of just setting it. Making it easier for users to add value to a cell inside loops for instance.

#115

### DIFF
--- a/src/PhpSpreadsheet/Cell.php
+++ b/src/PhpSpreadsheet/Cell.php
@@ -215,8 +215,6 @@ class Cell
      * @param mixed $pValue Value
      * @param bool $breakLine Breaks the line(true) before concatenating value or not(false, default)
      *
-     * @throws Exception
-     *
      * @return Cell
      */
     public function addValue($pValue = null, $breakLine = false)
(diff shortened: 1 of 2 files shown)


PowerKiKi:
I am not confortable with merging this for several reasons.

First there is no unit tests. From now on all new features should be unit tested.

Second this code only consider the case of string data. What happens if it's a number ? or a date ? or anything else ?

And finally, and more important of all, accepting this PR would create a precedent that value manipulation/formatting can be done within PhpSpreadsheet itself. And this is IMHO a bad idea, there is no way we can support and maintain all use-cases for value manipulation (appending, prepending, subtracting, adding, multiplying, dividing, etc.) for all data types (string, number, boolean, null, formula, etc.).

This kind of very specific value manipulation is best kept the user's code, not this PhpSpreadsheet.

murilozilli:
@PowerKiKi Thanks for the feedback. I actually just used this code and thought why the heck not give it back to all. For reasons 1 and 2 I could still improve the code if it was something you guys would want but after reading reason 3 I totally agree with you.

Issue #17, opened 2016-08-13
Update all examples to use new structure


MarkBaker:
Fix all example code to use new bootstrap, namespacing and correct classes where changes have been made internally

MarkBaker:
Fixed typehinted arguments where null values being passed to methods, allowing null as defaults

PowerKiKi:
I'd like to help you with that. If not completed already I could finish the example updates, and I had in mind to use phpunit to make sure that all examples can actually be run (no actually testing the results, but at least that they run without errors). What do you think ? Or would you rather finish this by yourself ?

Progi1984:
@PowerKiKi You can use the template HTML/JS/CSS that I made for PhpPresentation for samples.

https://github.com/PHPOffice/PHPPresentation/blob/develop/samples/Sample_Header.php#L419
https://github.com/PHPOffice/PHPPresentation/blob/develop/samples/Sample_Footer.php

PowerKiKi:
Forgot to mention that I am working on that. Using PHPPresentation as foundation.

Progi1984:
@PowerKiKi Perfect ;)

PowerKiKi:
Finally pushed my work. `Examples/` was moved to `samples/` and mimic as much as possible what is done for PhpPresentation.

Compared to PhpPresentation, I moved some logic from Header.php to a dedicated class `\PhpSpreadsheet\Helper\Sample`. That seemed cleaner. And it is also used in unit tests to get the complete list of samples. Meaning that all samples are now ran by phpunit. Or almost all of them, because I noticed a few of them crashed. I'll open separate issue for those.

One last thing to do would be to move `Documentation/Examples` into the same structure, most likely keeping the folder structure and re-organizing existing sample in folders too. And finally reflect this folder structure in the menu on the web interface, to have some kind of categories.

PR #123, opened 2017-03-24
Fix sample which does't work

tsmsogn:


PowerKiKi:
Great, thanks !

Instead of `require`ing `Bootstrap.php`, we should actually require `samples/Header.php`, but that can be done later as part of #17.

tsmsogn:
Thank you for merging!

Issue #124, opened 2017-03-27
Update all 3rd-party modules to latest versions


Quix0r:
If I run `composer update` it gives me some updated packages:
````
$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Package operations: 1 install, 16 updates, 0 removals
  - Updating squizlabs/php_codesniffer (2.8.0 => 2.8.1) Loading from cache
  - Updating symfony/yaml (v3.2.4 => v3.2.6) Downloading: 100%
  - Updating phpunit/php-timer (1.0.8 => 1.0.9) Loading from cache
  - Updating sebastian/code-unit-reverse-lookup (1.0.0 => 1.0.1) Loading from cache
  - Updating phpunit/php-token-stream (1.4.10 => 1.4.11) Loading from cache
  - Updating phpunit/php-code-coverage (4.0.6 => 4.0.7) Loading from cache
  - Updating phpspec/prophecy (v1.6.2 => v1.7.0) Loading from cache
  - Updating phpunit/phpunit (5.7.14 => 5.7.17) Downloading: 100%
  - Updating symfony/stopwatch (v3.2.4 => v3.2.6) Downloading: 100%
  - Updating symfony/process (v3.2.4 => v3.2.6) Downloading: 100%
  - Installing symfony/polyfill-xml (v1.3.0) Loading from cache
  - Updating symfony/finder (v3.2.4 => v3.2.6) Downloading: 100%
  - Updating symfony/filesystem (v3.2.4 => v3.2.6) Downloading: 100%
  - Updating symfony/event-dispatcher (v3.2.4 => v3.2.6) Downloading: 100%
  - Updating symfony/debug (v3.2.4 => v3.2.6) Downloading: 100%
  - Updating symfony/console (v3.2.4 => v3.2.6) Downloading: 100%
  - Updating friendsofphp/php-cs-fixer (v2.1.0 => v2.1.2) Downloading: 100%
Writing lock file
Generating autoload files
$
````
Please consider updating them and let me know.

PowerKiKi:
I am closing this as it is not really helpful. It is not a bug report, nor a contribution for improvement. Moreover while we try to stay up to date, packages are bound to be out of date rather quickly.

Next time either provide a PR that directly solve your concern, or only submit a bug report that explain what in great details what is broken and why it should be fixed.

Quix0r:
[...]

PR #125, opened 2017-03-29
updated libraries

Quix0r:
As mentioned in #124 used libraries should be kept up-to-date, like as an engineer you try to repair a newer device with old tools (like older screw drivers which have changed recently) so you cannot repair it because you are running around with old tools in your hands.

Stay up-to-date with libraries you use quickly (don't wait until they are declared obsolete).

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -125,16 +125,16 @@
         },
         {
             "name": "friendsofphp/php-cs-fixer",
-            "version": "v2.1.0",
+            "version": "v2.1.2",
             "source": {
                 "type": "git",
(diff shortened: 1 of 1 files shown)


gmsantos:
Why `composer.lock` is not ignored after all? Ignoring it avoid this kind of issue.

PowerKiKi:
For all the reasons explained in https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/

To be clear, being out of sync of a few minor releases for dev dependencies is not an issue to begin with. If that were major versions​ for production dependencies I would be more willing to do something about it, but for now there are more urgent matters than updating ever changing deps for a project that itself was never even released once.

PR #128, opened 2017-03-31
Adding an empty cell condition

CrazyBite:
Adding an empty cell condition like this:
```
<conditionalFormatting sqref="A1:B1048576">
    <cfRule type="containsBlanks" dxfId="4" priority="3">
        <formula>LEN(TRIM(A1))=0</formula>
    </cfRule>
</conditionalFormatting>
```


PowerKiKi:
Thanks !

I merged even though Travis failed, because the failure was unrelated.

Issue #137, opened 2017-04-14
Throw exception for Excel5 with more than IV columns


simivar:
PHPSpreadsheet should throw an exception or return some information when there are columns over IV in spreadsheet. Right know XLS gets generated without any warning but with only first row with data to IV column. 

References issue [PHPOffice/phpexcel#792](https://github.com/PHPOffice/PHPExcel/issues/792). 

Code that should generate 2 rows with number from 0 to 255 but generates only one because of overflow:
```php
$numberOfColumnsFromAtoIZ = 260; // if you change it to 255 (number of columns from A to IV) it'll work
		
header('Content-Disposition: attachment; filename="test_over_iv.xls"');
header('Content-Type: application/vnd.ms-excel');
		
$phpExcel = new \PHPExcel();
$activeSheet = $phpExcel->getActiveSheet();
		
for($row = 1; $row < 3; $row++){
    for($column = 0; $column < $numberOfColumnsFromAtoIZ; $column++){
        $activeSheet->setCellValueByColumnAndRow($column, $row, $column);
    }
}
		
$excelWriter = new \PHPExcel_Writer_Excel5($phpExcel);
$excelWriter->save('php://output');
// saved an spreadsheet with only one row without raising any error
```

PR #138, opened 2017-04-14
Throw exception in XLS with columns or rows overflow

simivar:
#137


PowerKiKi:
Thanks !

Next time, if can you come up with a PR, you don't have to create an issue first and then a PR. Just a PR with a nice explanation like you did would be enough.

PR #139, opened 2017-04-14
Add more entries to gitattributes file

jdufresne:
Used by composer to ignore tests and other development files in the
distribution of a release.

For more details, see information in the article:

https://blog.madewithlove.be/post/gitattributes/

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,10 @@
-/tests export-ignore
-README.md export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
+/.php_cs.dist export-ignore
+/.travis.yml export-ignore
+/README.md export-ignore
+/composer.lock export-ignore
+/docs/ export-ignore
+/phpunit.xml.dist export-ignore
+/samples/ export-ignore
+/tests/ export-ignore


PowerKiKi:
This is a duplicate of https://github.com/PHPOffice/PHPExcel/pull/1083, and my opinion did not change from back then. That is, I am still torn between excluding docs, sample and tests or keeping them. To me they are an invaluable resources to understand a project from within the IDE itself.

And while the sample do take a significant part of the total weight, I still feel it is relatively resaonnable:
image

Composer itself [stopped recommending the practice](https://github.com/composer/composer/commit/bf328697e1ae7d290fb17df2e902a652ec4d9378) to exclude them. And interestingly enough the discussion went full circle when composer removed all mention of [light weight packaging](https://github.com/composer/composer/commit/b34220edc8ed738bce3f682248622976543d6ec4#commitcomment-2867313) and somebody refereed to an issue in PHPExcel. Said issue was fixed by excluding docs, but it was later reverted by Mark in https://github.com/PHPOffice/PhpSpreadsheet/commit/5efd01f35384226c057e2d5c799dda61f15f4e8f.

So, like before, I am closing this for now. But if @MarkBaker has a different opinion, or if the package growth goes out of hand in the future, the idea could be re-evaluated.

PowerKiKi:
I would be interested in seeing statistics of packages on composer that do exclude things. If you know where/how to get those, le me know...

PowerKiKi:
Actually I did it myself and scrapped the 1504 most popular composer package's `.gitattributes` files to see how it's used in the wild:

c

So we can see that `.gitattributes` is not used often to exclude things (21% all projects), but when it does, it's often to exclude tests (19%) and rarely to exclude docs or samples (cumulative 5%).

So it tends to confirm that excluding things is more an exception than a de-facto standard. And in our cases excluding tests would be only ~1.5M saved which seems quite insignificant, compared to our docs/samples of ~7M. But excluding docs/samples is even more the minority.

So I stand on my position to not exclude things for now.

For the record, I attached the raw `.gitattributes` files if you want to have a look for yourself: 
[gitattributes for popular composer packages.zip](https://github.com/PHPOffice/PhpSpreadsheet/files/923407/gitattributes.for.popular.composer.packages.zip)

jdufresne:
Thank you for the very informative response. I found that very helpful.

PR #142, opened 2017-04-20
Writer ignores PreCalculateFormulas Flag

azine:
The OpenDocument Writer does not respect the "preCalculateFormulas" flag that can be set.

With my proposed solution it would. But as I have only just used the PhpSpreadsheet for the first time, I'm not sure if this is the best way to fix it.

Please comment or merge.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Ods/Content.php
+++ b/src/PhpSpreadsheet/Writer/Ods/Content.php
@@ -224,9 +224,14 @@ class Content extends WriterPart
                     throw new Exception('Writing of error not implemented yet.');
                     break;
                 case DataType::TYPE_FORMULA:
-                    try {
-                        $formulaValue = $cell->getCalculatedValue();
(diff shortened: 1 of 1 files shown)


PowerKiKi:
It is not really obvious but `$this->getParentWriter()` is always an instance of `PhpOffice\PhpSpreadsheet\Writer\Ods`, so there is no need for `method_exists()` check (I'll make change type hinting to make that clearer).

Also it seems that `$formula_value` might be undefined in the `empty()` check. 

And finally it would be nice to have some unit tests for that. I suggested to add a case in `PhpOffice\PhpSpreadsheetTests\Writer\Ods\Content\ContentTest`.

PowerKiKi:
Since 870d8640f525bc282bfa7908c994483342f4f9a0 the writer parent is enforced to be an implementation so `method_exists()` must be removed.

Issue #144, opened 2017-04-27
Support for mPDF 7.x


KacerCZ:
Hi,
fix of issue https://github.com/PHPOffice/PhpSpreadsheet/issues/37 used class `\Mpdf\Mpdf`.
Later pull request https://github.com/PHPOffice/PhpSpreadsheet/pull/76 changed used class to `\Mpdf`.

mPDF up to version 6.1 uses class `\mPDF`. 
From version 7.0 was added namespace and main class was renamed to `\Mpdf\Mpdf`, see https://github.com/mpdf/mpdf/blob/development/CHANGELOG.md

Is would be great if both versions could be supported, because mPDF 6.1 will not get support for PHP 7.1.

PowerKiKi:
Thank you for the clarification. Clearly going back to non-namespaced code was a mistake. It should not have been merged that way.

Support for mPDF 6 must be dropped in favor of mPDF 7. Because mPDF 7 support all the same PHP versions as us. And we really don't have the resource to support two different versions.

Would you be able to submit a PR for that ?

PowerKiKi:
Thank you for the clarification. Clearly going back to non-namespaced code was a mistake. It should not have been merged that way.

Support for mPDF 6 must be dropped in favor of mPDF 7. Because mPDF 7 support all the same PHP versions as us. And we really don't have the resource to support two different versions.

Would you be able to submit a PR for that ?

KacerCZ:
PR with fix is here - https://github.com/PHPOffice/PhpSpreadsheet/pull/149

PR #149, opened 2017-05-01
Updated mPDF to version 7.0.

KacerCZ:
Fix for issue https://github.com/PHPOffice/PhpSpreadsheet/issues/144.

Stability of mPDF package is reduced to 'dev' because there is no stable release of version 7.0 yet.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Pdf/MPDF.php
+++ b/src/PhpSpreadsheet/Writer/Pdf/MPDF.php
@@ -73,7 +73,7 @@ class MPDF extends Core implements \PhpOffice\PhpSpreadsheet\Writer\IWriter
         }
 
         //  Create PDF
-        $pdf = new \mPDF();
+        $pdf = new \Mpdf\Mpdf();
         $ortmp = $orientation;
         $pdf->_setPageSize(strtoupper($paperSize), $ortmp);
         $pdf->DefOrientation = $orientation;
--- a/composer.json
+++ b/composer.json
@@ -32,7 +32,7 @@
         "psr/simple-cache": "^1.0"
     },
     "require-dev": {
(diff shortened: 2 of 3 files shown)


PowerKiKi:
I didn't realize it was only in beta. I am not so sure anymore this should be merged since support for unreleased software might bring bring up issue...

Any idea about the speed of development of mPDF and when a stable might be released ?

b1rdex:
Here is 7.0 release milestone: https://github.com/mpdf/mpdf/milestone/2

titanrat:
Mpdf 6.1 is fully not compabitable with php 7.1.
I think this is a problem and it's better to upgrade to Mpdf 7 faster.

PowerKiKi:
As soon as it is released, yes. But I don't think it fair to force PhpSpreadsheet users to use an unreleased version of Mpdf meanwhile

src/PhpSpreadsheet/Writer/Pdf/MPDF.php
```diff
@@ -73,7 +73,7 @@ public function save($pFilename)
         }
 
         //  Create PDF
```
titanrat: Doesn't work without this config.
        $config = ['tempDir' => 'tmp'];
        $pdf = new \Mpdf\Mpdf($config);

KacerCZ:
Release Candidate version of mPDF 7.0 was tagged - https://github.com/mpdf/mpdf/releases/tag/v7.0.0-RC1

PowerKiKi:
Closed via https://github.com/PHPOffice/PhpSpreadsheet/commit/360db8dbcdac658953ef4c6ad745c63bd6e68cf2, thanks for keeping me up to date.

PR #159, opened 2017-05-17
fix countifs definition

Remo:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [X] Code style is respected

The countifs function can have more than 2 parameters, it's defined as `COUNTIFS(criteria_range1, criteria1, [criteria_range2, criteria2]…)`, see here: https://support.office.com/en-us/article/COUNTIFS-function-dda3dc6e-f74e-4aee-88bc-aa8c2a866842

I'm not sure how you guys want to cover something like this in a unit test. Should I add a new file and test whether we can read it? Shall I modify an existing one?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Fix to getCell() method when cell reference includes a worksheet reference - @MarkBaker
 - Ignore inlineStr type if formula element exists - @ncrypthic [#570](https://github.com/PHPOffice/PHPExcel/issues/570)
 - Excel 2007 Reader freezes because of conditional formatting - @rentalhost [#575](https://github.com/PHPOffice/PHPExcel/issues/575)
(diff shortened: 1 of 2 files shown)


PowerKiKi:
It seems this is only the first part needed to support multiple criteria. We would need to modify `Statistical::COUNTIF()` method to support it, and cover that method with unit tests. Unit tests can most likely be heavily inspired from `MathTrigTest::testSUMIF()` (and `tests/data/Calculation/MathTrig/SUMIF.php`) since the methods have similar signatures.

Remo:
I wasn't aware that you're evaluating the result of that function too. It's not something I'll be able to look into anytime soon, I have no experience with the internals of PhpSpreadsheet, but I'll keep this on my long list.

PowerKiKi:
Calculation is an important part of PhpSpreadsheet too. But I understand it can be time consuming to add support for it...

Meanwhile I don't think this PR can be merged as is, because it would break calculation sooner or later. So I'll close it for now, but feel free to re-open with a more complete solution.

PR #165, opened 2017-05-28
PHPExcel#1110

MewesK:
Fixes the PHP 7.1 bug reported over at https://github.com/PHPOffice/PHPExcel/issues/1110
The behavior should not change since the default value used should behave the same way as the string value in earlier PHP versions.

Fixes a return type in the doc block of IReader.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,13 +32,15 @@
         "psr/simple-cache": "^1.0"
     },
     "require-dev": {
-        "mpdf/mpdf": "^6.1",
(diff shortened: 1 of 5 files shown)


PowerKiKi:
It seems that the system is weakened by this PR, because unexpected data are silently ignored and written as default value. It seems to fix the symptom, but not the cause. What was the rationale behind removing the exceptions ?

PowerKiKi:
Nevermind, I got confused. Merged manually.

PR #173, opened 2017-06-13
Added method Worksheet::cleanRowDimensions()

mahagr:
This is:

- [x] a bugfix
- [x] a new feature

Checklist:

- [ ] Code style is respected

What does it change?

Added method Worksheet::cleanRowDimensions() to work around memory running out with LibreOffice files having useless dimensions over 1M+. Also good for cleaning up the document after loading or removing a bunch of rows.

Fix was made for my own use, please be free to include it if you find the change useful.

### DIFF
--- a/src/PhpSpreadsheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet.php
@@ -493,6 +493,31 @@ class Worksheet implements IComparable
     }
 
     /**
+     * Clean up row dimensions.
+     *
+     * @param int $maxRow Maximum row to keep. Defaults to highest data row.
+     *
+     * @return $this
+     */
+    public function cleanRowDimensions($maxRow = null)
+    {
+        if ($maxRow === null) {
+            $maxRow = $this->getHighestDataRow();
+        }
+
+        foreach ($this->rowDimensions as $i => $dimension) {
+            if ($dimension->getRowIndex() > $maxRow) {
+                unset($this->rowDimensions[$i]);
+            }
+        }
+
+        // Garbage collect...
+        $this->garbageCollect();
+
+        return $this;
+    }
+
+    /**
      * Get default row dimension.
      *
      * @return Worksheet\RowDimension


PowerKiKi:
If I understand correctly there are  too many `RowDimension` written.  And you would prefer to have nothing written at all to fallback to default dimension ?

Or is it that `RowDimension` is written even for rows that are empty ?

mahagr:
@PowerKiKi the issue isn't the number of `RowDimension`, but how the PHP app seems to think that all the rows between 1 and max `RowDimension` contain data. For some unknown reason, LibreOffice creates files with a few `RowDimension` items being 1 000 000+ and my server is running out PHP memory, no matter how much memory I give to it.

My workaround is to just to have a function in case if you want to ignore any `RowDimension` items which are higher than the data itself. It is not being used automatically, it is just there if you want to remove those items.

PowerKiKi:
It's still unclear what happens, would you mind sharing a code example for the usage of this method ?

mahagr:
Here's simplest example of breaking code:

```
        /** @var Spreadsheet $spreadsheet */
        $spreadsheet->getSheet(0)->getRowDimension(1000000);

        // Save Excel file
        /** @var Xlsx $writer */
        $writer->setIncludeCharts(true);
        $writer->save('php://output');
```

If you add `$spreadsheet->getSheet(0)->cleanRowDimensions();` before saving, you can "fix" the file.

mahagr:
Hmm, maybe it is a bug in save method? Missing second parameter when calling `Worksheet::getRowDimension()` which basically causes the save to accidentally create all the objects instead of just getting the existing ones?

Even if it is so, I still think that it would be benefitical to be able to remove extra RowDimensions from the document. :)

PowerKiKi:
That's what I was thinking. It should be an internal optimization when writing, instead of having to be called explicitly by end-users.

mahagr:
Honestly, in this case, it looks like that some properties from row dimension are used to write each individual row. What the writer should do instead is not to write anything per row but add the row dimensions back to the document...?

I would still like to have the function that I wrote, though, in case if you remove rows and want to remove also row dimensions which aren't used anymore.

PowerKiKi:
To me its sounds like even the second use-case of removed rows should be an internal optimization. If a row is removed, its dimension should be automatically removed too. Don't you think so ?

mahagr:
I'm not sure about that. I think that dimensions are applied to all rows between the dimension values, so to me, it looks like it sets the defaults for a number of rows (maybe all rows smaller than x).

PowerKiKi:
I am not quite sure what to do with this PR. If you are able to turn it into an internal optimization, covered by unit tests, I'll merge it. But I don't think I would merge a "helper" method that has to be used explicitly by end-user. What do you think ?

PowerKiKi:
Take your time, there is no rush here. But I'll appreciate if you let us know if you decide to give it up at a later time.

oleibman:
RowDimensions are no longer automatically created or written with the sample code.

PR #174, opened 2017-06-13
Added support for custom indexed colors in Xlsx reader

mahagr:
This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [ ] Code style is respected

What does it change?

Added support for custom indexed colors in Xlsx reader. Previously the built-in index was always used, which resulted in wrong colors in the cells.

The fix was made for my own use, please be free to include it if you find the change useful.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -546,6 +546,13 @@ class Xlsx extends BaseReader implements IReader
                         'SimpleXMLElement',
                         Settings::getLibXmlLoaderOptions()
                     );
+                    if ($xmlStyles->colors->indexedColors) {
+                        $colors = [];
+                        foreach ($xmlStyles->colors->indexedColors as $rgbColor) {
+                            $colors[] = (string) $rgbColor['rgb'];
+                        }
+                        Style\Color::setIndexedColors($colors);
+                    }
                     $numFmts = null;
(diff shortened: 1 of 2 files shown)


src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -546,6 +546,13 @@ public function load($pFilename)
                         'SimpleXMLElement',
                         Settings::getLibXmlLoaderOptions()
                     );
+                    if ($xmlStyles->colors->indexedColors) {
+                        $colors = [];
+                        foreach ($xmlStyles->colors->indexedColors as $rgbColor) {
+                            $colors[] = (string) $rgbColor['rgb'];
```
PowerKiKi: The index here starts at 0, but our hardcoded `$indexedColors` starts at 1. Is this intended ?

mahagr: Good catch, I need to check it out.

PowerKiKi: Did you have time to double-check that index thing ?

mahagr: Unfortunately, I've not had chance to check this out yet. Please keep it open for now.

oleibman:
Equivalent code added with PR #2595 (Feb. 2022) and PR #3096 (Sep. 2022).

Issue #170, opened 2017-06-12
Invalid content in VML comments


emeraldjava:
### What are the steps to reproduce?

Hi,

I'm trying to load a excel file using the load function via the code below. I want to be able to write data to the workbook

```php
$spreadsheet = PhpOffice\PhpSpreadsheet\IOFactory::load("BHAA.Race.Master.xlsm");
```

see https://github.com/emeraldjava/bhaa-racemaster/blob/master/BHAA.Race.Master.xlsm for the target file.

### What is the current behavior?

I get this error when loading, which indicated there are invalid XML details in the VML comments

```
[12-Jun-2017 09:14:36 UTC] PHP Warning:  simplexml_load_string():  &lt;/v:shape&gt;&lt;/xml&gt; in /home/pauloconnell/projects/github/bhaawp/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Excel2007.php on line 1354
[12-Jun-2017 09:14:36 UTC] PHP Warning:  simplexml_load_string():                  ^ in /home/pauloconnell/projects/github/bhaawp/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Excel2007.php on line 1354
[12-Jun-2017 09:14:36 UTC] PHP Fatal error:  Call to a member function registerXPathNamespace() on boolean in /home/pauloconnell/projects/github/bhaawp/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Excel2007.php on line 1355
```

Looking at the code in the Excel2007 file

```
// Loop through VML comments
foreach ($vmlComments as $relName => $relPath) {
    // Load VML comments file
    $relPath = \PhpOffice\PhpSpreadsheet\Shared\File::realpath(dirname("$dir/$fileWorksheet") . '/' . $relPath);
    $vmlCommentsFile = simplexml_load_string(
        $this->securityScan($this->getFromZipArchive($zip, $relPath)),
        'SimpleXMLElement',
        \PhpOffice\PhpSpreadsheet\Settings::getLibXmlLoaderOptions()
    );
    $vmlCommentsFile->registerXPathNamespace('v', 'urn:schemas-microsoft-com:vml');
```

because the 'simplexml_load_string' fails, the subsequent '$vmlCommentsFile->registerXPathNamespace()' is not a valid reference.

### Which versions of PhpSpreadsheet and PHP are affected?
 
I'm using php 5.6.30. 

[...]

PR #175, opened 2017-06-13
#170 Handle if simplexml_load_string() returns false, and report the …

emeraldjava:
…which sheet contains the invalid xml.

This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [ ] Code style is respected

What does it change?

Extended the logic around processing the VML XML comments to ensure we only parse blocks of valid xml. An error message is shown with sheets that contain <br> elements in the xml.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -1357,63 +1357,67 @@ class Xlsx extends BaseReader implements IReader
                                         'SimpleXMLElement',
                                         Settings::getLibXmlLoaderOptions()
                                     );
-                                    $vmlCommentsFile->registerXPathNamespace('v', 'urn:schemas-microsoft-com:vml');
 
-                                    $shapes = $vmlCommentsFile->xpath('//v:shape');
-                                    foreach ($shapes as $shape) {
-                                        $shape->registerXPathNamespace('v', 'urn:schemas-microsoft-com:vml');
+                                    if($vmlCommentsFile !== FALSE) {
+                                        $vmlCommentsFile->registerXPathNamespace('v', 'urn:schemas-microsoft-com:vml');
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -1357,63 +1357,67 @@ public function load($pFilename)
...
+                                                        }
+                                                        if ($stylePair[0] == 'visibility') {
+                                                            $comment->setVisible($stylePair[1] == 'visible');
+                                                        }
                                                     }
                                                 }
                                             }
                                         }
+                                    } else {
+                                        error_log('Invalid VML Comment XML File '.$fileWorksheet.'/'.$relPath);
```
PowerKiKi: IMHO, if the xlsx file is invalid an exception should be thrown instead of an error being logged somewhere

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -1357,63 +1357,67 @@ public function load($pFilename)
...
-                                                if (isset($clientData['ObjectType']) && (string) $clientData['ObjectType'] == 'Note') {
-                                                    $temp = $clientData->xpath('.//x:Row');
-                                                    if (is_array($temp)) {
-                                                        $row = $temp[0];
-                                                    }
 
-                                                    $temp = $clientData->xpath('.//x:Column');
-                                                    if (is_array($temp)) {
-                                                        $column = $temp[0];
+                                    if($vmlCommentsFile !== FALSE) {
```
PowerKiKi: Instead of adding one more level of nested block, it should be extracted as a private method.

PowerKiKi:
Travis fails and the PR lack unit tests to cover new code. Please fix those in order to be merged.

oleibman:
No update in 7 years, no test case, no idea what problem this is trying to solve. Removing stale label, leaving closed.

Issue #176, opened 2017-06-14
Unnecessarily enforces maximum of 31 characters in sheet titles


MikkelPaulson:
This is:

- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)

### Summary

XLSX files (at the very least) support sheet titles over 31 characters, but PhpSpreadsheet enforces a 31-character limit regardless of file format. This would be annoying if it only affected sheet creation, but the validation code is also run on load. This means that PhpSpreadsheet is unable to load any valid XLSX file (or file of any other format) containing a sheet with a name over 32 characters.

### What is the expected behavior?

A worksheet title containing 32 characters is set correctly. An exception should only be thrown when saving to a format that doesn't support longer titles.

### What is the current behavior?

An exception is thrown with the message "Maximum 31 characters allowed in sheet code name.".

### What are the steps to reproduce?

(Shows creating a new file, but the same behaviour is encountered when loading an existing file with a sheet title over 31 chars.)

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();
$sheet = $spreadsheet->getSheet();
$sheet->setTitle('ten chars ten chars ten chars +2');
```

### Which versions of PhpSpreadsheet and PHP are affected?

PHPExcel 1.8, PhpSpreadsheet develop

PowerKiKi:
Thanks for this contribution of quality. Would you be able to find official specs for a few formats about sheet title length ? I'd say at least XLSX and ODS would be nice

MikkelPaulson:
Reviewing the ODS spec, the relevant XML property is [table:name](https://docs.oasis-open.org/office/v1.2/os/OpenDocument-v1.2-os-part1.html#attribute-table_name). It specifies the data type as "string" without additional constraints. Obviously applications will impose their own limits in terms of what they'll render, but LibreOffice happily opened a file containing an entire paragraph of lorem ipsum.

<img width="1061" alt="screen shot 2017-06-14 at 20 33 46" src="https://user-images.githubusercontent.com/585944/27160265-e643d29c-5140-11e7-991f-11b53e6b9f1c.png">

PowerKiKi:
[...]

PR #177, opened 2017-06-14
Remove 31-character limit from sheet title

MikkelPaulson:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

What does it change?

There is no documented reason why sheet titles were being limited to 31 characters (this code has been in place since PHPExcel was initially migrated to GitHub 7 years ago). It's possible that some archaic format does have trouble with this, but imposing that limit globally presents problems for parsing all formats.

If this is needed for some format, the restriction should be imposed on save (and only on save) to the affected format, so as not to affect other formats.

Issue #176

### DIFF
--- a/src/PhpSpreadsheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet.php
@@ -430,11 +430,6 @@ class Worksheet implements IComparable
             throw new Exception('Invalid character found in sheet code name');
         }
 
-        // Maximum 31 characters allowed for sheet title
-        if ($CharCount > 31) {
-            throw new Exception('Maximum 31 characters allowed in sheet code name.');
-        }
-
         return $pValue;
     }
 
@@ -452,11 +447,6 @@ class Worksheet implements IComparable
         // Some of the printable ASCII characters are invalid:  * : / \ ? [ ]
         if (str_replace(self::$invalidCharacters, '', $pValue) !== $pValue) {
             throw new Exception('Invalid character found in sheet title');
-        }
-
-        // Maximum 31 characters allowed for sheet title
-        if (Shared\StringHelper::countCharacters($pValue) > 31) {
-            throw new Exception('Maximum 31 characters allowed in sheet title.');
         }
 
         return $pValue;


MikkelPaulson:
This PR is for discussion; other options include:

* Skip length validation on load but enforce when setting manually
* Add validation to Writer object for affected format, which would be possible if there was some indication of what problem is supposed to be solved by this limit

PowerKiKi:
LibreOffice 5.1 and Google Spreadsheet are able to read and write Ods, Xls, Xlsx and Xml with sheet names at least up to 50 characters, most likely much more.

Excel 2016 refuse to input anything longer than 31 characters:
image

And it will read Ods, Xls, Xlsx and Xml with sheet names at least up to 50 characters, but it will show an error when opening:
image

And it will truncate the name and show the input error when editing it.

All of this make me thinks that the limitation is purely in Excel GUI, and not in file formats ([confirmed here](https://stackoverflow.com/a/3681908/37706)). But unfortunately that still makes for a very poor user experience for an end-user using even the latest version of Excel. And for that reason I am rather uncomfortable removing that limit.

If Excel was not the most popular spreadsheet soft out there, it may be different, but I don't think it's a good idea to allow creating errors for a large majority of end-users that are not necessarily the most tech-literate.

MikkelPaulson:
@PowerKiKi Thanks for the detailed information. Unfortunately, this is still a significant bug, since it doesn't just affect _saving_ data, it also affects _loading_. I can create a file in Excel, or OpenOffice, or whatever, with a sheet name over 31 characters, and PHPExcel/PhpSpreadsheet will straight-up refuse to parse the file. No way to recover, no way to override, just "invalid file" and exit.

Perhaps removing the check altogether isn't the solution, but there are alternative ways to deal with it, as I mentioned in my comment above. What about simply mimicing Excel's behaviour of truncating the sheet name at 31 characters? Or the Excel writer could truncate/error out if a sheet has over 31 characters, which limits the damage to the affected format.

PowerKiKi:
The loading aspect went over my head. You are correct in saying that parsing must work regardless of sheet name length.

Doing something for a specific format would not work, because it has nothing to do with format at all, it really is a (stupid) limitation of Excel GUI. So I'd say truncating the sheet in `Worksheet::setTitle()` should be the way to go. What do you think ?

MikkelPaulson:
Not the best option IMO, but better than nothing. Truncating titles at load time could potentially result in a file with duplicate sheet titles, so doing so requires the addition of logic to apply unique naming (ABCDEF, ABCDE1, ABCD10).

I would prefer to have the truncation/validation applied only when setting the title manually, so that users could load and re-save a file without having unintended side-effects.

PowerKiKi:
Good point, that would be great if we can avoid side-effects, and still prevent users from shooting themselves in the foot. 

Could we add a flag `$validate = true` to `setTitle()` and modify all readers to specify `false` ? or would you have something else in mind ?

dadaxr:
any news on that PR ?

MikkelPaulson:
@dadaxr It was merged – see #186

cmanley:
Instead of just removing the limit, why not make it an option (default = 31) so that the user can decide what to do with it and the risks involved (if any)?

PR #178, opened 2017-06-16
Add feature to get text & background color from inline css, for td & …

kifni41:
…th element

This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [x] Code style is respected

What does it change?

I'm not sure if others need this feature, but this feature is quite handy for me and my team to directly generate excel file from html text, usually our client want specific coloring for the excel file and usually we already have the html for that. 

This is limited for inline css inside <td> and <th> tag.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Html.php
+++ b/src/PhpSpreadsheet/Reader/Html.php
@@ -442,6 +442,30 @@ class Html extends BaseReader implements IReader
                     case 'td':
                         $this->processDomElement($child, $sheet, $row, $column, $cellContent);
 
+                        // add color styles (background & text) from dom element,currently support : td & th, using ONLY inline css style with RGB color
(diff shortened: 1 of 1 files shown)


PowerKiKi:
This looks interesting. Would you have time to cover it with unit tests and fix the code style ?

kifni41:
@PowerKiKi okay bro, i'll try,
I forgot to make new branch for this commit, should be on feature/ branch right?

PowerKiKi:
Yes, it would be best to be on its on branch, so you can keep pushing commits relating to that topic. But if you know what you are doing and only work on this topic, that coud be ok too...

kifni41:
@PowerKiKi okay, created new branch for that. new PR submitted.
https://github.com/PHPOffice/PhpSpreadsheet/pull/180

PR #180, opened 2017-06-20
Add feature to get text & background color from inline CSS : td&th

kifni41:
This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [x] Code style is respected

What does it change?

> Updated Pull Request from https://github.com/PHPOffice/PhpSpreadsheet/pull/178 , forgot to create new branch

I'm not sure if others need this feature, but this feature is quite handy for me and my team to directly generate excel file from html text, usually our client want specific coloring for the excel file and usually we already have the html for that.

This is limited for inline css inside and tag.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Html.php
+++ b/src/PhpSpreadsheet/Reader/Html.php
@@ -443,22 +443,22 @@ class Html extends BaseReader implements IReader
                         $this->processDomElement($child, $sheet, $row, $column, $cellContent);
 
                         // add color styles (background & text) from dom element,currently support : td & th, using ONLY inline css style with RGB color
-                        if (isset($attributeArray['style'])){
-                            $styles = explode(';',$attributeArray['style']);
+                        if (isset($attributeArray['style'])) {
+                            $styles = explode(';', $attributeArray['style']);
                             foreach ($styles as $st) {
                                 $value = explode(':', $st);
                                 if (!empty($value[0])) {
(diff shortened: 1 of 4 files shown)


PowerKiKi:
I think it would be best to extract your new code into a private method. Also `docs/references/features-cross-reference.md` should be updated accordingly.

kifni41:
@PowerKiKi that's good idea, i'll update later..

src/PhpSpreadsheet/Reader/Html.php
```diff
@@ -584,4 +587,48 @@ public function securityScan($xml)
...
+     * @param array $attributeArray
+     * @param int $row
+     * @param string $column
+     */
+    private function applyInlineStyle(&$sheet, $row, $column, $attributeArray)
+    {
+        $supported_styles = ['background-color', 'color'];
+
+        // add color styles (background & text) from dom element,currently support : td & th, using ONLY inline css style with RGB color
+        if (isset($attributeArray['style'])) {
```
PowerKiKi: I think we can return early here to avoid too many nested blocks:

```php
if (!isset($attributeArray['style'])) {
    return;
}
```

PowerKiKi:
Thanks for the contribution

kifni41:
you're welcome, i hope it'll be useful for others and i can contribute more in the future...

PR #186, opened 2017-07-10
Add option to suppress validation of sheet titles

### LINKED ISSUE #176: Unnecessarily enforces maximum of 31 characters in sheet titles

MikkelPaulson:
**This is:**

- [x] a bugfix
- [ ] a new feature

**Checklist:**

- [x] Code style is respected

**What does it change?**

Based on a "lowest common denominator" approach to compatibility, we will continue to enforce a 31-character limit for sheet titles. However, this limit should not be enforced when loading an existing file.

Added a new optional parameter to Worksheet::setTitle() and Worksheet::setCodeName() to suppress validation and massaging, based on the premise that existing files should be given a best-effort approach to loading and parsing. Unfortunately, it's not possible with the current architecture to prevent users from making use of this functionality, aside from with a strongly-worded warning.

Added test coverage. I didn't see any existing unit tests of the Worksheet class, so I created a new test to cover these methods.

Fixes #176.

### DIFF
--- a/src/PhpSpreadsheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet.php
@@ -825,52 +825,54 @@ class Worksheet implements IComparable
      * Set title.
      *
      * @param string $pValue String containing the dimension of this worksheet
-     * @param string $updateFormulaCellReferences boolean Flag indicating whether cell references in formulae should
+     * @param bool $updateFormulaCellReferences Flag indicating whether cell references in formulae should
      *            be updated to reflect the new sheet name.
      *          This should be left as the default true, unless you are
      *          certain that no formula cells on any worksheet contain
      *          references to this worksheet
+     * @param bool $validate False to skip validation of new title. WARNING: This should only be set
+     *                       at parse time (by Readers), where titles can be assumed to be valid.
      *
      * @return Worksheet
      */
-    public function setTitle($pValue, $updateFormulaCellReferences = true)
+    public function setTitle($pValue, $updateFormulaCellReferences = true, $validate = true)
     {
         // Is this a 'rename' or not?
         if ($this->getTitle() == $pValue) {
             return $this;
         }
 
-        // Syntax check
-        self::checkSheetTitle($pValue);
-
         // Old title
         $oldTitle = $this->getTitle();
 
-        if ($this->parent) {
-            // Is there already such sheet name?
-            if ($this->parent->sheetNameExists($pValue)) {
-                // Use name, but append with lowest possible integer
+        if ($validate) {
+            // Syntax check
+            self::checkSheetTitle($pValue);
 
-                if (Shared\StringHelper::countCharacters($pValue) > 29) {
-                    $pValue = Shared\StringHelper::substring($pValue, 0, 29);
-                }
-                $i = 1;
-                while ($this->parent->sheetNameExists($pValue . ' ' . $i)) {
-                    ++$i;
-                    if ($i == 10) {
-                        if (Shared\StringHelper::countCharacters($pValue) > 28) {
-                            $pValue = Shared\StringHelper::substring($pValue, 0, 28);
-                        }
-                    } elseif ($i == 100) {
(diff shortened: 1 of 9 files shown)


src/PhpSpreadsheet/Worksheet.php
```diff
@@ -825,52 +825,54 @@ public function getTitle()
...
+                            if (Shared\StringHelper::countCharacters($pValue) > 27) {
+                                $pValue = Shared\StringHelper::substring($pValue, 0, 27);
+                            }
                         }
                     }
-                }
 
-                $altTitle = $pValue . ' ' . $i;
-
-                return $this->setTitle($altTitle, $updateFormulaCellReferences);
```
MikkelPaulson: Rather than calling the method recursively, pre-validated and then massaged the value prior to setting.

tests/PhpSpreadsheetTests/WorksheetTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+    public function setTitleInvalidProvider()
+    {
+        return [
+            [str_repeat('a', 32), 'Maximum 31 characters allowed in sheet title.'],
+            ['invalid*title', 'Invalid character found in sheet title'],
+        ];
+    }
+
+    /** @dataProvider setTitleInvalidProvider */
```
PowerKiKi: It seems code style is [choking on this](https://travis-ci.org/PHPOffice/PhpSpreadsheet/jobs/252118853#L484). Could you make it a standard docblock please ? Then I'll merge it.

tests/PhpSpreadsheetTests/WorksheetTest.php
```diff
@@ -0,0 +1,118 @@
...
+
+    public function setCodeNameInvalidProvider()
+    {
+        return [
+            [str_repeat('a', 32), 'Maximum 31 characters allowed in sheet code name.'],
+            ['invalid*code*name', 'Invalid character found in sheet code name'],
+        ];
+    }
+
+    /** @dataProvider setCodeNameInvalidProvider */
```
PowerKiKi: same here

MikkelPaulson:
IMO the ideal solution would be to have the constructor take a title that is not validated, but there's no reason to think a user will use `Spreadsheet::createSheet()` exclusively rather than `Spreadsheet::addSheet(new Worksheet())`.

tests/PhpSpreadsheetTests/WorksheetTest.php
MikkelPaulson: `/** ... */` is [standard](https://www.phpdoc.org/docs/latest/guides/docblocks.html#anatomy-of-a-docblock), but I think what the code style is choking on is the presence of a docblock that does not document  all parameters.

Regardless, should be fixed now.

MikkelPaulson:
@PowerKiKi Look good?

PowerKiKi:
Thanks for this great PR. I fixed camelCase variable names and merged it.

sumit22:
Please let me know, which version should i use to get this fix?

PowerKiKi:
@sumit22 1.0.0-beta

PR #189, opened 2017-07-28
Update Border.php

manuakasam:
Changing the switch case to `strtolower` as internally the case is still `allBorders` given the following execution stack.

```php
$this->excel->getActiveSheet()
    ->getStyle($style)
    ->getBorders()
    ->getAllBorders()
    ->setBorderStyle(ExcelStyle\Border::BORDER_THIN);
```

This likely is a hacking change and may not get merged, but it highlights the issue as well. May count this as a bug report ;)

The alternative would likely to fix the occurence of the camelcased `allBorders` and change it to the lowercased `allborders`

This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

What does it change?

### DIFF
--- a/src/PhpSpreadsheet/Style/Border.php
+++ b/src/PhpSpreadsheet/Style/Border.php
@@ -146,7 +146,7 @@ class Border extends Supervisor implements IComparable
      */
     public function getStyleArray($array)
     {
-        switch ($this->parentPropertyName) {
+        switch (strtolower($this->parentPropertyName)) {
             case 'allborders':
             case 'bottom':
             case 'diagonal':


PowerKiKi:
Why not change everything to camelcase `allBorders` ? it seems it would make more sense, wouldn't it ?

manuakasam:
I have not enough insight of the project and don't know how many configurable keys there are and if they are camelCased, under_scored or lowercased. Should be unified with the rest of the project of course.

ProNoob13:
As displayed in getSharedComponent, right above the method modified in this pull request, the property should only contain the camelCase version of the string. That means it isn't necessary to check for differently typed versions of the string, or to use `strtolower` if you ask me.

PowerKiKi:
I opted to standardize all keys to camelCase with the same wording as their getter/setter counterpart to have a more coherent experience.

PR #206, opened 2017-08-28
Remove namespaces form Xlsx files

peterrehm:
This is:

- [X] a bugfix

Checklist:

- [ ] Code style is respected

What does it change?

I have ported the workaround from here: https://github.com/PHPOffice/PHPExcel/issues/1187#issuecomment-295032648

I have not added Tests as it seems the Xlsx Reader is currently not under test.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -88,9 +88,9 @@ class Xlsx extends BaseReader implements IReader
         if ($zip->open($pFilename) === true) {
             // check if it is an OOXML archive
             $rels = simplexml_load_string(
-                $this->securityScan(
+                $this->removeNamespaces($this->securityScan(
                     $this->getFromZipArchive($zip, '_rels/.rels')
-                ),
+                )),
                 'SimpleXMLElement',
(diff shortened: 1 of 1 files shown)


PowerKiKi:
It is way too dangerous to `str_replace()` in strings that can contains user-generated content and thus this PR won't be merged as is. This should be done with real parsing of XML instead.

Would you be able to come up with a more robust `removeNamespaces()` method ?

peterrehm:
Hm, what would you suggest? I just wanted to provide an option I do not know how commonly it is needed. 

It took me quit a while to figure out what is wrong and fortunately I found the above linked post.

PowerKiKi:
I don't have a solution out of the box, but it should probably involve `simplexml` or `XMLReader`, or other robust XML libraries to avoid replacing user content by mistake.

PowerKiKi:
I'll close it for now, not to give false hopes. But if a more robust solution appears, feel free to re-submit.

xakepsoft:
Actually this is a safe and robust solution, a bit "hacky" though. Collision with user content is impossible because <CDATA is not allowed in XLSX files.  
This can be a viable solution until next major version of PhpSpreadsheet is released.

PR #210, opened 2017-08-31
add installation section

brutalcrozt:
give 2 way via json or cli

This is:

- [ ] a bugfix
- [ ] a new feature
- [x] documentation update

Checklist:

- [ ] Code style is respected

What does it change?
Change the README.md , as i informed [here](https://github.com/PHPOffice/PHPExcel/pull/1303#event-1227480293)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,6 +16,38 @@ Develop:
 
 PhpSpreadsheet is a library written in pure PHP and providing a set of classes that allow you to read from and to write to different spreadsheet file formats, like Excel and LibreOffice Calc.
 
+## Requirements
+  * [Composer](https://getcomposer.org/) installed
+<!---  (this is based on the old requirement on phpExcel)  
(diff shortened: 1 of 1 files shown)


PowerKiKi:
Install instructions already are in the documentation and we don't want to duplicate then here.

Also install "via json" is a poor practice and should never be recommended, because it increases the chance of messing up and more critically it forces you to manually choose a version. And that is a documentation maintenance nightmare.

brutalcrozt:
ok i got it, are you interested add support for offline documentation on devdocs.io ? The requirement is listed [here](https://trello.com/c/PVnfdeaN/26-suggest-new-docs-here)

PowerKiKi:
I'm not against it, but I don't think we are popular enough according to the requirement. But if you are willing to submit it anyway, be my guest.

PR #218, opened 2017-09-09
Prevent color palette changing

al-lala:
This is:

- [X] a bugfix
- [ ] a new feature

Checklist:

- [X] Code style is respected

What does it change?
Don't create color in palette when already exist.

### DIFF
--- a/docs/topics/recipes.md
+++ b/docs/topics/recipes.md
@@ -910,15 +910,18 @@ practice...
 
 ## Setting security on a spreadsheet
 
-Excel offers 3 levels of "protection": document security, sheet security
-and cell security.
+Excel offers 3 levels of "protection":
 
-Document security allows you to set a password on a complete
+- Document: allows you to set a password on a complete
 spreadsheet, allowing changes to be made only when that password is
-entered.Worksheet security offers other security options: you can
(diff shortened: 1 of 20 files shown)


PowerKiKi:
Could you please explain why this should be done ? what happened before this modification ?

Also it will need to be covered by unit tests to be merged.

al-lala:
> Sorry for my bad english ( i'm frech ) :

**Before :** 
Palette colors are changed systematicaly when color are find in workbook.
Original palette setting are useless in this case.

**After :** 
Palette colors are preserved if colors already exist in initial palette.
Original palette setting are used and conserved.

**This resolve this issue :** 
If you copy sheet from xls file saved by phpspreadsheet to other xls file, the colors changes.

**Unit tests**
For me, new unit test are unnecessary.

PowerKiKi:
Thanks for the explanation, I understand better the reason for the change. However unit tests are important to prevent regression in the future. I won't be able to merge without unit tests. It could be as simple as writing a file, reading it and checking the color palette.

al-lala:
I have correct an error and make Unit Tests.

> Thanks for PHPOffice/PHPSpreadsheet, this is the best way to read/write Excel in php context.

PowerKiKi:
Thank you, I made a few minor tweaks and merged it.

PR #222, opened 2017-09-18
Added SimpleXMLElement Exception check

GreatHumorist:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

What does it change?
When the xml file get is not a standard xml file, the simplexml_load_string will return false, this will cause an error on "$xml->getNamespaces(true);" . So, instead of show the error, throw an exception may be better for it.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xml.php
+++ b/src/PhpSpreadsheet/Reader/Xml.php
@@ -139,7 +139,7 @@ class Xml extends BaseReader implements IReader
             Settings::getLibXmlLoaderOptions()
         );
         if (!($xml instanceof \SimpleXMLElement)) {
-            throw new Exception('SimpleXMLElement can not load ' .$pFilename);
+            throw new Exception('SimpleXMLElement can not load ' . $pFilename);
         }
         $namespaces = $xml->getNamespaces(true);
 
@@ -171,7 +171,7 @@ class Xml extends BaseReader implements IReader
             Settings::getLibXmlLoaderOptions()
         );
         if (!($xml instanceof \SimpleXMLElement)) {
-            throw new Exception('SimpleXMLElement can not load ' .$pFilename);
+            throw new Exception('SimpleXMLElement can not load ' . $pFilename);
         }
         $namespaces = $xml->getNamespaces(true);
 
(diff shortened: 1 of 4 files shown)


src/PhpSpreadsheet/Reader/Xml.php
```diff
@@ -167,6 +170,9 @@ public function listWorksheetInfo($pFilename)
             'SimpleXMLElement',
             Settings::getLibXmlLoaderOptions()
         );
+        if (!($xml instanceof \SimpleXMLElement)) {
```
PowerKiKi: Should compare against `false` instead of checking types, so `if ($xml === false)`.

Error message should include error details from simplexml.

src/PhpSpreadsheet/Reader/Xml.php
```diff
@@ -138,6 +138,9 @@ public function listWorksheetNames($pFilename)
...
+            throw new Exception('SimpleXMLElement can not load ' . $pFilename);
```
PowerKiKi: Should compare against `false` instead of checking types, so `if ($xml === false)`.

Error message should include error details from simplexml.

src/PhpSpreadsheet/Reader/Xml.php
```diff
@@ -344,6 +350,9 @@ public function loadIntoExisting($pFilename, Spreadsheet $spreadsheet)
             'SimpleXMLElement',
             Settings::getLibXmlLoaderOptions()
         );
+        if (!($xml instanceof \SimpleXMLElement)) {
```
PowerKiKi: idem

src/PhpSpreadsheet/Reader/Xml.php
GreatHumorist: function **simplexml_load_string** returns:
>SimpleXMLElement an object of class SimpleXMLElement with * properties containing the data held within the xml document, or <b>FALSE</b> on failure.

It just return **false** when error occurred, so there's no more error information.

GreatHumorist:
@PowerKiKi I had added the unit test and change log and fixed the code style.

PowerKiKi:
Thanks !

GreatHumorist:
@PowerKiKi my pleasure

PR #230, opened 2017-09-27
fix DAY() function with 0<x<1 input

SailorMax:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

What does it change?

  =DAY(0.9) has to return 0. As in Excel.


PowerKiKi:
This can't be merged, because it breaks existing tests, and it lack new test case to cover the new case.

As a sidenote unfortunately the three major spreadsheet have different result for those cases:

image

So ideally we should use `Functions::getCompatibilityMode()` to switch between behaviors. But this PR as is would already be an improvement worth having.

SailorMax:
But current solution use only Excel method of calculation. DAY(1) corrently return 1 and DAY(0) return 0.
I just made this more accurate. :)

To support LibreOffice method here is new commit.

SailorMax:
Don't understand why in tests DAY(#NUM!) has to return -1?
Excel and LibreOffice has another result...

PowerKiKi:
Use the command `composer check` before pushing to be sure that everything is working.

For `DAY(#NUM!)`, if you are sure that Excel 2016 result is different than the test, then it should probably be changed in the test data (and code).

SailorMax:
Sorry, I misunderstood where is input data and where is result in test case :)

SailorMax:
Is it ready to merge now?

PowerKiKi:
That looks good. The only thing I see missing is tests for `COMPATIBILITY_OPENOFFICE`, I would suggest something like the following and adapt data:

```php
    public function testDAY($expectedResultExcel, $expectedResultOpenOffice, ...$args)
    {
        $resultExcel = DateTime::DAYOFMONTH(...$args);
        self::assertEquals($expectedResultExcel, $resultExcel, null, 1E-8);

        Functions::setCompatibilityMode(Functions::COMPATIBILITY_OPENOFFICE);
        $resultOpenOffice = DateTime::DAYOFMONTH(...$args);
        self::assertEquals($expectedResultOpenOffice, $resultOpenOffice, null, 1E-8);
    }
```

SailorMax:
I'm not expert in unit tests for this.
Looks like for Open Office we need separate file with input data => separate providerDAY() method=> separate testDAY() method => How we can name it?
Or I wrong?

PowerKiKi:
@SailorMax Modify the test method like I suggested and modify the data like so:

Before:
```php

return [
    [
        19,
        22269,
    ],
// ...
```
After:
```php

return [
    [
        19, // Result for Excel
        19, // Result for OpenOffice
        22269,
    ],
// ...
```

SailorMax:
done?

tests/PhpSpreadsheetTests/Calculation/DateTimeTest.php
```diff
@@ -178,8 +178,19 @@ public function providerWEEKDAY()
      */
     public function testDAY($expectedResult, ...$args)
```
PowerKiKi: you can simplify the method body if you use a declaration like `testDAY($expectedResultExcel, $expectedResultOpenOffice, ...$args)`. So you don't have to use tricks like `array_shift()`

SailorMax: done.

PowerKiKi:
Thank you very much for your contribution !

PR #233, opened 2017-09-28
fix return format of INDEX() function

SailorMax:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

What does it change?
  INDEX() lost row number in result => users has warning messages:
```
Calculation.php: dataTestReference() at 3417
 array_keys() expects parameter 1 to be array, float given
Calculation.php: dataTestReference() at 3417
 array_keys() expects parameter 1 to be array, null given
Calculation.php: dataTestReference() at 3418
 array_shift() expects parameter 1 to be array, null given
Calculation.php: getMatrixDimensions() at 2824
 count(): Parameter must be an array or an object that implements Countable
```


src/PhpSpreadsheet/Calculation/LookupRef.php
```diff
@@ -647,7 +647,7 @@ public static function INDEX($arrayValues, $rowNum = 0, $columnNum = 0)
                     if (isset($arrayColumn[$rowNum])) {
                         $returnArray[] = $arrayColumn[$rowNum];
                     } else {
-                        return $arrayValues[$rowNum];
+                        return array( $rowNum => $arrayValues[$rowNum] );
```
GreatHumorist: array should write to []

SailorMax: array() in this case is much intuitive. But if you insist... :)

GreatHumorist: cause phpspreadsheet requires php5.6 and newer, the new feature recommend use [] instead of array(), or it won't pass the CI system.

PowerKiKi:
This should be covered by unit tests. You could add some in `tests/PhpSpreadsheetTests/Calculation/LookupRefTest.php`. Have a look at existing tests to see how to structure them.

SailorMax:
Is it ready to merge now?

PowerKiKi:
Thank you !

PR #234, opened 2017-09-28
Added Chinese documentation support

GreatHumorist:
This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [ ] Code style is respected

What does it change?
Added Chinese documentation support, the rest topic I will translate later.

### DIFF
--- a/docs/i18n/zh_cn/index.md
+++ b/docs/i18n/zh_cn/index.md
@@ -1,0 +1,106 @@
+# 欢迎阅读 PhpSpreadsheet 中文文档
+
+![Logo](../../assets/logo.svg)
+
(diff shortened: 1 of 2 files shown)


PowerKiKi:
Thank you for the effort, however this will not be merged.

Maintaining the documentation in one language is already a huge workload. We cannot officially support more than one language, because it would be too much work, and would raise even further issues when doc goes out of sync between languages.

I suggest automated tools such as Google Translate to translate the latest official doc on the fly. Or if you insist on a manual translation, then create your own project and host it somewhere, but it will not be part of this project.

GreatHumorist:
Ok, it's fine.

Issue #240, opened 2017-10-04
Raising Scrutinizer index


SenseException:
This is:

- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)

### What is the expected behavior?

Having no major issues in Scrutinizer.

### What is the current behavior?

Multiple issues, that are error prone.

For instance, missing properties in a class are fine, if someone writes into them first, but will result in an error, if a read operation comes first.

### What are the steps to reproduce?

Those issues are already reproduced by Scrutinizer.

https://scrutinizer-ci.com/g/PHPOffice/PhpSpreadsheet/issues/develop?selectedSeverities%5B0%5D=10&orderField=path&order=asc&honorSelectedPaths=0

### Which versions of PhpSpreadsheet and PHP are affected?

develop and master Branches are affected. This project need to reduce listed issues with multiple clean-up pull requests to prevent current and future issues and bugs. It needs a cleaner code handling too, why `var_dump` and similar debugging code shouldn't be committed into the develop branch.

This needs to be handled before new features are introduces. Also, no PRs introducing new crucial Severities should be merged without reasons.

PowerKiKi:
The best would probably to submit a PR for each type of issue found by scrutinizer. Or for a single issue if the changes are too important. Would you be willing to work on that?

SenseException:
I can provide some. Also, anybody is invited to pick up a few severities too.

endelwar:
@PowerKiKi you should ask Scrutinizer to enable their Security Analysis, it has been very valuable on another project that I follow

PowerKiKi:
@endelwar, I believe this is only available for paid account. So we won't be able to use it unfortunately...

endelwar:
@PowerKiKi it's free for non-commercial open-source project! Just send them an email (support@scrutinizer-ci.com) asking for it and pointing to project's github page.

PowerKiKi:
@endelwar, it [has been enabled](https://scrutinizer-ci.com/g/PHPOffice/PhpSpreadsheet/issues/develop), though since we don't deal directly with user inputs, it seems it's almost pointless for us... but just in case...

endelwar:
good, don't understimate the "bad guys" ;)

SenseException:
I'll see that I will find some time for another Scrutinizer fix PR.

[...]

PR #241, opened 2017-10-07
Add properties which do not seem to exist

SenseException:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

What does it change?

It adds the missing properties of classes like mentioned in #240 that were reported by Scrutinizer. I looked where properties were missing and if those are have to be public based on the current code state. `JpGraph` use classes that don't currently exist, so I wasn't able add a fix for them. The existing tests are still working.

Some of the properties look strange to me like having a `startColor` and `endColor` in `PhpOffice\PhpSpreadsheet\Style\Fill` while other classes like `PhpOffice\PhpSpreadsheet\Reader\Xls` are accessing properties like `startColorIndex` and `endColorIndex` on `Fill`.


PowerKiKi:
This is great, thanks !

PR #242, opened 2017-10-08
Introducing Scrutinizer configuration

SenseException:
This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [ ] Code style is respected

What does it change?

It introduces a Scrutinizer configuration (#240), so developers can test their PRs based on the config and extend it, in case some checks have to be added.


.scrutinizer.yml
```diff
@@ -0,0 +1,16 @@
+checks:
+    php: true
+
+filter:
+    excluded_paths:
```
PowerKiKi: Why would you want to exclude folders with actual PHP code ? I think we should strive to improve the entire code base, including tests and samples.

I'd remove the entire `filter` section

SenseException:
I've added the original config. Thank you for sharing it.

Issue #245, opened 2017-10-12
Migration tool doubles leading backslashes


KacerCZ:
This is:

- [X] a bug report
- [ ] a feature request
- [X] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)

### What is the expected behavior?
After migration class `\PHPExcel` will be `PhpOffice\PhpSpreadsheet\Spreadsheet`.

### What is the current behavior?
After migration class `\PHPExcel` is converted to `\\PhpOffice\PhpSpreadsheet\Spreadsheet`.

### What are the steps to reproduce?
Run migration tool on any PHP file containig fully qualified PHPExcel class name.

### Which versions of PhpSpreadsheet and PHP are affected?
PHP 7.1.10
Latest version from develop branch.

What fixed the issue was removal of leading backslashes from mapping in `Migrator` helper:
https://github.com/PHPOffice/PhpSpreadsheet/blob/develop/src/PhpSpreadsheet/Helper/Migrator.php

PR #246, opened 2017-10-12
Removed leading slashes from migrator map

KacerCZ:
Fixes https://github.com/PHPOffice/PhpSpreadsheet/issues/245


arturock:
Won't be better to use the `::class` constant?

KacerCZ:
I don't think so, because migration uses simple string replacement (so any 'PHPExcel' string is replaced also in comments). After this fix I was able to successfuly use migration tool without any further modifications to class references.

arturock:
I guess it may not be needed, but `::class` resolves to the same string than the hardcoded FQN without leading slash, however as the real class is being used it's safer and easier to spot any typo.

PowerKiKi:
Thanks to both of for your contributions. I ended up going with `::class` notation.

KacerCZ:
@arturock You are right, I didn't realize that `::class` returns class name without leading slash.

PR #248, opened 2017-10-13
Temporary directory is passed to mPDF

KacerCZ:
This is:

- [X] a bugfix
- [ ] a new feature

Checklist:

- [X] Code style is respected

What does it change?
Temporary directory set to PDF writer is passed to mPDF object in constructor.
mPDF 7.0 changed temporary directory configuration from constant to constructor parameter.
Without this change it is not possible to change default temporary directory.
See: https://mpdf.github.io/installation-setup/folders-for-temporary-files.html


PowerKiKi:
Thanks ! Next time you can include the nice explanation you wrote directly in the commit message body. So it will we be available when doing `git annotate` in the future.

PR #254, opened 2017-10-19
change array_push() to $array[]

mosaxiv:
http://php.net/manual/en/function.array-push.php#refsect1-function.array-push-description
> Note: If you use array_push() to add one element to the array it's better to use $array[] = because in that way there is no overhead of calling a function.


PowerKiKi:
Thank you

PR #257, opened 2017-10-20
Cell->hasValidValue() - test value on list validation

SailorMax:
This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [x] Code style is respected

What does it change?
  Add possibility to check value of cell on validity based on list of possible values.


src/PhpSpreadsheet/Cell.php
```diff
@@ -410,7 +410,53 @@ public function setDataValidation(Cell\DataValidation $pDataValidation = null)
...
+        $cellValidation = $this->getDataValidation();
+
+        if (!$cellValidation->getAllowBlank() && ($cellValue === null || $cellValue == '')) {
+            return false;
+        }
+
+        // TODO: write check on all cases
+        switch ($cellValidation->getType()) {
+            case Cell\DataValidation::TYPE_LIST:
+                $formula1 = $cellValidation->getFormula1();
```
PowerKiKi: We should extract the validation specific for a type into their own method. That would prevent ` hasValidValue()` to become too huge in the future when we add validation for other types.

src/PhpSpreadsheet/Cell.php
```diff
@@ -410,7 +410,53 @@ public function setDataValidation(Cell\DataValidation $pDataValidation = null)
...
+                        return in_array(strtolower($cellValue), explode(',', strtolower(trim($formula1, '"'))), true);
+                    } elseif (strpos($formula1, ':') > 0) {            // values list cells
+                        $match_formula = '=MATCH(' . $this->getCoordinate() . ',' . $formula1 . ',0)';
+
+                        try {
+                            $result = Calculation::getInstance(
+                                $this->getWorksheet()->getParent()
+                            )->calculateFormula($match_formula, $this->getCoordinate(), $this);
+
+                            return $result != '#N/A';
```
PowerKiKi: Let's use `Functions::NA()` instead of hardcoded string and a strict comparison `!==`.

PowerKiKi:
Oh btw, next time use `composer fix` to fix code style.

SailorMax:
in `PhpOffice\PhpSpreadsheet\Cell\Validator`
or in `PhpOffice\PhpSpreadsheet\Cell\DataValidation`?

as regular method or as static method?

PowerKiKi:
It should definitely be a different class than DataValidation.
Maybe DataValidator, if you want to make the relation between the two a bit
more obvious. And I'd go for instance methods to prevent pitfall of
singleton pattern in future code evolution.

So usage could be like:

    $validator = new DataValidator ();
    $isValid = $validator->isValid($cell);

SailorMax:
For this function I need $cell and $dataValidation.
Where in you sample I can put $dataValidation?

src/PhpSpreadsheet/Cell.php
```diff
@@ -410,7 +410,21 @@ public function setDataValidation(Cell\DataValidation $pDataValidation = null)
     }
 
     /**
-     *    Does this cell contain a Hyperlink?
+     * Does this cell contain valid value?
+     *
+     * @return bool
+     */
+    public function hasValidValue()
```
PowerKiKi: I think we can drop entirely this method. Keeping the cell class unchanged. This limit the public API to the strict minimum for easier maintenance in the future.

src/PhpSpreadsheet/Cell/DataValidator.php
```diff
@@ -0,0 +1,53 @@
...
+class DataValidator
+{
+    /**
+     * Does this cell contain valid value?
+	 * 
+     * @param Cell $cell Cell to check the value
+     *
+     * @return bool
+     */
+    public function isValidCellValue(&$cell)
```
PowerKiKi: Because there is only one thing that can be validated, the method can simply be "isValid". If you think it might not be clear enough, then we should find a better name for the class, but not the method.

Also the parameter must be type hinted with Cell, and it should not be passed by reference.

src/PhpSpreadsheet/Cell/DataValidator.php
```diff
@@ -0,0 +1,53 @@
...
+        $cellValue = $cell->getValue();
+        $dataValidation = $cell->getDataValidation();
+
+        if (!$dataValidation->getAllowBlank() && ($cellValue === null || $cellValue == '')) {
+            return false;
+        }
+
+        // TODO: write check on all cases
+        switch ($dataValidation->getType()) {
+            case DataValidation::TYPE_LIST:
```
PowerKiKi: Let's refactor the code for list validation in a private method. Something like "isValueInList()" or similar...

src/PhpSpreadsheet/Cell/DataValidator.php
```diff
@@ -0,0 +1,53 @@
...
+                $formula1 = $dataValidation->getFormula1();
+                if (!empty($formula1)) {
+                    if ($formula1[0] == '"') {                         // inline values list
```
PowerKiKi: This should use strict equality ===

src/PhpSpreadsheet/Cell.php
SailorMax: But this is part of cell as object. Isn't it?
Where is border? Which method has to be in the Cell, and which has to work as helper?

src/PhpSpreadsheet/Cell.php
SailorMax: done.

src/PhpSpreadsheet/Cell.php
SailorMax: @PowerKiKi 
So? Do you think? Or do you insist? :)
What is in priority? Maintenance or Usability?

PowerKiKi: I had another look and the `Cell` class is not as huge as I though it was, so let's keep the method here as you suggest. But the `if (!$this->hasDataValidation())` should still be moved in the validator, otherwise the validation logic would be split into two different places.

SailorMax: moved. But now we have some kind of conflict. GitHub do not allow me even view where is the problem.
Can you fix it?

PowerKiKi:
`src/PhpSpreadsheet/Cell.php` was moved to `src/PhpSpreadsheet/Cell/Cell.php` yesterday. So the file you are trying to modify no longer exists. You should rebase your branch onto latest develop and then force push. You can try that by yourself, or else I can give it a try, but not today...

src/PhpSpreadsheet/Cell.php
SailorMax: may be rebase was not so beauty, but done. :)

SailorMax:
Not beauty rebase, but done.

Now I trying to make it better in local branch, but don't really understand how to do it. Can you help me? Do I need shift split point to end of develop-branch? How?

SailorMax:
Looks like simple merge is better :)

PowerKiKi:
Thanks !

PR #259, opened 2017-10-23
setSize fix

gabidj:
setSize is allowing any type of value
setSize should only allow int or float
Added float `type hinting` for setSize.

In this case if user inputs (`24 pt`) will get a notice.
`Notice: A non well formed numeric value encountered in ../myproject/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Style/Font.php on line 261`
This will set the size 24. And will not generate a corrupt excel workbook.

> Note: if migrating to PHP `>7.0` return type `: float` can also be added to `getSize()` method

This is:

- [x] a bugfix
- [x] a new feature

Checklist:

- [x] Code style is respected

What does it change?

### DIFF
--- a/src/PhpSpreadsheet/Style/Font.php
+++ b/src/PhpSpreadsheet/Style/Font.php
@@ -257,9 +257,9 @@ class Font extends Supervisor implements IComparable
      *
      * @return Font
      */
-    public function setSize($pValue)
+    public function setSize(float $pValue = null)
     {
-        if ($pValue == '') {
+        if ($pValue === null) {
             $pValue = 10;
         }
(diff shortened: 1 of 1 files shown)


gabidj:
Unit tests not affected. 
`CHANGELOG` and `Documentation` updates are not necessary as this just protects from generating a corrupt excel workbook.

PowerKiKi:
Unfortunately we cannot use type hinting for scalar because we still support PHP 5.6.

gabidj:
I'm sorry for that. Type hinting is available from 5.0, but float `type hinting` is only available from 7.0 . 

Thanks for the note!

PR #261, opened 2017-10-23
Fix freeze pane for xlsx files

### LINKED ISSUE #260: Freeze pane should use xsplit/ysplit instead of topLeftCell
This is:

- [x] a bug report
- [ ] a feature request

The freezed pane should be the same after loading and saving an xlsx file

The freezed pane changes of place.

For example if I have a xlsx with this 
`<pane xSplit="5" ySplit="11" topLeftCell="GG646" activePane="bottomRight" state="frozen"/>`
It displays this in Excel and LibreOffice
xlsx

topLeftCell defines the bottom right pane to show but not the "frozen" part, we need to use xsplit/ysplit to define the frozen part. The current version uses topLeftCell, so with topLeftCell="GG646" you will not be able to navigate in your spreassheet because most rows/columns are frozen.

I don't think that this bug can't be reproduce without an example file.
By executing this script you will see that the frozen pane (top left one) will move.
[strange.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/1409074/strange.xlsx)

```php
<?php


$spreadsheet = $reader->load('strange.xlsx');
$writer = new \PhpOffice\PhpSpreadsheet\Reader\Xlsx($spreadsheet);
$writer->save("out.xls");
```

PhpSpreadsheet : 1.0.0-beta
php -v
PHP 7.1.2 (cli) (built: Feb 23 2017 19:07:37) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies

adriencohen:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

What does it change?
Fix #260 

Always use xsplit and ysplit to define the freeze pane


PowerKiKi:
Thank you for the contribution. Would it be possible to cover that with unit tests without adding an excel file to the repository ? Maybe by writing a file from scratch and then reading it ?

adriencohen:
After looking a bit about making a unit test I realized that I had to change Xlsx writer too.
Finally I feel a bit like the whole thing about freeze pane is a bit confusing. 

I think the main problem is that we don't make any difference between xsplit/ysplit and topLeftCell. I took a look at Apache POI and for freeze pane they have

`
/**
     * Creates a split (freezepane). Any existing freezepane or split pane is overwritten.
     * @param colSplit      Horizontal position of split.
     * @param rowSplit      Vertical position of split.
     */
    @Override
    public void createFreezePane(int colSplit, int rowSplit) {
        createFreezePane( colSplit, rowSplit, colSplit, rowSplit );
    }
`

`
 /**
     * Creates a split (freezepane). Any existing freezepane or split pane is overwritten.
     *
     * <p>
     *     If both colSplit and rowSplit are zero then the existing freeze pane is removed
     * </p>
     *
     * @param colSplit      Horizontal position of split.
     * @param rowSplit      Vertical position of split.
     * @param leftmostColumn   Left column visible in right pane.
     * @param topRow        Top row visible in bottom pane
     */
    @Override
    public void createFreezePane(int colSplit, int rowSplit, int leftmostColumn, int topRow) {
`
Where leftmostColumn and topRow represent topLeftCell. I'm not sure why they splitted it like that as in all Excel documentation it's always topLeftCell but anyway I think this way is more clear and will fix the bug I found.

Should I continue on this PR or close this one and open a new one ?

PowerKiKi:
You can keep pushing new commits to this PR, or squash and force push. Up to you.

PowerKiKi:
I see a BC break in the method `createFreezePane()`. Couldn't we have kept something compatible, such as: 

```php
public function freezePane($splitCell, $topLeftCell = null)
```
At first glance, it seems easier to use, give it a cell coordinate, instead of two integers.

Also `$topLeftCell` purpose is not immediately clear from the docblock. Is it something like the current scroll position of the freeze pane ? Maybe we could slightly reword the comment ?

adriencohen:
Ok for renaming the function but if we keep `$splitCell` it would mean that if we want to freeze the first 3 columns we would have to put `C0` which might be more confusing ? From the VBA documentation of Excel it seems like they are using "my method"

Well it's not the current position of the freeze pane, this one doesn't move. If when I load a file I have this : 
xlsx
GG646 is the topLeftCell
I don't really have any idea how to reword it to make it really clear. Office documentation says 

>  Location of the top left visible cell in the bottom right pane (when in Left-To-Right mode).
> [https://msdn.microsoft.com/en-us/library/office/documentformat.openxml.spreadsheet.pane.aspx](https://msdn.microsoft.com/en-us/library/office/documentformat.openxml.spreadsheet.pane.aspx)

PowerKiKi:
Thanks for your contribution. I reworked it a bit, to avoid breaking API, and merged it.

PR #276, opened 2017-11-17
Reader / Lazy init cells

SailorMax:
This is:

- [ ] a bugfix
- [ ] a new feature

Checklist:

- [ ] Code style is respected

What does it change?
  Add to Xlsx and Ods readers `lazyInitCells` option. It allow faster (30%) read large documents, because do not initialize each cell as internal object. Internal object will create on first access to the cell.
  This method can be helpful in cases when need to read few cells from large documents.

### DIFF
--- a/src/PhpSpreadsheet/Collection/Cells.php
+++ b/src/PhpSpreadsheet/Collection/Cells.php
@@ -50,6 +50,13 @@ class Cells
     private $index = [];
 
     /**
+     * An index of deferred created cells.
+     *
+     * @var array[]
+     */
+    private $lazyCellsData = [];
+
+    /**
      * Prefix used to uniquely identify cache data for this worksheet.
      *
      * @var string
@@ -95,7 +102,7 @@ class Cells
         }
 
         // Check if the requested entry exists in the index
-        return isset($this->index[$pCoord]);
+        return isset($this->index[$pCoord]) || isset($this->lazyCellsData[$pCoord]);
     }
 
     /**
@@ -407,6 +414,17 @@ class Cells
     }
 
     /**
+     * Add or update a lazy cell identified by its coordinate into the collection.
+     *
+     * @param string $pCoord Coordinate of the cell to update
+     * @param array $cellData Cell data to deferred create
+     */
+    public function addLazyCellData($pCoord, array $cellData)
+    {
+        $this->lazyCellsData[$pCoord] = $cellData;
+    }
+
+    /**
      * Get cell at a specific coordinate.
      *
      * @param string $pCoord Coordinate of the cell
@@ -425,6 +443,13 @@ class Cells
         // Return null if requested entry doesn't exist in collection
         if (!$this->has($pCoord)) {
             return null;
+        }
+
+        if (isset($this->lazyCellsData[$pCoord])) {
+            $cell = $this->parent->createNewPredefinedCell($pCoord, $this->lazyCellsData[$pCoord]);
+            $this->add($pCoord, $cell);
+            $this->storeCurrentCell();
+            unset($this->lazyCellsData[$pCoord]);
         }
 
         // Check if the entry that has been requested actually exists
--- a/src/PhpSpreadsheet/Reader/BaseReader.php
+++ b/src/PhpSpreadsheet/Reader/BaseReader.php
@@ -25,6 +25,14 @@ abstract class BaseReader implements IReader
     protected $readEmptyCells = true;
 
     /**
+     * Lazy init cells by first access or instantly
+     * Identifies when loaded cells will be initialized - instantly or on first access.
+     *
+     * @var bool
+     */
+    protected $lazyInitCells = false;
+
+    /**
      * Read charts that are defined in the workbook?
      * Identifies whether the Reader should read the definitions for any charts that exist in the workbook;.
      *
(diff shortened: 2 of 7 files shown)


PowerKiKi:
I had a quick look, and the idea seems interesting, but I have a few "high-level" questions before going into details.

- What are the downside of this lazy reading mechanism ? If there are no downsides, then shouldn't we remove the option (`setLazyInitCells()`) ? so it would be the only way to read a file ?
- If the option is kept, then it must be explained in end-user documentation
- We should try to limit the growth of `Worksheet.php` and the public API. So it would be best to not modify at all `Worksheet.php`, or at the very least, don't add new methods. Instead `lazyCreateNewCell()` and `createNewPredefinedCell()` should be move to the `Cells` class. So (lazy) cells management is centralized in a single place. (we could have something like `$sheet->getCellCollection()->createNewPredefinedCell()`
- Would you please explain why this is faster ? What part of the code during cell creation is slow ?
- What is the impact of this on memory consumption ? and how would it work (or not) with a custom cache for `Cells` ?

SailorMax:
- downside: if user need to get all data of imported cells, then lazy method can be slower because it made 1 additional step - temporary put imported data to temporary array:
  - regular method = read from xls -> create Cell object -> read value from Cell object
  - lazy method = read from xls -> put in temporary array -> create Cell object -> read value from Cell object
- I did not write about it in changelog/documentation because of waiting your verdict :)
- I not sure move these methods to `Cells` is correct:
  - We can try minimize `Cell.php`, because it really frequently using. But why `Worksheet.php`? In average Excel document has 3-10 worksheets. It is not hard for memory.
  - Worksheet has `createNewCell()` method => all similar methods has to be in it also, because they has similar logic. This can help in future, when anyone can want rewrite it.
  - `createNewPredefinedCell()` use `createNewCell()`. I not sure `$this->parent->createNewCell()` ideologically is correct.
  - `Cells` is just abstract collector of cells. It is not look inside of cells. But my methods use `setValue()`, `setXfIndex()`,..
- Creation of objects + `createNewCell()` is slow. Creating of arrays is faster. In our company I had to read documents with 10000+ rows of cells. Regular method takes 1.95s(import) + 0.26s(read require cells). Lazy method takes 0.96s(import) + 0.52s(read required cells). Difference = 33%!
- I don't see any additional memory consumption. Array takes smaller memory (compare to object) + it removes its item after created `Cell` object.
- It will work with custom cache, because I did not use it. Temporary array is inside `Cells` object.

SailorMax:
I think I'll remove `lazyCreateNewCell()` method... Will be only `createNewPredefinedCell()`. Wait, please next commit.

PowerKiKi:
> Array takes smaller memory

This is not true in recent versions of PHP: https://gist.github.com/nikic/5015323

Before merging this, we should come up with a benchmark script to measure both speed and memory.

SailorMax:
The article named "Why objects (usually) use less memory than arrays in PHP"
Keyword "usually". It not count time of constructors work.
My benchmark was my work document. Results was posted before (1.95s vs 0.96s on reading 10000+ rows with 8 column of cells = 80000+ cells). Tested on PHP 7.2

I not expert of your sources for moving `createNewCell()` to `Cells`. Can you do this? What to do with `$this->cachedHighestRow = max($this->cachedHighestRow, $aCoordinates[1]);`? `cachedHighestRow` is private property.

PowerKiKi:
I suppose `cachedHighestRow` could be moved to `Cells` too (and probably a few other related private things). This would indeed be a bit of work. But I think it would be for the best in the long term.

SailorMax:
So do you expect this "a bit of work" from me? :)

PowerKiKi:
I won't have time for that in the forseeable future. So if this feature is important to you, then yes you, or somebody else, should do that "bit of work" indeed. Sorry I was not more clear before.

SailorMax:
@PowerKiKi 
How about current status?
createNewCell() and few other method/properties now in Cells.
Is it good enough for merge? :)
Don't understand what I can do with Scrutinizer result. It doesn't like not my code...

criztovyl:
Have you found the issue Scrutinizer is complaining about? Anyway, here it is (image is a link):
[screenshot from 2018-03-23 12-45-31](https://scrutinizer-ci.com/g/PHPOffice/PhpSpreadsheet/inspections/15fab9d6-012c-4c8f-8435-195b68aa19fe/code-structure/class/PhpOffice%5CPhpSpreadsheet%5CCollection%5CCells)

Basically, it complains about moving the methods form `Worksheet` to `Cells`, it adds to much "complexity" to that class. But this is is what you where asked to do, so nothing wrong from your side, I guess.

If I get this correctly, either the moved (or new?) methods are simply too complex and should be refactored.
Or they are too complex in context of the `Cells` class, and that class and it's dependants need to be refactored. (While looking over the changes, I thought about maybe introducing a `CellFactory` :D)

I think the lines below (which you moved to `Cells`) are causing the issue.

https://github.com/PHPOffice/PhpSpreadsheet/blob/52ff262ddbdd709163f9ceaa62772f0f60d63698/src/PhpSpreadsheet/Reader/Xlsx.php#L963-L978

SailorMax:
@criztovyl
Possible, but I, as side developer, don't know what I can do in this case. This method gets list of options and I have to check each of them. How I can reduce 'ifs' in this case?
I can add move inside `->setCalculatedValue()`, for example, but in new commit it will get `$cellData['calculatedValue']` which can be undefined => I have to add `if (isset(...))` outside...

criztovyl:
I think you can't reduce the `if`s, but maybe put them somewhere else, somewhere downstream, in example into `Cell`, so Scruntinizer does not complain about it anymore. :D

PR #285, opened 2017-12-06
hotfix working date & weekend

thinhvoxuan:
This is:

- [x] a bugfix

What does it change?

Calculating working date is not correct.


PowerKiKi:
It seems this PR does not conform to code style, run `composer fix` to fix that.

Also please add new cases to unit tests for the specific bug you are trying to fix, and fix the existing one. 

And finally take some time to explain what was broken, and why this solution is better in the commit message.

oleibman:
WORKDAY is substantially changed since this PR. No reason to think it is not working, especially with no explanation or tests in this PR>

PR #292, opened 2017-12-15
Adding option to stop at a conditional styling, if it matches (only Xlsx-Files)

sbreiler:
This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [x] Code style is respected

What does it change?
Can read & write conditional cell formating with "stopIfTrue"-Option, which would prevent any further condition checks if matched. Only implemented for Xlsx-Files (Xlsx Writer/Reader).


src/PhpSpreadsheet/Style/Conditional.php
```diff
@@ -49,6 +52,13 @@ class Conditional implements IComparable
      */
     private $text;
 
+    /**
+     * Stop on this condition, if it matches
+     *
+     * @var int
```
PowerKiKi: Should be a boolean instead of leaking file format into our model

src/PhpSpreadsheet/Style/Conditional.php
```diff
@@ -28,6 +28,9 @@ class Conditional implements IComparable
     const OPERATOR_NOTCONTAINS = 'notContains';
     const OPERATOR_BETWEEN = 'between';
 
+    const IF_TRUE_STOP = 1;
+    const IF_TRUE_DONT_STOP = 0;
```
PowerKiKi: Should be dropped in favor of boolean

src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
```diff
@@ -491,6 +491,10 @@ private function writeConditionalFormatting(XMLWriter $objWriter, Phpspreadsheet
                         $objWriter->writeAttribute('text', $conditional->getText());
                     }
 
+                    if( $conditional->getStopIfTrue() === Conditional::IF_TRUE_STOP ) {
```
PowerKiKi: Fix code style with `composer fix` command

sbreiler: Sadly it seems i can't run composer fix on windows - are the three equal-signs a problem?

sbreiler:
@PowerKiKi can you have a look at the changes? As mentioned, i can't seem to run `composer fix` on my windows machine. I would add unit test, if everything else is fine.

src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
PowerKiKi: You can see a details of what wrong [on Travis](https://travis-ci.org/PHPOffice/PhpSpreadsheet/jobs/317967553#L597).

src/PhpSpreadsheet/Style/Conditional.php
```diff
@@ -28,6 +28,9 @@ class Conditional implements IComparable
     const OPERATOR_NOTCONTAINS = 'notContains';
     const OPERATOR_BETWEEN = 'between';
 
+    const IF_TRUE_STOP = true;
+    const IF_TRUE_DONT_STOP = false;
```
PowerKiKi: I meant those two `const` should be deleted entirely. There is no need to define a const for `true`.

sbreiler:
@PowerKiKi Could you have a look at the latest changes? Travis errored 'cause of some composer initialization problems...

PowerKiKi:
Thanks for your cooperation ! I tweaked it a little bit and merged it.

Issue #308, opened 2018-01-01
Missing Comments in HTML writer


criztovyl:
This is:
```
- [ ] a bug report
- [x] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

Cell comments are included in the HTML produced by the HTML writer.

### What is the current behavior?

Comments are not included.

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

$spreadsheet->getActiveSheet()->getCell("A1")->setValue("Without Comment");
$spreadsheet->getActiveSheet()->getCell("A2")->setValue("With Comment");

$spreadsheet->getActiveSheet()->getComment("A2")
    ->getText()->createTextRun("This is a comment.");

$writer = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($spreadsheet, "Html");
$writer->save("comments.html");

// Comments are missing

```

### Which versions of PhpSpreadsheet and PHP are affected?

All (feature request)

PR #310, opened 2018-01-01
Include comments in HTML writer (#308)

criztovyl:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:
- [x] Code style is respected

- To allow users to have Spreadsheet comments included in the HTML file generated by PhpSpreadsheet.
- #308

Submitted early to allow discussion.

HTML and CSS taken from LibreOffice HTML exporter.

### DIFF
--- a/docs/references/features-cross-reference.md
+++ b/docs/references/features-cross-reference.md
@@ -1231,14 +1231,14 @@
         <td style="text-align: center; color: orange;">●</td>
         <td style="text-align: center; color: orange;">●</td>
         <td style="text-align: center;">N/A</td>
-        <td></td>
+        <td style="text-align: center; color: orange;">● <sup>1</sup></td>
         <td style="text-align: center;">N/A</td>
         <td></td>
         <td></td>
     </tr>
     <tr>
         <td style="padding-left: 1em;">Rich Text</td>
-        <td style="text-align: center; color: red;">✖ <sup>1</sup></td>
+        <td style="text-align: center; color: red;">✖ <sup>2</sup></td>
         <td style="text-align: center; color: green;">✔</td>
         <td style="text-align: center; color: red;">✖</td>
         <td style="text-align: center; color: red;">✖</td>
@@ -1256,7 +1256,7 @@
     </tr>
     <tr>
         <td style="padding-left: 1em;">Alignment</td>
-        <td style="text-align: center; color: red;">✖ <sup>2</sup></td>
+        <td style="text-align: center; color: red;">✖ <sup>3</sup></td>
         <td style="text-align: center; color: red;">✖</td>
         <td style="text-align: center; color: red;">✖</td>
         <td style="text-align: center; color: red;">✖</td>
@@ -1568,5 +1568,6 @@
     </tr>
 </table>
 
-1. Only BIFF8 files support Rich Text. Prior to that, comments could only be plain text
-2. Only BIFF8 files support alignment and rotation. Prior to that, comments could only be unformatted text
+1. Only text contents
+2. Only BIFF8 files support Rich Text. Prior to that, comments could only be plain text
+3. Only BIFF8 files support alignment and rotation. Prior to that, comments could only be unformatted text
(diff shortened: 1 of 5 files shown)


criztovyl:
I am unsure about the unit test - there does not yet exists a test class for the HTML writer.

criztovyl:
Tests fail due to #311.

criztovyl:
The resulting HTML looks and behaves like this:

bildschirmfoto vom 2018-01-01 21-22-36

image


The PDF most likely will only show the red square in the cell. It won't break the PDF technically, but it would be bad UX.

A quick fix would be exclude them when writing a PDF. Or maybe it's possible to add the comments as real comments to the PDF.

I'll implement comments in the existing reader and check the existing tests to see how to create new ones :)

criztovyl:
I've added the comments to the reader and added the test :)

Furthermore I've disabled comments for PDF, as the default hiding via `display: none;` CSS does not work, at least not for mpdf.
For mpdf `title`-attributes can be converted to PDF annotations (comments), see [mpdf title2annots](https://mpdf.github.io/reference/mpdf-variables/title2annots.html). For tcpdf and dompdf I'm still searching how annotations from html work.

Additionally I've rebased my feature branch onto develop.

tests/PhpSpreadsheetTests/Functional/HtmlCommentsTest.php
```diff
@@ -0,0 +1,42 @@
...
+        return $this->spreadsheet->getActiveSheet()
+            ->getComment('A1')
+            ->getText()->createTextRun($this->value);
+    }
+
+    private function doTheTest($string)
+    {
+        $reloadedSpreadsheet = $this->writeAndReload($this->spreadsheet, 'Html');
+
+        $actual = (string) $reloadedSpreadsheet->getActiveSheet()->getComment('A1')->getText()->getPlainText();
```
PowerKiKi: `(string)` should not be needed here, since `getPlainText()` already returns a `string`

tests/PhpSpreadsheetTests/Functional/HtmlCommentsTest.php
```diff
@@ -0,0 +1,42 @@
+<?php
+
+namespace PhpOffice\PhpSpreadsheetTests\Functional;
+
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+
+class HtmlCommentsTest extends AbstractFunctional
+{
+    private $value = 'I am comment.';
```
PowerKiKi: Let's test a multiline comment, to test that it can be read read correctly.

PowerKiKi:
I did not mean that PDF should support comments, just not wreak havoc. So hiding them like you did would be enough, at least for now. But if you are willing to also implement comment for PDF, be my guest, but don't feel forced to do it.

I see you also use `<div>` instead of `<comment>` now. That seems better indeed.

A few minor changes in tests and we should be good to go. Unless you intent to implement PDF comments too...

tests/PhpSpreadsheetTests/Functional/HtmlCommentsTest.php
criztovyl: Hehe, that's an leftover before I used `getPlainText()` ^^

criztovyl:
If I take the time to implement comments as PDF annotations I'll do it in a different PR, to not mix features here :)

docs/references/features-cross-reference.md
```diff
@@ -1231,14 +1231,14 @@
         <td style="text-align: center; color: orange;">●</td>
         <td style="text-align: center; color: orange;">●</td>
         <td style="text-align: center;">N/A</td>
-        <td></td>
+        <td style="text-align: center; color: orange;">● <sup>1</sup></td>
```
criztovyl: I am currently thinking about changing to this:

bildschirmfoto vom 2018-01-02 21-10-00
(The second last column)

What do you think?

criztovyl: (Where <code><sup>2</sup></code> is the "only text contents" footnote)

criztovyl: @PowerKiKi Any thoughts on this? :)

PowerKiKi: You mean from `●` to `✖` ? I think `●` is the best fit according to legend "Partially supported"

criztovyl: Yup, that makes more sense. :)

PowerKiKi:
Thank you for the collaboration ! I made a few minor tweaks and merged it.

criztovyl:
Thanks for PhpSpreadsheet :)

PR #317, opened 2018-01-04
supress iconv error

wppd:
This is:
a bug fix

iconv will yield E_NOTICE error


PowerKiKi:
What is the input and output of `convertEncoding` without this PR ?

wppd:
oh, iconv will still returns gibberish string.

PR #321, opened 2018-01-08
Change array key from style to borderStyle

lukosius:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

This change is needed because 'style' array key was deprecated by 'borderStyle'.


PowerKiKi:
Thanks !

Issue #333, opened 2018-01-11
Support for autofitting the rows height


leandrogm:
This is:

```
- [ ] a bug report
- [X] a feature request
- [X] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

The lib should be able to have a function that autofits the rows based on the largest cell. I couldn't find any place to do it. 

After struggling with this issue I came up with a solution that I would like to share. That's why i'm creating this issue.

### What is the current behavior?

I'm not sure if it only happens on Libre Office Calc but if I have a text that has been wrapped before using `getAlignment()->setWrapText(true)` I'm was not able to fit the rows based on the new height of the cells, like when we double click excel/calc rows.

### Which versions of PhpSpreadsheet and PHP are affected?
"phpoffice/phpspreadsheet": "1.0.0-beta",
"php": ">=5.6.4"

ianfoulds:
Hi leandrodm. A good start but there are some errors in you code and it doesn't take into account merged cellls.

Here is a version of your method that can be placed into the client code and does not require your fork to implement the function.

Rather than use a row range, I've gone for performing the action on a single row but this could be adapted to your methodology easily. To get the row you need to do the following:

```
use PhpOffice\PhpSpreadsheet\Worksheet\Worksheet;
use PhpOffice\PhpSpreadsheet\Worksheet\Row;
use PhpOffice\PhpSpreadsheet\Cell\Cell;
use PhpOffice\PhpSpreadsheet\Cell\Coordinate;
...
    const DEFAULT_CELL_WIDTH = 9.14;
    const DEFAULT_ROW_HEIGHT = 15;
...
            $row = new Row($myWorksheet, $myRowNum);
            $this->autofitRowHeight($row);

```

```
    /**
     * Auto-fit the row height based on the largest cell
     *
     * @param Row $start
     * @return Worksheet
     */
    public function autofitRowHeight(Row $row, $rowPadding = self::ROW_PADDING)
    {
        $ws = $row->getWorksheet();
        $cellIterator = $row->getCellIterator();
        $cellIterator->setIterateOnlyExistingCells(true);

        $maxCellLines = 1;
        /* @var $cell Cell */
        foreach ($cellIterator as $cell) {
            $cellLength = strlen($cell->getValue());
[...]

PR #334, opened 2018-01-11
#333 Autofit the rows height in the interval based on the largest cell

leandrogm:
This is:

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [x] Code style is respected

I'm not sure if it only happens on Libre Office Calc but if I have a text that has been wrapped before using getAlignment()->setWrapText(true) I'm was not able to fit the rows based on the new height of the cells, like when we double click excel/calc rows.


criztovyl:
You should try running `composer run fix` to fix the code-style "errors". :)

leandrogm:
Thanks for the feedback guys, i'll work on the issues and get back to you.

Issue #348, opened 2018-01-28
Hard to extend functionality


jasverix:
This is:

```
- [ ] a bug report
- [ x ] a feature request
```

### What is the expected behavior?
I have some problems with formulas referencing columns in a table in excel. The formula looks like this: ```@[FirstName] & @[LastName]```.

I also have another problem where an Excel document exported from another system have invalid data zoomScaleNormal="0".

I have fixed both problems in an inherited class of Spreadsheet - but it is hard to implement, because the load() method for instance creates a ```new Spreadsheet()```. In simplexml_load_string() it is possible to specify classname. Is it possible to implement something like this? Or maybe a ```public static fuction setSpreadsheetClass(string $className)```

jasverix:
#349 - a proposal how it could be solved.

PowerKiKi:
It would probably better to fix bugs in the core, for everybody, rather than trying to fix them in a custom class that is actually hard to implement.

jasverix:
I agree... the problem is that my fix is too quick and should not be implemented in this project. But I will try to make PRs on the problems sometime soon.

PR #349, opened 2018-01-28
Use dynamic spreadsheet class

jasverix:
See issue #348 

This is:

```
- [ ] a bug report
- [ x ] a feature request
```

I have some problems with formulas referencing columns in a table in excel. The formula looks like this: ```@[FirstName] & @[LastName]```.

I also have another problem where an Excel document exported from another system have invalid data zoomScaleNormal="0".

I have fixed both problems in an inherited class of Spreadsheet - but it is hard to implement, because the load() method for instance creates a ```new Spreadsheet()```. In simplexml_load_string() it is possible to specify classname. Is it possible to implement something like this? Or maybe a ```public static fuction setSpreadsheetClass(string $className)```

### DIFF
--- a/src/PhpSpreadsheet/IOFactory.php
+++ b/src/PhpSpreadsheet/IOFactory.php
@@ -34,6 +34,8 @@ abstract class IOFactory
         'Mpdf' => Writer\Pdf\Mpdf::class,
     ];
 
+    private static $spreadSheetClass = Spreadsheet::class;
+
     /**
      * Create Writer\IWriter.
      *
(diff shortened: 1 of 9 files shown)


jasverix:
Just a proposal to how it could be solved

PowerKiKi:
This is unlikely to happen. Spreadsheet was never meant to be extended, and even if we wanted to go down that road, we should probably implement PSR-11 as a proper solution. But again, this probably won't happen anytime soon, because it will lead to issues for extending classes and harder maintenance work for us in the future.

PR #373, opened 2018-02-13
[docs] array short syntax

klimov-paul:
This is a doc fix, which applies array short syntax over PHPDoc and 'docs'

Since this library requires PHP >= 5.6 and uses array shortsyntax in its source code, its documentation should use it too.


PowerKiKi:
Thanks !

PR #383, opened 2018-02-19
add encode HTML option in HTML writer

moromete:
Added the feature to bypass the htmlspecialchars function when render as HTML.
This is useful if you want to put some html inputs instead of plain value to use the generated html as an excel form so it can be edited in browser or to attach some javascript event listeners (for example a right click for a context menu) The escaping of the text inside the input then became the responsibility of the input.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Html.php
+++ b/src/PhpSpreadsheet/Writer/Html.php
@@ -1274,7 +1274,7 @@ class Html extends BaseWriter
 
                         // Convert UTF8 data to PCDATA
                         $cellText = $element->getText();
-                        $cellData .= ($this->encodeHtml)?htmlspecialchars($cellText):$cellText;
+                        $cellData .= ($this->encodeHtml) ? htmlspecialchars($cellText) : $cellText;
 
                         if ($element instanceof Run) {
                             if ($element->getFont()->getSuperscript()) {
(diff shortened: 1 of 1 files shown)


PowerKiKi:
I don't think this should be merged because the suggested use case is a misusage of spreadsheet software in general. You would not write HTML in a cell in Microsoft Excel and then expect it to be interpreted. Instead, if you want to write forms, you should use the existing tools as defined in Excel formats (and variants).

Unfortunately forms are not supported in PhpSpreadsheet. But a PR to add proper support would be much more likely to be merged.

moromete:
The idea was to use all the features of PHPSpreadsheet formula calculation and html rendering that are more than ok the only feature I miss is to have the posibility to fill the excel data inside the browser and is blocked by the htmlspecialchars hardcoded in writeRow function. In this specific case I do not generate an xlsx so  I won't expect to be interpret it by Microsoft Excel since the excel object is created from php just html writer will be affected and the default case will still encode html data. . 
What is the tool that can use a phpSpreadsheet object and generate a html form that will look like an excel? Regards!
Regards !

PowerKiKi:
> What is the tool that can use a phpSpreadsheet object and generate a html form that will look like an excel?

It doesn't exist, see "Form Controls" [on that page](http://phpspreadsheet.readthedocs.io/en/develop/references/features-cross-reference/). But if somebody contribute it, I would merge it.

moromete:
And until that tool exists it would not be nice to have the possibility to use html writer instead ?

PowerKiKi:
No, because it goes against all spreadsheet software expectations, and it would be used as an excuse to avoid implementing the proper solution.

Issue #356, opened 2018-02-02
ISFORMULA() not recognized / throws exception


billynoah:
ISFORMULA() not recognized and throws exception.  Steps to reproduce.  Create a new spreadsheet.  In the cell A1 enter

    =ISFORMULA(B1)

The result should be `FALSE` and is valid.  Try loading the spreadsheet in PhpSpreadsheet and read cells like this:

```
$spreadsheet = \PhpOffice\PhpSpreadsheet\IOFactory::load($filename);
$rows = $spreadsheet->getSheet($sheet)->toArray();
```

Results in an exception:

> Sheet1!A1 -> undefined named range / function '_xlfn.ISFORMULA'

Related: https://github.com/PHPOffice/PHPExcel/issues/1136

[formula_test.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/1689719/formula_test.xlsx)

PowerKiKi:
I can reproduce it with:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

use PhpOffice\PhpSpreadsheet\Spreadsheet;

$spreadsheet = new Spreadsheet();
$sheet = $spreadsheet->getActiveSheet();
$cell = $sheet->getCell('A1');
$cell->setValue('=ISFORMULA(B1)');
var_dump($cell->getCalculatedValue());
```

Would you like to submit a PR to add support for `ISFORMULA` ?

PR #390, opened 2018-02-24
Add support for _xlfn. prefixed functions + implement ISFORMULA

Josh-G:
This is:
```
- [ ] a bugfix
- [x] a new feature
```

Checklist:
- [x] Code style is respected

This change adds support for newer functions that are prefixed by _xlfn. (#356). The calculation engine has been updated to recognise these as functions, and drop the _xlfn. part. I've added a couple of the new functions such as STDEV.S/P, MODE.SNGL, ISFORMULA. 

Whilst testing ISFORMULA, I see that loading and saving a spreadsheet that contains true or false values, without making any modifications causes the spreadsheet to contain integer 0 or 1 when loaded again in Excel. Could the following code block be added here https://github.com/PHPOffice/PhpSpreadsheet/blob/0084776160abcc2bd1d5be3012c7b9cf5edd7dd1/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php#L1058
```php
elseif (is_bool($calculatedValue)) {
    $objWriter->writeAttribute('t', 'b');
}
```
to resolve this?

Let me know if my PR needs amending.


Josh-G:
Thanks for the review. I wasn't sure under what circumstances it could be null? I've updated it to return a REF error instead and modified the test.

Any thoughts on the boolean save conversion I mentioned above?

PowerKiKi:
Thank you for this nice contribution.

For the boolean save conversion I suggest to create a new PR and it could be covered with functionnal tests, probably something similar to `tests/PhpSpreadsheetTests/Functional/FreezePaneTest.php`.

PR #391, opened 2018-02-26
add support to large excel file

MichaelTsao:
This is:

```
- [ ] add support to large excel file
```


PowerKiKi:
Could you please describe more what kind of file you are using ?

MichaelTsao:
When I process a large xlsx file, the _simplexml_load_string()_ function return a error:  

> internal error: Huge input lookup

This xlsx file has about 30 columns, 100'000 rows, some cells contain large text. The file is about 15MB, extracted xml content is about 50MB. Which make _simplexml_load_string()_ return error.

So I add  LIBXML_COMPACT | LIBXML_PARSEHUGE options to _simplexml_load_string()_ , then everything is working fine.

PowerKiKi:
I think this use case should be solved by using the already existing `\PhpOffice\PhpSpreadsheet\Settings::setLibXmlLoaderOptions()`, so the options are used everywhere in PhpSpreadsheet instead of an arbitrary location.

PR #394, opened 2018-02-26
Preserve calculated value type in Xlsx writer

Josh-G:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Fixes a bug causing TRUE/FALSE calculated values to be saved as an integer type in Xlsx writer (mentioned in #390). This PR ensures the correct boolean type attribute is set. The test case validates that the set of test values is correctly represented and unchanged after a write + load.


PowerKiKi:
Thanks I adapted slightly the tests and merged it.

Issue #404, opened 2018-03-04
setSelectedCells cannot be used with freezePane


LoranRendel:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
* When using `freezePane`, set the chosen cell active by default or do nothing
* `setSelectedCells` should work normally

### What is the current behavior?
* 1-st row below freezed line is scrolled out of view
* `setSelectedCells` doesn't work when using it before or after `freezePane`

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php
require_once "vendor/autoload.php";
try {
    $excelBook = new \PhpOffice\PhpSpreadsheet\Spreadsheet();
    $excelSheet = $excelBook->getActiveSheet();

    for ($i = 1; $i < 5; $i++) {
        for ($j = 1; $j < 5; $j++) {
            $excelSheet->getCellByColumnAndRow($i, $j)->setValue($i * 5 + $j);
        }
    }
    $excelSheet->freezePane('B2');
    $excelSheet->setSelectedCells('A1');

    $objWriter = new PhpOffice\PhpSpreadsheet\Writer\Xlsx($excelBook);
    $mime = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet";
    $output = 'file.xlsx';
    header("Content-Type: $mime");
    header("Content-Disposition: attachment;filename=\"$output\"");
    header("Cache-Control: max-age=0");
    $objWriter->save("php://output");
} catch (\PhpOffice\PhpSpreadsheet\Exception $e) {
}
```

### Which versions of PhpSpreadsheet and PHP are affected?
Tested on php 5.6 — 7.2, PhpSpreadsheet 1.1.0

PowerKiKi:
First point of this issue seems to be a duplicate of #389.

For the second part, `setSelectedCells()`, would you submit a PR to fix it ?

LoranRendel:
Yes, I will.

oleibman:
Fixed by PR #3841,

PR #405, opened 2018-03-04
Using setSelectedCells with freezePane in Xlsx writer

LoranRendel:
Issue #404


PowerKiKi:
Would you be able to add tests to cover that writer, and possibly others ? A good example for this case would be something similar to `\PhpOffice\PhpSpreadsheetTests\Functional\EnclosureTest`

LoranRendel:
Please check if anything wrong.

LoranRendel:
Please review my request.

PowerKiKi:
If you change your test value from `$selectedCell = 'A1';` to `$selectedCell = 'A2';`, then the tests will fail. It seems the patch is not working as expected. Would you be able to debug it further ?

oleibman:
Superseded by PR #3841.

Issue #375, opened 2018-02-14
 Reader\Xlsx loads one empty comment for each proper comment in spreadsheet


xklid101:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
For each comment load just one proper comment into sheet

### What is the current behavior?
When sheet has some comment, Xlsx reader loads two comments instead of one. The first has correct value and is placed in correct cell. The second one is empty and placed one cell left from proper comment

### What are the steps to reproduce?
Create some xlsx file with comment e.g. in D4, load file with Reader\Xlsx and get output. Someone will see 2 comments in sheet - one in D4 (this is ok), and one empty in C4 (this is wrong)

### Which versions of PhpSpreadsheet and PHP are affected?
branch develop, PHP 5.6

---
#### I believe the bug is here
https://github.com/PHPOffice/PhpSpreadsheet/blob/edae73b3e0d48b6411f64f42665fac01aa777ed3/src/PhpSpreadsheet/Reader/Xlsx.php#L1410

If I remove this:
```php
$comment = $docSheet->getCommentByColumnAndRow((string) $column, $row + 1);
```
and insert that instead:
```php
$comment = $docSheet->getCommentByColumnAndRow($column + 1, $row + 1);
```
everything is working and looks ok

madorin:
I have the same problem, just tried to update from am earlier beta build to latest stable one.

PowerKiKi:
Would one of you be able to submit a PR that would fix the bug and cover it with unit tests ?

xklid101:
I'll try to do it asap

xklid101:
Pullrequest to fix this issue is ready to merge (or some comment needed) 
https://github.com/PHPOffice/PhpSpreadsheet/pull/420

PR #420, opened 2018-03-15
Fix xlsx reader loading sheets with comments

xklid101:
Fix PHPOffice/PhpSpreadsheet#375

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

When opening xlsx file that contains comments with xlsx reader, phpspreadsheet worksheet contains one more empty comment on left next every original comment


xklid101:
Check my latest commits, I've done some changes. If anything else to be done, let mi know and I'll try to do it. Or post some comments... Thx

PR #431, opened 2018-03-21
Add notes for the Timezone functions in Date class

netpassprod1:
The \PhpOffice\PhpSpreadsheet\Shared\Date class has support for setting an alternate timezone which affects the conversion of Excel Date timestamps in PhpSpreadsheet. This amendment is an effort to document this support.

This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

The \PhpOffice\PhpSpreadsheet\Shared\Date class has functions that support using different timezones. This change provides documentation for how to update the default timezone as well as how to apply a different timezone via parameter for specific function calls.
I only discovered this capability by looking at the Date class source code.

### DIFF
--- a/docs/topics/calculation-engine.md
+++ b/docs/topics/calculation-engine.md
@@ -263,7 +263,7 @@ return an Excel date timestamp.
 Takes year, month and day values (and optional hour, minute and second
 values) and returns an Excel date timestamp value.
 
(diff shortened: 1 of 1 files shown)


PowerKiKi:
thanks!

PR #435, opened 2018-03-23
Better support of xlsm with forms

SailorMax:
This is:

```
- [x] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

Collect some currently not supported features of Xlsx when document loaded. And restore them when same document saved. This allow to fill Xlsm templates, which has form elements, macros, printer settings, protected elements and back compatibility drawing, and save result without lost important elements of document.


jbq:
I tried to incorporate this PR in my project because my form elements were discarded by PhpSpreadsheet, and now Excel crashes!

SailorMax:
@jbq Can you share sample of your document? You can erase all private data from it.

jbq:
@SailorMax Here are the details: https://www.dropbox.com/sh/xi9lrz42r8jctsd/AAAMCN0FanHKkVVFGrESFp2Za?dl=0

You will find the template XLSM file loaded by PhpSpreadsheet, and the one saved by the library with patch applied, without any extra processing.  The latter crashes Excel.

SailorMax:
@jbq Looks like fixed. Can you check latest update?

jbq:
@SailorMax sadly I had to quickly push to production and changed the form controls to plain cells with border styling to make it look like button, along with `BeforeDoubleClick` event handlers :-|  I like it even better than buttons :-)

PowerKiKi:
thanks for the contribution, it's been merged with a few tweaks

PR #438, opened 2018-03-27
Allow to pass explicit types to Worksheet::fromArray()

jbq:
Add optional array parameter 'explicitTypes'

This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [ ] Code style is respected

Currently `fromArray()` does not allow to set explicit types.  When the data in the array contain special characters such a leading equals sign that are meaningful to Excel, data is converted to formula, whereas I want to make sure it is typed as string.


PowerKiKi:
`fromArray()` is meant as a quick and simple use. If you have advanced need, this might not be the appropriate tool for your use-case.

I suggest you either use a custom value binder (which might increase performance if you pretty much hardcode a type), or that you fallback to set cell values individually.

PR #439, opened 2018-03-27
Update CONTRIBUTING guide

jbq:
Mention `php-cs-fixer`


PowerKiKi:
I don't think this is worth mentioning, because everything required is installed by composer. There is no need to single out php-cs-fixer.

jbq:
Indeed this makes sense, but then the scripts paths are wrong, see #441

PR #441, opened 2018-03-28
Use correct scripts paths for tools installed by composer

jbq:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [ ] Code style is respected

`composer fix` and `composer check` fail to run without this fix.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,12 +24,12 @@
     ],
     "scripts": {
         "check": [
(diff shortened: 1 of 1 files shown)


PowerKiKi:
This goes against composer documentation, https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands, and actual behavior seen locally and on Travis.

If this does not work for you, try updating composer.

PR #452, opened 2018-04-06
Best Pratices Fix

gabrielgmm:
This is:
Using logical operators such as or instead of || is generally not recommended.

```
- [x ] a bugfix
- [ ] a new feature
```

Checklist:

- [x ] Changes are covered by unit tests
- [ x] Code style is respected
- [ x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ x] CHANGELOG.md contains a short summary of the change

Best Pratices PSR - 2 fixed

### DIFF
--- a/samples/Basic/25_In_memory_image.php
+++ b/samples/Basic/25_In_memory_image.php
@@ -21,7 +21,7 @@ $spreadsheet->getProperties()->setCreator('Maarten Balliauw')
 
 // Generate an image
 $helper->log('Generate an image');
-$gdImage = @imagecreatetruecolor(120, 20) or die('Cannot Initialize new GD image stream');
+$gdImage = @imagecreatetruecolor(120, 20) || die('Cannot Initialize new GD image stream');
 $textColor = imagecolorallocate($gdImage, 255, 255, 255);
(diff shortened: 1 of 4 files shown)


samples/Basic/25_In_memory_image.php
```diff
@@ -21,7 +21,14 @@
 
 // Generate an image
 $helper->log('Generate an image');
-$gdImage = @imagecreatetruecolor(120, 20) or die('Cannot Initialize new GD image stream');
+
+try {
```
PowerKiKi: The try/catch won't work, because `imagecreatetruecolor` never throw anything.

src/PhpSpreadsheet/Calculation/Calculation.php
```diff
@@ -3097,7 +3097,7 @@ private static function mkMatrix(...$args)
      * @param string $formula
      * @param null|\PhpOffice\PhpSpreadsheet\Cell\Cell $pCell
      *
-     * @return bool
+     * @return array
```
PowerKiKi: This is should allow at least `array` and `false`. And maybe `void` as well, which is unfortunately a bit incorrect, so it might be best to leave it out.

PR #453, opened 2018-04-06
Fix issue with plain text and richtext being in the same cell

### LINKED ISSUE #442: RichText can not get the correct value, when one part is plain text and the other part is rich text.
This is:

```
- [x] a bug report
- [ ] a feature request
```

When I use RichText()->getPlainText(), I should get complete text, contain plain text and rich text.

In xlsx, Like cell value: "plain text, rich text", RichText()->getPlainText() only return the rich text part("rich text"). the first part is missing.

In source code https://github.com/PHPOffice/PhpSpreadsheet/blob/develop/src/PhpSpreadsheet/Reader/Xlsx.php, line 2070

```php
foreach ($is->r as $run) {
    if (isset($run->rPr)) {
        // ...
    }
}
```

missing the else part.

It should be
```php
if (isset($run->rPr)) {
    // ...
} else {
    $value->createText(StringHelper::controlCharacterOOXML2PHP((string) $run->t));
}
```
In Xlsx, a cell can contain part to plain text, and the other part of rich text.
In above, $run element can only have "t" attribute.

PhpSpreadsheet 1.2.0, PHP 7.1.6.

gintsmurans:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Fixes #442. Introduced in 0084776160abcc2bd1d5be3012c7b9cf5edd7dd1. I am not exactly sure why condition was removed, but returning it the way it was fixes the issue.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -2068,7 +2068,9 @@ class Xlsx extends BaseReader
         } else {
             if (is_object($is->r)) {
                 foreach ($is->r as $run) {
-                    if (isset($run->rPr)) {
+                    if (!isset($run->rPr)) {
+                        $objText = $value->createText(StringHelper::controlCharacterOOXML2PHP((string) $run->t));
+                    } else {
                         $objText = $value->createTextRun(StringHelper::controlCharacterOOXML2PHP((string) $run->t));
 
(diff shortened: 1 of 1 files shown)


gintsmurans:
Ha ha, I spent 20 minutes to find the cause of the issue, and just now saw that @expbenson had already found where the issue lays. Anyways, if you find this not relevant, just close the pull request.

PR #455, opened 2018-04-06
Fix add a maximum column width parameter for Worksheet and check it for autoSize

Aketos:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

If the autoSize() calculate a width for a column superior to 255 (maximum Excel width column) the column size is set at minimum width instead of maximum

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


Aketos:
I understand, however, the fix is on an old big function, which is not already unit tested.
I don't know if I will have time to do it for you, I will see

oleibman:
Superseded by PR #4581. Removing stale label, leaving closed.

Issue #244, opened 2017-10-10
Add streaming mode for writing excel files


jontro:
Currently PhpSpreadsheet buffers the xlsx until it's written out to a stream which will be a problem if the generated dataset is very large.

Apache POI has a streaming implementation (https://poi.apache.org/spreadsheet/) where it solves this by restricting the operations if using it in a streaming mode.

- Only a limited number of rows are accessible at a point in time.
- Sheet.clone() is not supported.
- Formula evaluation is not supported

Would it be possible to add something similar to PHPSpreadsheet?

PowerKiKi:
Do you mean create a spreadsheet from scratch, in-memory, and while you add new cells/rows, some "previous cells" are written down on disk and deleted from memory ? Keeping the memory usage very low at all time ?

jontro:
I was thinking that you could open a file pointer to php://output right away so the contents will start being served immediately.
I.e. 


    $objWriter = PHPExcel_IOFactory::createWriter('php://output', 'Excel2007');
    $objPHPExcel = new PHPExcel($objWriter);
    while ($result = $stmt->fetch(PDO::FETCH_ASSOC)) {
        $row++;
        $objPHPExcel->getActiveSheet()->setCellValueByColumnAndRow(1, $row, $result['column']);
    }
    ...

For very large datasets php will run out of memory if it's kept in buffer all the time

PowerKiKi:
This would most likely be a huge undertaking and Mark's input might be needed. Would you be willing to work on that?

jontro:
I would absolutely be willing to help but I think it would have to be coordinated with multiple people. I have no real experience with working with the xlxs implementation details. What do you think? If it's a huge implementation change I wonder if it's really worth it, depending on how many library users would think this is useful. 

Especially since the mode would limit other aspects of the api

PowerKiKi:
I agree that having multiple person involved might be best. Unfortunately I won't have enough time for that myself. So if you can find others, or invest enough time by yourself, then go for it. Otherwise I guess we'll have to do without that feature.

@MarkBaker any opinion on the matter ?

glenndavey83:
I could really use this right about now :/

felansu:
I too my bro... 💃

eselskas:
me too

xKZL:
same :(

fcoded:
:(

PowerKiKi:
[...]

Issue #28, opened 2016-10-18
Get a resource from Writers for PSR-7 integration


lucascorbeaux:
Hi there,

I want to suggest a feature for the Writers. Not easy to implement if I trust the current codebase, but pretty helpful for many integrations.

Modern PSR-7 frameworks needs to work with, well, PSR-7 Response objects. These objects use Stream objects that rely on resources to fetch data.

In the current version of PhpSpreadsheet, it's only possible to write on a file or php://output / php://stdout. It's quite tricky to get a resource this way, and writing to a temporary file could not be a viable option on many application architecture.

Adding an abstract getResource method to BaseWriters would be a great improvement for PSR-7 frameworks integration.

PowerKiKi:
I am not familiar with the details of PSR-7 yet, and can't find much related to "resource" in [the spec](http://www.php-fig.org/psr/psr-7/). Would you care to detail what you had in mind ? Would it be an implementation of `Psr\Http\Message\ResponseInterface` ? or usage of that interface ?

lucascorbeaux:
Hi, and sorry I probably wasn't clear. The resource isn't at all part of PSR-7 specs, I was talking about the PHP resource type : http://php.net/manual/en/language.types.resource.php

In this case, it will be a stream resource. What could be a great add to PhpSpreadsheet is to provide a getResource method that returns the stream resource of the current document. If document is provided by stream, it's still possible to use a php://memory stream, for example.

Behing able to retrieve this resource will be really useful especially for PSR-7 integration (PSR-7 Responses need a PHP resource to generate a response body).

PowerKiKi:
My take on this would be to modify the signature of `PhpOffice\PhpSpreadsheet\Writer\IWriter::save()` to always return the file descriptor of the written file. Then it's up to you to do whatever you need with that file descriptor. The `save()` method already supports writing to `php://output` (via temp files for some writers), so we only need to add support for `php://memory`, which should not be too difficult hopefully.

This would be for convenience only and would not bring any gain of performance, because we would still be creating temp files on disk.

So usages would be something like:

```php
$writer = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet);

$writer->save('/some/path/to/file.xlsx'); // standard, save to disk
$writer->save('php://output');            // to download the file in the browser
$handle = $writer->save('php://memory');  // to get a file descriptor to the stream in memory
```

Would this solve your use case ?

[...]

PR #473, opened 2018-04-24
allow return in save function param

eltharin:
set first param to return to get the content instead of put in a file

This is:

- [ ] a bugfix
- [ X] a new feature

Checklist:

    Changes are covered by unit tests
    [ X] Code style is respected
    [ X] Commit message explains why the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
    CHANGELOG.md contains a short summary of the change
    Documentation is updated as necessary

Why this change is needed?

insted of make :

$Excel->save($tempfile);
$content = file_get_content($tempfile);
unset($tempfile);

We have just :

$content = $Excel->save('return');

Very usefull when we want to add many Excel Files to a unique zip.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Xlsx.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx.php
@@ -172,13 +172,14 @@ class Xlsx extends BaseWriter
      */
     public function save($pFilename)
     {
+        $return = null;
         if ($this->spreadSheet !== null) {
             // garbage collect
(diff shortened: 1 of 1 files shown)


PowerKiKi:
This is somewhat similar to #244 and #28. And my comments from there still stand. This should be a change across all writers and should avoid any disk write to provide real benefit, instead of just a quick-and-dirty convenience.

PR #474, opened 2018-04-27
Fix format on conditional formatting

MaximePS:
This is:

- [x] a bugfix
- [ ] a new feature

## Why this change is needed?

If you read a xlsx file with specific format (eg: percent) on conditional formatting, the format is ereasing.


oleibman:
Equivalent change was made by PR #1296 in Dec. 2019.

Issue #468, opened 2018-04-18
Xlsx Reader doesn't use readEmptyCells


thisispaul:
This is:

```
- a bug report
```

### What is the expected behavior?
When I set `readEmptyCells` to `false` on a Xlsx Reader and then call `$spreadsheet->getActiveSheet()->toArray()` I would expect an array with all the values in the sworksheet.

### What is the current behavior?
When I call `$spreadsheet->getActiveSheet()->toArray()` I get a huge array [~1000 rows] where most of the rows are filed with `null` arrays (in my example everything after row 100 is null).

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$reader = IOFactory::createReaderForFile($path);
$reader->setReadDataOnly(true);
$reader->setReadEmptyCells(false);
$spreadsheet = $reader->load($path);
$data = $spreadsheet->getActiveSheet()->toArray();
```

### Which versions of PhpSpreadsheet and PHP are affected?
`phpoffice/phpspreadsheet: 1.2.1`
`php: 7.1.15`

gabrielgmm:
Hi ThisPaul,

I have been tested with a xlsx with  this file:
https://drive.google.com/open?id=1ka_KpKPJhIQqfAgRvNDgzMrS-wKh1tub
I installed in same directory of teste file

I used a PHP 7.1.11 and 7.1.16 

My code
[$inputFileName = "teste.xlsx";
$inputFileType = \PhpOffice\PhpSpreadsheet\IOFactory::identify($inputFileName);
$reader = \PhpOffice\PhpSpreadsheet\IOFactory::createReader($inputFileType); // teste pass ok
$reader = \PhpOffice\PhpSpreadsheet\IOFactory::createReaderForFile($inputFileName); // teste pass ok
$reader->setReadDataOnly(true);
$reader->setReadEmptyCells(true);
$spreadsheet = $reader->load($inputFileName);
$data = $spreadsheet->getActiveSheet()->toArray();](url)

thisispaul:
yes yours seems to work.

I think something must be going on with the file I have. It was downloaded from google drive and has about 21 different sheets in it. I tried a different file download from google drive with much less data and that seemed to work fine. Would having loads of sheets in the file cause any issues you are aware of?

zaifeng:
[...]

PR #479, opened 2018-04-30
Add cache entry for highest row and column

monofone:
this prevents to much calls to expensive sscanf and strlen

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

The calls to Cell::getHighestRowAndColumn() are extremely expensive and increase the runtime on spreadsheets for reading.
This does not directly belong to #468 but was inspired by finding when debugging this. The PR is just to start the discussion about the performance in this part.


monofone:
The failing test seems to have side effects as running then on its own it does not fail.Any suggestions what to look for ?

PR #486, opened 2018-05-03
Fix #485 Missing required parameters

### LINKED ISSUE #485: Required function parameters are not set (Error in PHP7)
This is:

```
- [x] a bug report
```

### Problem
PHP 7 throws an [_ArgumentCountError_](https://3v4l.org/dnfYQ) if a function call doesn't set all required parameters.

### Affected files

- Worksheet.php _(missing 1 argument)_
  > https://github.com/PHPOffice/PhpSpreadsheet/blob/36acc372875c4d894dc093825ce4f62209db5a76/src/PhpSpreadsheet/Writer/Xls/Worksheet.php#L830
  > https://github.com/PHPOffice/PhpSpreadsheet/blob/36acc372875c4d894dc093825ce4f62209db5a76/src/PhpSpreadsheet/Writer/Xls/Worksheet.php#L654
- PolynomialBestFit.php _(missing 3 arguments)_
  > https://github.com/PHPOffice/PhpSpreadsheet/blob/36acc372875c4d894dc093825ce4f62209db5a76/src/PhpSpreadsheet/Shared/Trend/PolynomialBestFit.php#L167
  > https://github.com/PHPOffice/PhpSpreadsheet/blob/36acc372875c4d894dc093825ce4f62209db5a76/src/PhpSpreadsheet/Shared/Trend/BestFit.php#L304

branch develop on PHP 7

KoenigsKind:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

This fixes issue #485.

I don't know if those are the proper values, but those are the ones PHP5 would have used till now.

### DIFF
--- a/src/PhpSpreadsheet/Shared/Trend/PolynomialBestFit.php
+++ b/src/PhpSpreadsheet/Shared/Trend/PolynomialBestFit.php
@@ -164,7 +164,7 @@ class PolynomialBestFit extends BestFit
         $this->intersect = array_shift($coefficients);
         $this->slope = $coefficients;
 
-        $this->calculateGoodnessOfFit($x_sum, $y_sum, $xx_sum, $yy_sum, $xy_sum);
+        $this->calculateGoodnessOfFit($x_sum, $y_sum, $xx_sum, $yy_sum, $xy_sum, 0, 0, 0);
         foreach ($this->xValues as $xKey => $xValue) {
             $this->yBestFitValues[$xKey] = $this->getValueOfYForX($xValue);
         }
--- a/src/PhpSpreadsheet/Shared/Trend/PolynomialBestFit.php
+++ b/src/PhpSpreadsheet/Shared/Trend/PolynomialBestFit.php
@@ -164,7 +164,7 @@ class PolynomialBestFit extends BestFit
         $this->intersect = array_shift($coefficients);
         $this->slope = $coefficients;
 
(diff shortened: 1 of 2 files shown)


criztovyl:
AFAIKT the values are all numeric, so a default value (`0`) in the parameter declaration should be enough, compare <https://3v4l.org/3krEW>.

criztovyl:
Or `null` instead of `0` if that's the better default, which depends on the usage of the parameter further down the stack.

KoenigsKind:
`In PHP5 the default value for a not set parameter is .`
wow, I'm not used to use git on a command line; seems like bash replaces the string `null` with a null character o.O

> Or `null` instead of `0` if that's the better default, which depends on the usage of the parameter further down the stack.

That's what I'm thinking too, it's better to let PHP handle the type conversion. Though there might be some better values to use, if somebody knows what those functions actually do.

criztovyl:
`$xfIndex` for `writeString` is a short:
https://github.com/PHPOffice/PhpSpreadsheet/blob/36acc372875c4d894dc093825ce4f62209db5a76/src/PhpSpreadsheet/Writer/Xls/Worksheet.php#L710

The missing values to `calculateGoodnessFit()` are numbers:
https://github.com/PHPOffice/PhpSpreadsheet/blob/36acc372875c4d894dc093825ce4f62209db5a76/src/PhpSpreadsheet/Shared/Trend/BestFit.php#L304-L325

PowerKiKi:
I ended up merging your first commit, as recommended recommended by @criztovyl, because it seems to best option we have so far. Thanks to both of you.

Issue #492, opened 2018-05-08
CSV reader – backslash is always treated as an escape character


rdarcy1:
This is:

```
- [x] a bug report
- [ ] a feature request
```

### What is the expected behavior?
When required, the backslash should be treated as a normal character and not parsed as an escape character.

### What is the current behavior?
The backslash acts as an escape character, so when at the end of a cell/field it escapes the closing quote and merges cells/lines.

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// CSV file with backslash at end of field
$contents = '"test field\"' . "\n" . '"another field"';
$filename = __DIR__ . '/test.csv';
file_put_contents($filename, $contents);


// Create a new reader
$reader = new PhpOffice\PhpSpreadsheet\Reader\Csv;

$spreadsheet = $reader->load($filename);

var_dump($spreadsheet->getActiveSheet()->getCell('A1')->getValue());
```

Output:
```
string(27) "test field\"
another field""
```

I know the backslash is commonly used as an escape character, but it would be good to at least have an option of turning this off, maybe with a `$reader->setEscaping(false)` method call. I'm currently dealing with third-party CSVs which contain backslashes (a poor design choice, but considering there's no universal standard I think it's perfectly valid).

[RFC 4180](https://tools.ietf.org/html/rfc4180) makes no mention of backslashes, using instead a quote character to escape quote characters.

Happy to put together a pull request if this change is deemed worthwhile.

PowerKiKi:
This seems to make sense. As you probably figured out we use PHP native [fgetcsv](http://php.net/manual/en/function.fgetcsv.php) and we don't expose its `$escape` parameter. So we you can create a PR for that is covered with unit tests, I'll merged it.

PR #510, opened 2018-05-21
PHPOffice/PhpSpreadsheet#492 Allow CSV escape character to be set

rdarcy1:
From issue #492.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Allows the escape character of the CSV reader to be set, enabling the parsing of files that include the default escape character.


PowerKiKi:
Thanks !

rdarcy1:
No problem, thanks for merging :)

PR #521, opened 2018-05-30
Fix #519 Prevent infinite loop when getting references from invalid range

### LINKED ISSUE #519: Invalid cell coordinate range causes infinite loop
This is:

```
- [x] a bug report
- [ ] a feature request
```

An exception should be thrown for an invalid range, or maybe just the start cell should be returned.

The program gets stuck in a loop and PHP runs out of memory.

I think the offending loop is \PhpOffice\PhpSpreadsheet\Cell\Coordinate line 372

```php
while ($currentCol != $endCol) {
    ...
}
```

Probably needs a magnitude comparison rather than a straight equality comparison, since if an invalid range is given the current column will never equal the end column.

Please provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

// Gets stuck in a loop
\PhpOffice\PhpSpreadsheet\Cell\Coordinate::extractAllCellReferencesInRange('Z1:A1');


// Which means the following gets stuck, causing a hard to debug memory error for
// code that isn't necessarily obviously wrong.
$spreadsheet->getActiveSheet()
    ->mergeCellsByColumnAndRow(0, 1, 1, 1);
```

PhpSpreadsheet 1.2

rdarcy1:
Fixes #519 

Didn't realise it was such an easy fix..! Currently it returns an empty array of references when passed an invalid range, not sure if an exception would be more suitable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `Helper\Html` support UTF-8 HTML input - [#444](https://github.com/PHPOffice/PhpSpreadsheet/issues/444)
 - Xlsx loaded an extra empty comment for each real comment - [#375](https://github.com/PHPOffice/PhpSpreadsheet/issues/375)
 - Xlsx reader do not read rows and columns filtered out in readFilter at all - [#370](https://github.com/PHPOffice/PhpSpreadsheet/issues/370)
+- `Coordinate::ExtractAllCellReferencesInRange()` returns empty array instead of looping when passed an invalid range – [#519](https://github.com/PHPOffice/PhpSpreadsheet/issues/519)
 
 ## [1.2.1] - 2018-04-10
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - `Helper\Html` support UTF-8 HTML input - [#444](https://github.com/PHPOffice/PhpSpreadsheet/issues/444)
 - Xlsx loaded an extra empty comment for each real comment - [#375](https://github.com/PHPOffice/PhpSpreadsheet/issues/375)
 - Xlsx reader do not read rows and columns filtered out in readFilter at all - [#370](https://github.com/PHPOffice/PhpSpreadsheet/issues/370)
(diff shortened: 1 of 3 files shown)


rdarcy1:
I've updated to throw an exception on invalid range and amended the unit test to match.

criztovyl:
Try executing `./vendor/bin/php-cs-fixer fix` to fix the remaining code-style issues that Travis does not like.

Regarding Scrutinizer I am not sure, but last time PowerKiki suggested to extract the added code into it's own method, maybe try that. :)

Thx!

criztovyl:
In this case just your added code in `extractAllCellReferencesInRange()` would need to be extracted.

rdarcy1:
I've also refactored the large `extractAllCellReferencesInRange` method down to a couple of smaller methods and extracted a couple of duplicated lines to improve the code rating – hopefully that's not outside the scope of the original PR?

I think much of the logic concerning cell ranges and references could be extracted to their own objects rather than dealing with range and reference arrays if you were open to cleaning up the class further.

criztovyl:
By personal preference I would have only extracted as much as is required to pass CI and included everything else in a additional PR. :)
But that seems to be exactly what you did ^^

Could provide an additional PR for your suggested refactorings? Otherwise you could open an a feature request issue.

Thanks for your contributions so far. :)

PowerKiKi:
Thanks ! The refactoring is always welcomed. The existing code tends to use way too big methods...

boredom2:
Hi there,
we currently experience Tons of Problems with this Change :)
After a quick Investigation, I suppose, that $startCol>=$endCol wrongly fails in the case "B" >= "AG".
Obviously, "B" is smaller than "AG", but due to PHPs String Comparison, it returns false and therefore wrongly emits the Exception.

rdarcy1:
Ah, sorry I wasn't thorough enough with my tests then. I can re-submit with a better comparison function.

PowerKiKi:
@boredom2, I can confirm what you say. It was merged too quickly. It was also reported in #545, and I just pushed a fix on develop branch. Can you confirm it fixes the issue for you ?

PowerKiKi:
It's been released as [1.3.1](https://github.com/PHPOffice/PhpSpreadsheet/releases/tag/1.3.1)

PR #527, opened 2018-06-02
Fix #526: Fixed parsing of conditionals in COUNTIF functions

### LINKED ISSUE #526: COUNTIF() does not properly parse comparison operators
This is:

```
- [x] a bug report
- [ ] a feature request
```

COUNTIF() should properly parse comparison expressions in the selection criteria parameter.

PhpSpreadsheet is generating the error `Formula Error: Unexpected operator '<'` when it tries to evaluate the function:

```
COUNTIF(F12,"<>< Please Select >")
```

The underlying issue is that method `Calculations\Functions::ifCondition()` is not properly parsing the conditional operator in the front of the selection parameter.   For example, it thinks that the selection parameter above means:

```
 <>< "PLEASE SELECT >"
```

when it should mean

```
<> "< PLEASE SELECT >"
```

This problem is due to an overly aggressive regular expression that matches any length of '<', '>', and '=' characters, when it should be limited to just '=', '<', '>', '<>', '<=', or '>=', (see section 18.17.2.2 in [ECMA 376 5th Edition Part 1](https://www.ecma-international.org/publications/standards/Ecma-376.htm) for all allowable comparison operators in Open Office Spreadsheet XML formulas.)

Other statistical functions that use ifCondition() also probably suffer from this issue.  The MAXIF(), MINIF(), and AVERAGEIF() functions use ifCondition() as well.

Calling `Calculations\Functions::ifCondition()` on formula `COUNTIF(F12,"<>< Please Select >")` will generate an exception with message `Formula Error: Unexpected operator '<'`

This occurs in 1.2.1 of PhpSpreadsheet.  I haven't checked earlier versions.  I did not see this in PhpExcel.

billblume:
This is:

```
- [ x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

PhpSpreadsheet generates an exception when one tries to write
a Xlsx spreadsheet with the following formula

```
COUNTIF(F12,"<>< Please Select >")
```
The underlying cause is that it problems parsing conditionals when there are
'<', '>', or '=' characters after the conditional.


billblume:
Closing this pull request, since it was contaminated with my pull request for issue 523 and the revert in the commits prevents the both of them to be merged together.  I'll provide an updated pull request later.

PR #558, opened 2018-06-21
Fix #557: Handle escaped spaces in date format would cause incorrect AM/PM conversion

### LINKED ISSUE #557: Error in AM/PM conversion when using an escaped space in custom format
This is:

```
- [x] a bug report
```

Dates with escaped spaces in their format should properly handle AM/PM conversion. 

Dates formats escaped with a backslash `\` do not handle AM/PM conversion correctly.

```
diff --git a/tests/data/Style/NumberFormatDates.php b/tests/data/Style/NumberFormatDates.php
--- a/tests/data/Style/NumberFormatDates.php
+++ b/tests/data/Style/NumberFormatDates.php
@@ -52,4 +52,14 @@ return [
         22269.0625,
         '[DBNum1][$-804]m"月"d"日";@',
     ],
+    [
+        '07:35:00 AM',
+        43270.315972222,
+        'hh:mm:ss\ AM/PM'
+    ],
+    [
+        '02:29:00 PM',
+        43270.603472222,
+        'hh:mm:ss\ AM/PM'
+    ]
 ]
```

See test cases above.

Current

See https://github.com/PHPOffice/PhpSpreadsheet/pull/558/files for possible solution.

deekthesqueak:
This is:

```
- [x] a bugfix
```

Checklist:

- [x] Code style is respected

This custom formatting commonly shows up in exported timesheet data from a large company.

When the escaped space isn't kept it will lead this to the format array being exploded into 3 pieces. Then at https://github.com/PHPOffice/PhpSpreadsheet/blob/1.3.1/src/PhpSpreadsheet/Style/NumberFormat.php#L468 the check only only take the firs block which is 'hh:mm:ss' so it assumes 24 hour but then later in the loop you'll have 'am/pm' which gets converted into A. When all is said and done you can get AM/PM on a 24 hour value.


PowerKiKi:
Thanks !

PR #575, opened 2018-07-01
Change usage of boolean to check delimiter

tygalive:
Instead of checking whether the delimiter has already been set, made it part of the line

This is:

- [ ] performance & bugfix

Checklist:

- [ ] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Writer/Csv.php
+++ b/src/PhpSpreadsheet/Writer/Csv.php
@@ -312,31 +312,27 @@ class Csv extends BaseWriter
      */
     private function writeLine($pFileHandle, array $pValues)
     {
-        // No leading delimiter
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Writer/Csv.php
```diff
@@ -312,31 +312,27 @@ public function setSheetIndex($pValue)
      */
     private function writeLine($pFileHandle, array $pValues)
     {
-        // No leading delimiter
-        $writeDelimiter = false;
+        // For no leading delimiter
+		$_delimiter = "";
```
PowerKiKi: underscore should not be used

oleibman:
Equivalent change was made by PR #1518.

PR #586, opened 2018-07-06
Only migrate PHP files, and exclude the vendor folder.

### LINKED ISSUE #481: Migration faux pas
This is:

```
- [x] a bug report
- [ ] a feature request
```

The migration tool does a number of somewhat dumb, unexpected things.

As far as it can, the migration tool should make the changes necessary to move from PHPExcel to PHPSpreadsheet, *but no more*.

The migration makes some changes that are inadvisable and easily avoided.

It appears to do a blanket global replacement of `PHPExcel` with `\PhpOffice\PhpSpreadsheet\Spreadsheet`, but this includes places that this change should not be made, causing breakages. For example, I had a private property that held a PHPExcel instance, and it did this:

    private $PHPExcel;

became:

    private $\PhpOffice\PhpSpreadsheet\Spreadsheet

Which is not even valid PHP.

The instructions for the migration suggest running it in the project root, which includes the vendor folder - so the migration is applied to the installed PHPExcel package, and also makes changes to PHPSpreadsheet package wherever it happens to mention PHPExcel.

The migration should not touch anything in a `vendor` folder.

You don't even need any code to demonstrate this; an empty project is enough:

```sh
composer init
composer require phpoffice/phpexcel
composer require phpoffice/phpspreadsheet
vendor/phpoffice/phpspreadsheet/bin/migrate-from-phpexcel
```

The output shows that it has made changes to both the PHPExcel and PHPSpreadsheet files (excerpt):

```sh
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpexcel/install.txt converted
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpexcel/Classes/PHPExcel.php converted
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpexcel/Classes/PHPExcel/Autoloader.php converted
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpexcel/Classes/PHPExcel/CachedObjectStorageFactory.php converted
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpexcel/Classes/PHPExcel/Calculation.php converted
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpexcel/Classes/PHPExcel/Cell.php converted
...
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpspreadsheet/docs/topics/migration-from-PHPExcel.md converted
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpspreadsheet/samples/Basic/46_ReadHtml.php converted
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Helper/Migrator.php converted
/Users/marcus/Sites/pxtest/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Ods.php converted
/
```
Note that this also applies the replacements in non-PHP files, which is quite unnecessary.

I'm using PHPExcel 1.8.1 and PHPSpreadsheet 1.2.1. It probably affects others.

As far as I can see, the current blanket search and replace mechanism is extremely unlikely to produce working code, and probably causes more problems than it solves. The matches and replacements could be made much more specific, or better still, be based on an AST search rather than simple text.

blunket:
Fixes #481

This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

See #481

### DIFF
--- a/src/PhpSpreadsheet/Helper/Migrator.php
+++ b/src/PhpSpreadsheet/Helper/Migrator.php
@@ -248,6 +248,9 @@ class Migrator
     private function recursiveReplace($path)
     {
         $patterns = [
+            '/*.md',
+            '/*.txt',
+            '/*.TXT',
             '/*.php',
             '/*.phpt',
             '/*.php3',
@@ -262,6 +265,7 @@ class Migrator
         foreach ($patterns as $pattern) {
             foreach (glob($path . $pattern) as $file) {
                 if (strpos($path, '/vendor/') !== false) {
+                    echo $file . " skipped\n";
                     continue;
                 }
                 $original = file_get_contents($file);
--- a/src/PhpSpreadsheet/Helper/Migrator.php
+++ b/src/PhpSpreadsheet/Helper/Migrator.php
@@ -248,11 +248,12 @@ class Migrator
     private function recursiveReplace($path)
     {
         $patterns = [
-            '/*.md',
             '/*.php',
+            '/*.phpt',
+            '/*.php3',
+            '/*.php4',
+            '/*.php5',
             '/*.phtml',
-            '/*.txt',
-            '/*.TXT',
         ];
 
         $from = array_keys($this->getMapping());
@@ -260,6 +261,9 @@ class Migrator
 
         foreach ($patterns as $pattern) {
             foreach (glob($path . $pattern) as $file) {
+                if (strpos($path, '/vendor/') !== false) {
+                    continue;
+                }
                 $original = file_get_contents($file);
                 $converted = str_replace($from, $to, $original);
 


blunket:
Just had a thought:  If it'd be good, we can add "Skipping file '/vendor/...' " to the output right before the `continue` statement I've added.

Let me know.

src/PhpSpreadsheet/Helper/Migrator.php
```diff
@@ -248,18 +248,22 @@ public function getMapping()
     private function recursiveReplace($path)
     {
         $patterns = [
-            '/*.md',
```
PowerKiKi: non-PHP files should not be removed, because it may contains documentation on how to use it and thus must be migrated too (this is how PhpSpreadsheet documentation was migrated).

src/PhpSpreadsheet/Helper/Migrator.php
```diff
@@ -248,18 +248,22 @@ public function getMapping()
...
-            '/*.TXT',
         ];
 
         $from = array_keys($this->getMapping());
         $to = array_values($this->getMapping());
 
         foreach ($patterns as $pattern) {
             foreach (glob($path . $pattern) as $file) {
+                if (strpos($path, '/vendor/') !== false) {
+                    continue;
```
PowerKiKi: Adding a "Skipped $path" is a good idea indeed

blunket:
Thanks. Hope this is sufficient now.

PR #595, opened 2018-07-11
Function EXACT support

rolandsusans:
This is: Function EXACT support

```
- [x] a new feature
```

Checklist:

- [x] Code style is respected

Implements excel EXACT function support ( instead of using dummy wrapper use actual implementation)


README.md
```diff
@@ -31,3 +31,46 @@ Because all efforts have shifted to PhpSpreadsheet, PHPExcel will no longer be m
 ## License
 
 PhpSpreadsheet is licensed under [LGPL (GNU LESSER GENERAL PUBLIC LICENSE)](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/LICENSE)
+
+### TODO
```
PowerKiKi: This file should be kept as short as possible. Instead we have a page in documentation that cover supported functions. That should be where we keep a complete list of supported/non-supported functions

PowerKiKi:
Oh, you should also use `composer fix` to fix code style

rolandsusans:
@PowerKiKi is there chance to get this PR in master?

PR #601, opened 2018-07-15
Add new Complex Number Functions introduced in MS Excel 2013

MarkBaker:
- Refactored Complex Engineering Functions to use external complex number library
 - Added calculation engine support for the new complex number functions that were added in MS Excel 2013
    - IMCOSH()  Returns the hyperbolic cosine of a complex number
    - IMCOT()   Returns the cotangent of a complex number
    - IMCSC()   Returns the cosecant of a complex number
    - IMCSCH()  Returns the hyperbolic cosecant of a complex number
    - IMSEC()   Returns the secant of a complex number
    - IMSECH()  Returns the hyperbolic secant of a complex number
    - IMSINH()  Returns the hyperbolic sine of a complex number
    - IMTAN()   Returns the tangent of a complex number

This is:

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [X] Code style is respected

This change provides the new Complex number functions that were introduced in MS Excel 2013 for the calculation engine.
Rather than implementing those functions directly in PHPSpreadsheet, they are implemented through the markbaker/complex repo, and existing complex number functions have been modified to use that library as well to provide the benefits of consistency and to fix a couple of issues in the previous implementation.


PowerKiKi:
Next time, try `composer fix` to fix many CS issues easily. And then `composer check` for a complete check process.

PowerKiKi:
I see `\PhpOffice\PhpSpreadsheet\Calculation::parseComplex()` was removed, this would qualify as a breaking change requiring a major version. What's your opinion about that ? should we keep this PR for a 2.0 ?

MarkBaker:
`parseComplex()` should really have been a private or a protected function in the first place, it was never intended as a part of the public API... possibly it comes down to the number of people who are using complex numbers with PHPSpreadsheet, which I think is a pretty small group... it should probably be saved till 2.0 to avoid breaking BC in a minor release, though I wonder just how many people it would affect

MarkBaker:
For that matter, the number of people likely to make use of the new functions is probably just as small; but they're there for completeness, and for consistency with the MS Excel Engineering function set. I'm also planning on adding the Bitwise functions for the same reason.

And for luck, I'm planning on moving matrix functions out to a separate library as well (and fixing the broken ones too), so that they'll then become a dependency as well... and after that I can do some work on the Trends classes to modernise them as an external library too.

PowerKiKi:
Externalizing libs sounds like a good idea. I've been wondering whether we should split reader/writer as separate packages too in order to modularize required PHP modules (maybe similar to https://github.com/php-cache). But I am not quite ready to commit to that idea yet, as it could increase maintenance work...

PR #603, opened 2018-07-16
Add new Bitwise Functions introduced in MS Excel 2013

MarkBaker:
- Added calculation engine support for the new bitwise functions that were added in MS Excel 2013
   - BITAND()     Returns a Bitwise 'And' of two numbers
   - BITOR()      Returns a Bitwise 'Or' of two number
   - BITXOR()     Returns a Bitwise 'Exclusive Or' of two numbers
   - BITLSHIFT()  Returns a number shifted left by a specified number of bits
   - BITRSHIFT()  Returns a number shifted right by a specified number of bits

This is:

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [X] Code style is respected

This change provides the new Bitwise functions that were introduced in MS Excel 2013 for the calculation engine.


src/PhpSpreadsheet/Calculation/Engineering.php
```diff
@@ -2423,6 +2423,175 @@ public static function erfVal($x)
         return self::$twoSqrtPi * $sum;
     }
 
+    protected static function validateBitwiseArgument($value)
```
PowerKiKi: I think this should be private to limit our API surface and make future maintenance easier. Also maybe a doc block would be nice to document types.

Also could/should the Functions::flattenSingleValue() call be part of this method ?

MarkBaker: Valid comments; and I've also tightened up the validation to handle the different responses for different errors. See my next commit.
I'm very much tempted to write a dedicated validator class, that can be called passing in the argument with a set of validation rules/responses; but that's a more significant piece of work that really needs its own PR rather than bundling with this one.

PR #607, opened 2018-07-17
Fix last column with relative/absolute flags

christian-forgacs:
This is:

- [x] a bugfix

Checklist:

- [x] Code style is respected
- [-] Documentation is updated as necessary

If you read a xls file and in this file you use an absolut reference in SUM function e.g. =SUM(SHEET!A1:SHEET!B1) the script only returns =SUM(SHEET!A1:SHEET!A1)


christian-forgacs:
@PowerKiKi the unit tests were not broken by my changes. Currently your develop branch is broken (2 errors from unit tests). I add some unit tests for this bugfix and push it this weekend.

christian-forgacs:
@PowerKiKi can you explain in witch case the readBIFF8CellRangeAddressB function is used? In my regular project xls file this function is called by an cell with sum function with a range reference.

I reconstruct the case from this project xls file but always readBIFF8CellRangeAddress is called.

What is the different between tArea and tAreaN? With this information I can create a unit test with an xls file because I can publish the xls from my project.

christian-forgacs:
@PowerKiKi I add the unit test for this bugfix and fix the other unit tests with a update for dompdf (from 0.8.0 to 0.8.2).

tests/PhpSpreadsheetTests/Reader/XlsTest.php
```diff
@@ -0,0 +1,82 @@
...
+
+            foreach ($testsForSheet as $cellCoordinate => $result) {
+                $calculatedValue = $sheet->getCell($cellCoordinate)->getCalculatedValue();
+
+                if ($calculatedValue != $result) {
+                    if (empty($calculatedValue) === true) {
+                        $calculatedValue = 'empty';
+                    }
+
+                    throw new \Exception($cellCoordinate . ' != ' . $result . ' - Result in sheet with name ' . $sheetName . ' is ' . $calculatedValue . '.');
```
PowerKiKi: This should be a proper assertion such as `self::assertSame`

tests/PhpSpreadsheetTests/Reader/XlsTest.php
```diff
@@ -0,0 +1,82 @@
...
+use PHPUnit\Framework\TestCase;
+
+class XlsTest extends TestCase
+{
+    /**
+     * Test load Xls file without cell reference.
+     */
+    public function testLoadXlsWithoutCellReference()
+    {
+        $filename = './data/Reader/Xls/without_cell_reference.xls';
```
PowerKiKi: We strive to limit the number of binary files in the repository to keep its size down. Can you replace those with by writing them with PhpSpreadsheet and then reloading them instead ? (this is why I suggested functional tests, see `\PhpOffice\PhpSpreadsheetTests\Functional\AbstractFunctional`)

oleibman:
Superseded by PR #4140.

Issue #437, opened 2018-03-27
Can't escape from Worksheet's cloning loop and exhausts memory


t-iri:
This is:

- [x] a bug report
- [ ] a feature request

/PhpSpreadsheet/Worksheet/Worksheet.php
Method: Worksheet->__clone()
Related: #92

### What is the expected behavior?
I hope to succeed in copying Worksheet containing relatively many cells but it fails in whiteout as below.
```php
<?php

//Load
require __DIR__ . '/vendor/autoload.php';

$reader = \PhpOffice\PhpSpreadsheet\IOFactory::createReader('Xlsx'); 
$workbook = $reader->load('sample.xlsx');

//Prepare copying
$worksheet = $workbook->getActiveSheet(); //contains cells about 2,000 in this sheet
$worksheet->setTitle('Let me copy', false);

//Time to be copied but
$copysheet = $worksheet->copy(); //Infinite loop occurs...

$copysheet->setCellValue('B2', 'Copied?'); //Not reach here
```

### What is the current behavior?
In Worksheet->__clone(), I can see the problem at 'drawingCollection' logic.
```php
                } elseif ($key == 'drawingCollection') {
                    $newCollection = new ArrayObject();
                    foreach ($this->drawingCollection as $id => $item) {
                        if (is_object($item)) {
This Line! =>               $newCollection[$id] = clone $this->drawingCollection[$id];
                        }
                    }
                    $this->drawingCollection = $newCollection;
```
When $item is MemoryDrawing, it also has __clone() and try to clone own value. But in case that value is the Worksheet, it causes nested loop(= back to Worksheet->__clone(), to and fro). I haven't copied the sheet never once...

### What are the steps to reproduce?
f99eb8db 's Message:
>We need to make note of two things:
>
>- The old code was trying to clone an objectArray, and hence was performing a shallow copy of memoryDrawing objects (the __clone magic function was not getting invoked).
>Instead, if one loops over the objectArray, and then clones the individual memory drawing objects, then the __clone function for the MemoryDrawing object is invoked.
>
>- The __clone function for memory drawing was using the clone keyword which does not deal with circular references (Since memoryDrawing object had references to worksheet object, it was encountering an infinite loop). However, serializing and unserializing objects deals with circular references pretty well.

The latter is that but I don't grasp whether it has been solved.
I tried old code(previous of the commit), only declaring clone keyword.
```php
                } elseif ($key == 'drawingCollection') {
[...]

PR #613, opened 2018-07-24
Skipping to clone Worksheet object when cloning DrawingCollection

wnasich:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Fixing issue #437


tests/PhpSpreadsheetTests/Worksheet/DrawingTest.php
```diff
@@ -0,0 +1,22 @@
...
+namespace PhpOffice\PhpSpreadsheetTests\Worksheet;
+
+use PhpOffice\PhpSpreadsheet\Reader\Xls;
+use PHPUnit\Framework\TestCase;
+
+class DrawingTest extends TestCase
+{
+    public function testCloningWorksheetWithImages()
+    {
+        $filename = './data/Reader/Xls/with_images.xls';
```
PowerKiKi: Please refrain from adding binary files to the repo and instead create them from scratch, in memory with PhpSpreadsheet.

tests/PhpSpreadsheetTests/Worksheet/DrawingTest.php
```diff
@@ -0,0 +1,22 @@
...
+        $reader = new Xls();
+        $spreadsheet = $reader->load($filename);
+        $worksheet = $spreadsheet->getSheetByName('Sheet1');
+        $originDrawingCount = count($worksheet->getDrawingCollection());
+
+        $clonedWorksheet = clone $worksheet;
+        $clonedDrawingCount = count($clonedWorksheet->getDrawingCollection());
+        self::assertEquals($originDrawingCount, $clonedDrawingCount);
+    }
```
PowerKiKi: You should assert that the drawing collection is not the exact same object and was indeed cloned. Probably with `assertNotSame`.

 And same for worksheet, it must be cloned. So the patch might be too simple as it is now.

wnasich:
@PowerKiKi thanks for your feedback, I completed your suggestions I think

PR #631, opened 2018-08-09
Fix warning when reading xlsx without styles

AegirLeet:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

`Checklist:`

- [x] Code style is respected

This is basically the same problem as https://github.com/PHPOffice/PHPExcel/issues/1134.
I ran into an .xlsx file which, apparently, doesn't contain styles - this results in a `PHP Warning:  Invalid argument supplied for foreach() in .../vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xlsx.php on line 646` when trying to read it with PhpSpreadsheet.


PowerKiKi:
Thanks !

PR #644, opened 2018-08-23
Remove locale from format string to prevent formatting error

deekthesqueak:
This is:

```
- [x] a bugfix
```

Checklist:

- [x] Code style is respected

When a formatting string has a locale in it an error can occur when outputting. For example when the format string with a locale such as `[$-1010409]#,##0.00;-#,##0.00` appears, a value of 9.98 comes back as $9.98. This is because at https://github.com/PHPOffice/PhpSpreadsheet/blob/1.4.0/src/PhpSpreadsheet/Style/NumberFormat.php#L711 the numberFormat regex will match to the zeros inside the locale ([$-1010409]). Attempts to adjust the numberFormat regex caused regressions in other tests. Adding another step to filter out the locale caused no regression.


PowerKiKi:
Thanks !

PR #649, opened 2018-08-28
fix font color issue in css {inlineStyle}

vipmaa1:
This is:

```
- [X ] a bugfix

to fix font color in css not applied because missing one practice "**{**"


oleibman:
Fixed by PR #831 in Jan. 2019.

PR #660, opened 2018-09-05
fixes for continue inside switch statement

brainfoolong:
This a bugfix for php 7.3 (or even 7.2?) related errors where using continue inside a switch raises a php warning. Either use `continue 2` or `break`. Previously `continue` behaves like `break` but the intended usage is to continue the for loop instead, i guess.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xls.php
+++ b/src/PhpSpreadsheet/Reader/Xls.php
@@ -1127,7 +1127,7 @@ class Xls extends BaseReader
                             // TODO: Why is there no BSE Index? Is this a new Office Version? Password protected field?
                             // More likely : a uncompatible picture
                             if (!$BSEindex) {
-                                continue;
+                                continue 2;
                             }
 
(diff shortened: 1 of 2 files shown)


PowerKiKi:
I cannot find anything related to that in http://php.net/manual/en/control-structures.switch.php. Do you another reference ?

brainfoolong:
Yep, 7.3 change -> see https://github.com/php/php-src/blob/php-7.3.0RC2/UPGRADING  Point 1 ->  Backward Incompatible Changes -> 

    "continue" statements targeting "switch" control flow structures will now
    generate a warning. In PHP such "continue" statements are equivalent to
    "break", while they behave as "continue 2" in other languages.

PowerKiKi:
This demonstrate the error: https://3v4l.org/la8dg

Thanks for the PR !

funtaps:
Are you sure that it is a proper fix?
When I look at _readPpsWks
It seems like, if first $type go to default - there will be error (since $pps wouldn't be defined).
If it is not first iteration, that goes to default - previous $pps will be modified.

Issue #325, opened 2018-01-09
"!" characters in Worksheet title will make reference formulas crash


hampusiggstrom:
This is:

```
- [X] a bug report
- [ ] a feature request
- [X] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
The calculation should be valid. The "!" needs to be added to `$invalidCharacters` or calculations needs to be updated to support exclamation mark characters in sheet titles. 

### What is the current behavior?
```Uncaught PhpOffice\PhpSpreadsheet\Calculation\Exception: Title!!C1 -> Cell coordinate can not be zero-length string```

### What are the steps to reproduce?

```php
<?php

require 'vendor/autoload.php';

use PhpOffice\PhpSpreadsheet\Spreadsheet;
use PhpOffice\PhpSpreadsheet\Writer\Xlsx;

$spreadsheet = new Spreadsheet();
$sheet = $spreadsheet->getActiveSheet();
$sheet->setTitle('Title!');
$sheet->fromArray([1, 2,'=SUM(A1:A2)']);

$writer = new Xlsx($spreadsheet);
$writer->save('hello world.xlsx');
```

### Which versions of PhpSpreadsheet and PHP are affected?
all

hampusiggstrom:
https://github.com/PHPOffice/PhpSpreadsheet/blob/4635d39b4ab341904562e0079d054e117fc56341/src/PhpSpreadsheet/Calculation/Calculation.php#L3463

PowerKiKi:
Can confirm it will crash even wihtout saving like so:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

use PhpOffice\PhpSpreadsheet\Spreadsheet;

$spreadsheet = new Spreadsheet();
$sheet = $spreadsheet->getActiveSheet();
$sheet->setTitle('Title!a');
$sheet->fromArray([1, 2,'=SUM(A1:A2)']);
$sheet->getCell('C1')->getCalculatedValue();
```

hampusiggstrom:
Bug still present in latest release

JonHaymanFCX:
I'm having this issue as well.
When trying to reference another worksheet within a formula, I get an exception thrown.

`$spreadsheet->getActiveSheet()
[...]

PR #662, opened 2018-09-06
Fix for: "!" characters in Worksheet title will make reference formulas crash #325

gurrumpalad:
Bugfix: replace explode by "!" with special function, that extracts sheet name from reference

This is:

- [x] a bugfix

Checklist:

- [x] Code style is respected

When sheet title extracting from string reference (like "Work!sheet1!A1"), php function "explode" divide this string to three parts (['Work', 'sheet1', 'A1']). And then these wrong values using in formulas, ranges, etc.
This change fix that problem by using special function Worksheet::extractSheetTitle (this function also has been changed to make sure that worksheet title can contain "!" character). So, that function search last position of "!" in reference string and divide it to 2 parts correctly (['Work!sheet1', 'A1']).


PowerKiKi:
Thanks !

PR #665, opened 2018-09-07
fix .gitignore

GreatHumorist:
This is:

- [x] an update
- [ ] a bugfix
- [ ] a new feature

Checklist:

- [ ] Code style is respected

The old .gitignore file is not enough to keep up with times, like phpunit.xml now is phpunit.xml.dist;
And cannot use absolute path, but should use relative path instead.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,11 +1,11 @@
-/tests/codeCoverage
-/analysis
-/vendor/
-/phpunit.xml
-/.php_cs.cache
+tests/codeCoverage
+analysis
+vendor/
+phpunit.xml.dist
+.php_cs.cache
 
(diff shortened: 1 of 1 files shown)


PowerKiKi:
Leading slash is intended and correct, as [described in the doc](https://git-scm.com/docs/gitignore):

> A leading slash matches the beginning of the pathname. For example, "/*.c" matches "cat-file.c" but not "mozilla-sha1/sha1.c".

`phpunit.xml.dist` must not be ignored, it is distributed with this project, but `phpunit.xml` might be created locally and must be ignored.

So I see nothing that should be kept in this PR. I'll close it for now. But let me know if you find something else...

PR #694, opened 2018-09-29
Handling Error by unlink() as Exception using try catch

maksible:
**1) Handling "Resource Unavailable" error by unlink() as Exception using try catch so that It doesn't show error _when file is being used by some applications_ when we tries to delete it using unlink.**

Major Changes:

> if (file_exists($pFilename)) {
				//	Added by muhammad.begawala@gmail.com
				//	'@' will stop displaying "Resource Unavailable" error because of file is open some where.
				//	'unlink($pFilename) !== true' will check if file is deleted successfully.
				//  Throwing exception so that we can handle error easily instead of displaying to users.
                if( @unlink($pFilename) !== true )
					throw new WriterException('Could not delete file: ' . $pFilename . ' Please close all applications that are using it.');
            }

**2) $writer->save('hello world.xlsx') will return TRUE if file is successfully created else throw Exception.**

Usage:

> try {
	if( $writer->save('hello world.xlsx') === true )
		echo 'File Created';
}
catch (WriterException $e) {
    echo "Unable to save file. Please close any other applications(s) that are using it: [",  $e->getMessage(), "]\n";
}

This is:
- [x] a bugfix
- [x] a new feature

Checklist:
- [x] Code style is respected

1) Handling "Resource Unavailable" error by unlink() as Exception using try catch so that It doesn't show error when file is being used by some applications when we tries to delete it using unlink.
2) $writer->save('hello world.xlsx') will return TRUE if file is successfully created else throw Exception.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Xlsx.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx.php
@@ -210,7 +210,13 @@ class Xlsx extends BaseWriter
             $zip = new ZipArchive();
 
             if (file_exists($pFilename)) {
-                unlink($pFilename);
+                //Added by muhammad.begawala@gmail.com
+                //'@' will stop displaying "Resource Unavailable" error because of file is open some where.
+                //'unlink($pFilename) !== true' will check if file is deleted successfully.
+                //Throwing exception so that we can handle error easily instead of displaying to users.
+                if (@unlink($pFilename) !== true) {
+                    throw new WriterException('Could not delete file: ' . $pFilename . ' Please close all applications that are using it.');
+                }
             }
             // Try opening the ZIP file
             if ($zip->open($pFilename, ZipArchive::OVERWRITE) !== true) {
@@ -395,7 +401,7 @@ class Xlsx extends BaseWriter
             Calculation::getInstance($this->spreadSheet)->getDebugLog()->setWriteDebugLog($saveDebugLog);
 
             // Close file
-            if ($zip->close() === false) {
+            if (@$zip->close() === false) { // updated by muhammad.begawala@gmail.com
                 throw new WriterException("Could not close zip file $pFilename.");
             }
 
@@ -406,6 +412,8 @@ class Xlsx extends BaseWriter
                 }
                 @unlink($pFilename);
             }
+            return true;
+            // Return True to make sure that file is created successfully with no ExceptionsAdded by muhammad.begawala@gmail.com
         } else {
             throw new WriterException('PhpSpreadsheet object unassigned.');
         }
--- a/src/PhpSpreadsheet/Writer/Xlsx.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx.php
@@ -210,7 +210,6 @@ class Xlsx extends BaseWriter
             $zip = new ZipArchive();
 
             if (file_exists($pFilename)) {
-                //Added by muhammad.begawala@gmail.com
                 //'@' will stop displaying "Resource Unavailable" error because of file is open some where.
                 //'unlink($pFilename) !== true' will check if file is deleted successfully.
                 //Throwing exception so that we can handle error easily instead of displaying to users.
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Writer/Xlsx.php
```diff
@@ -210,7 +210,13 @@ public function save($pFilename)
             $zip = new ZipArchive();
 
             if (file_exists($pFilename)) {
-                unlink($pFilename);
+                //Added by muhammad.begawala@gmail.com
```
PowerKiKi: Don't include author mention in code, you will be credited in the commit

src/PhpSpreadsheet/Writer/Xlsx.php
```diff
@@ -395,7 +401,7 @@ public function save($pFilename)
             Calculation::getInstance($this->spreadSheet)->getDebugLog()->setWriteDebugLog($saveDebugLog);
 
             // Close file
-            if ($zip->close() === false) {
+            if (@$zip->close() === false) { // updated by muhammad.begawala@gmail.com
```
PowerKiKi: Idem, don't include author mention in code, you will be credited in the commit

src/PhpSpreadsheet/Writer/Xlsx.php
```diff
@@ -406,6 +412,8 @@ public function save($pFilename)
                 }
                 @unlink($pFilename);
             }
+            return true;
```
PowerKiKi: IWriter does not specify any return values. We should not return anything.

src/PhpSpreadsheet/Writer/Xlsx.php
```diff
@@ -210,7 +210,13 @@ public function save($pFilename)
             $zip = new ZipArchive();
 
             if (file_exists($pFilename)) {
-                unlink($pFilename);
+                //Added by muhammad.begawala@gmail.com
+                //'@' will stop displaying "Resource Unavailable" error because of file is open some where.
+                //'unlink($pFilename) !== true' will check if file is deleted successfully.
+                //Throwing exception so that we can handle error easily instead of displaying to users.
+                if (@unlink($pFilename) !== true) {
```
PowerKiKi: This would be a better fit for `\PhpOffice\PhpSpreadsheet\Shared\File` so that it can used somewhere else too

src/PhpSpreadsheet/Writer/Xlsx.php
maksible: So, How can we check if file is created successfully or not by $writer->save()? @PowerKiKi 
For Eg. If I want to continue execution of my script *only if file is created*.

Lets say:
```try {
if ($writer->save('hello world.xlsx') === true)
    echo 'File Created'; // Insert into database
}
catch (WriterException $e) {
    echo $e->getMessage();
    // die('File could not be created');
}
```

PowerKiKi: This would be enough:

```php
try {
    $writer->save('hello world.xlsx');
    echo 'File Created'; // Insert into database
}
catch (WriterException $e) {
    echo $e->getMessage();
    // die('File could not be created');
}
```

src/PhpSpreadsheet/Writer/Xlsx.php
maksible: You are right, thats good Idea, But somehow I can's pass travis check, I fixed coding style issues like spaces But still shows Failed with no error, Can you please advise me how to pass that so that I can pull changes.
Thanks @PowerKiKi

PowerKiKi: See https://travis-ci.org/PHPOffice/PhpSpreadsheet/jobs/435169066#L540

maksible: I have checked there already, No errors, Anyways but I have added new pull request as you mentioned to create function in `/Shared/File.php` and updated `/Writer/Xlsx.php` as required.
Please check and let me know, Hope this time it will pass travis test. @PowerKiKi

PowerKiKi:
Made obsolete by #695

PR #695, opened 2018-09-30
Deleteing file using File::deleteFile() created in ..\Shared\File.php

maksible:
**Deleteing file using File::deleteFile() created in** `..\Shared\File.php`
As mentioned by @PowerKiKi I created function into ..\Shared\File so that It can be used any where we need.

**Deletes given file or returns False ITS BEING USED BY SOME APPLICATIONS in** `..\Writer\Xlsx.php`
As mentioned by @PowerKiKi I created function into here so that It can be used any where we need, It returns False If file is exists but is being used by some application instead of giving Error.

This is:

```
- [+] a bugfix
- [+] a new feature
```

Checklist:

- [x] Code style is respected

Handling "Resource Unavailable" error by unlink() as Exception using try catch so that It doesn't show Error when **file is being used by some applications** when we tries to delete it using `unlink()`.

### DIFF
--- a/src/PhpSpreadsheet/Shared/File.php
+++ b/src/PhpSpreadsheet/Shared/File.php
@@ -159,6 +159,7 @@ class File
                 return false;
             }
         }
+
         return true;
     }
 }
--- a/src/PhpSpreadsheet/Shared/File.php
+++ b/src/PhpSpreadsheet/Shared/File.php
@@ -141,22 +141,23 @@ class File
             throw new InvalidArgumentException('Could not open "' . $filename . '" for reading.');
         }
     }
-    
+
     /**
-     * Deletes given file if its exists and is NOT BEING USED BY SOME APPLICATIONS.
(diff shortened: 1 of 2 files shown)


src/PhpSpreadsheet/Shared/File.php
```diff
@@ -141,4 +141,25 @@ public static function assertFile($filename)
...
     }
+
+    /**
+     * Deletes given file if an existing its exists and is NOT BEING USED BY SOME APPLICATIONS.
+     *
+     * @param string $file
+     *
+     * @returns false If file could not be deleted.
+     */
+    public static function deleteFile($file)
```
PowerKiKi: This should be called just "delete" and never return anything, but instead throw exception in case of failure.

Also the message while "not being used by some application" is correct, we should first mention file permissions, which would be a more common case of errors (only Windows prevent a file to be deleted if it's in use, and PhpSpreadsheet would typically not be used on a Windows machine).

Issue #683, opened 2018-09-20
sumif() not calculating when comparing numerical values


DuckDensity:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Calculation of the resultant value in the saved spreadsheet

### What is the current behavior?
The value is calculated correctly if the criteria for `sumif()` is a string.
The value is zero if the criteria is numeric.

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

// Some sample data
$spreadsheet->getActiveSheet()->setCellValue('A1', "Text-Test")
                              ->setCellValue('A2', "Text-Test")
                              ->setCellValue('A3', "Text-Testing")
                              ->setCellValue('A4', "Text-Testing")
                              ->setCellValue('B1', "5")
                              ->setCellValue('B2', "8")
                              ->setCellValue('B3', "12")
                              ->setCellValue('B4', "15")
                              ->setCellValue('C1', 0)
                              ->setCellValue('C2', 0.2)
                              ->setCellValue('C3', 0)
                              ->setCellValue('C4', 0.2);

// This sumif() works when comparing text, and gives result of 13 (5+8) (Cells B1 + B2)
$spreadsheet->getActiveSheet()->setCellValue('A5', "=SUMIF(A1:A4,\"Text-Test\",B1:B4)");
$spreadsheet->getActiveSheet()->setCellValue('A6', "A5 Should be 13");

// This sumif() does not calculate and zero is saved in the spread sheet
// Selecting the cell for editing in Excel and then accepting forces Excel
// to calculate the cell value of 17 (5+12) (Cells B1 + B3)
$spreadsheet->getActiveSheet()->setCellValue('C5', "=SUMIF(C1:C4,0,B1:B4)");
$spreadsheet->getActiveSheet()->setCellValue('C6', "C5 Should be 17");

// Write spreadsheet
[...]

PR #701, opened 2018-10-02
Fix sumif #683

ilicmsreten:
This is:

```
- [y] a bugfix
- [] a new feature
```

Checklist:

- [y] Changes are covered by unit tests
- [y] Code style is respected
- [y] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)

### DIFF
--- a/src/PhpSpreadsheet/Calculation/Functions.php
+++ b/src/PhpSpreadsheet/Calculation/Functions.php
@@ -267,7 +267,7 @@ class Functions
     public static function ifCondition($condition)
     {
         $condition = self::flattenSingleValue($condition);
-        if (!isset($condition[0])) {
+        if (!isset($condition[0]) && !is_numeric($condition)) {
             $condition = '=""';
         }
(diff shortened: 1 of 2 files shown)


DuckDensity:
This is a different issue to #618. This PR affects the calculation of the condition for the sum, count, average etc, whereas #618 relates to the arguments that are to be summed if that condition matches. A similar fix to #618 needs to be put into "averageif".

This PR still needs merging for numeric conditions to be matched. I raised #683, and when I found time to look at the code I came to the same conclusion as this PR.

#618 is required to skip ignore non numeric sum arguments, otherwise a "A non well formed numeric value encountered" notice is generated. Additionally the final calculation will be incorrect where the non numeric value starts with a number, due to the way varibles are evaluated in PHP. Non numeric value "4T" would be ignored in Excel, but evaulated as numeric 4 in PHP and added to the sum.

PowerKiKi:
Indeed, I was wrong, thanks @DuckDensity for letting me know. I rebased and merged it.

PR #703, opened 2018-10-03
Added delete() in Shared/File.php and used in Writer/Xlsx.php

maksible:
This is:

```
- [x] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

1) Handling "Resource Unavailable" error by unlink() as Exception using try catch so that It doesn't show error when file is being used by some applications when we tries to delete it using unlink.
2) $writer->save('hello world.xlsx') will throw Exception if no proper permissions or file us being used by some applications.

### DIFF
--- a/src/PhpSpreadsheet/Shared/File.php
+++ b/src/PhpSpreadsheet/Shared/File.php
@@ -141,4 +141,23 @@ class File
             throw new InvalidArgumentException('Could not open "' . $filename . '" for reading.');
         }
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


maksible:
@PowerKiKi Please review and merge PR, Thanks.

maksible:
So should I create new test file in test folder?

PowerKiKi:
You can add tests in the existing `\PhpOffice\PhpSpreadsheetTests\Shared\FileTest`

PR #706, opened 2018-10-07
Write generated HTML into Mpdf in chunks, rather than as one gigantic string (fixes #637)

### LINKED ISSUE #637: Mpdf export can exceed pcre.backtrack_limit
This is:

```
- [x] a bug report
- [ ] a feature request
```

It should be possible to write large spreadsheets from PhpSpreadsheet to PDF, using the Mpdf writer. In order to achieve this, PhpSpreadsheet should pass HTML into Mpdf in reasonably-sized chunks using multiple calls to `WriteHTML`.

After the Mpdf writer converts a spreadsheet to HTML, it passes the entirety of the spreadsheet as one big HTML string into Mpdf's `WriteHTML` method. The string size can easily exceed the PHP setting `pcre.backtrack_limit` if a larger spreadsheet is being rendered, causing Mpdf to throw an exception and refuse to write the PDF.

Please provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

for ($row = 1; $row < 100; ++$row) {
    for ($col = 1; $col < 100; ++$col) {
        $sheet->getCellByColumnAndRow($col, $row)->setValue('foo bar baz');
    }
}

$writer = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($spreadsheet, 'Mpdf');
$writer->save('/dev/null');
```

I'm working with PHP 7.1.20 and PhpSpreadsheet 1.3.1, but I believe this issue still affects newer versions - the Mpdf writer in PhpSpreadsheet 1.4.0 still passes all the HTML as a single string, so the behaviour is unlikely to have changed.

00dani:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Changes are covered by unit tests (This change isn't covered, since the Mpdf writer doesn't have tests at all. ☹️)
- [x] Code style is respected
- [x] Documentation is updated as necessary (no update necessary)

Due to a limitation in Mpdf, the HTML string passed to its WriteHTML method must not exceed a particular length. PhpSpreadsheet produces one HTML string containing all spreadsheet data when writing to HTML, which can easily exceed Mpdf's size limit. Thus, it was impossible to write large spreadsheets to PDF using the Mpdf writer - this change fixes that issue.


00dani:
Hmm. The thing is, this implementation _definitely_ doesn't pass entire HTML elements into Mpdf - the result of `generateSheetData()` is wrapped in a `<table>` and that `<table>` gets chopped up into multiple strings passed to `WriteHTML()`. And as far as I can tell, it works perfectly. So I can only assume Mpdf is fine with being given fragments like that.

It's a hack for sure, and I'd really prefer to pull out individual rows or even cells from PhpSpreadsheet rather than haphazardly cutting up the HTML. But it definitely works at least.

PowerKiKi:
I tested on `samples/templates/largeSpreadsheet.php` and it seemed to be working. So I merged it, thanks !

PR #713, opened 2018-10-09
Update NumberFormat.php

kdaykin:
Adding GBP number format

This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

GBP is missing, would be nice to have in the UK


PowerKiKi:
I am not willing to merge this, because there are hundreds of currencies that should be also be added if start going that route, and it's probably not a good idea to maintain that kind of data. Especially because it's very easy to declare your own format in your code.

PR #717, opened 2018-10-10
Fix CSV delimiter detection on line breaks (fixes #716)

### LINKED ISSUE #716: CSV delimiter counting can't handle line breaks inside an enclosure
This is:

```
- [*] a bug report
```

When counting delimiters of a CSV that has text with line breaks inside an enclosure it should not count the lines inside the enclosure as separate lines when calculating its deviation.

If you have a file like [Delimiter Test.txt](https://github.com/PHPOffice/PhpSpreadsheet/files/2462089/Delimiter.Test.txt) which has line breaks in it. It incorrectly determines the delimiter as `" "` and not `","`.

Parse this CSV
```
Name,Copy,URL
Test,"This is a test
this is a test
with line breaks
that breaks the
delimiters",http://google.com
Test,"This is a test
this is a test
with line breaks
that breaks the
delimiters",http://google.com
Test,"This is a test
this is a test
with line breaks
that breaks the
delimiters",http://google.com
Test,"This is a test
this is a test
with line breaks
that breaks the
delimiters",http://google.com
```
Then it determines the delimiter as a space and not a comma.

 Version 1.4.1

thisispaul:
The CSV Reader can now correctly ignore line breaks inside
enclosures which allows it to determine the delimiter
correctly.

This is:

```
- [X] a bugfix
```

Checklist:

- [X] Code style is respected

The CSV Reader was incorrectly infering the separator due to it treating new lines inside enclosures as brand new lines in the CSV when it should have been treating it as part of the same line on the CSV.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 - Sheet title can contain exclamation mark - [#325](https://github.com/PHPOffice/PhpSpreadsheet/issues/325)
 - Xls file cause the exception during open by Xls reader - [#402](https://github.com/PHPOffice/PhpSpreadsheet/issues/402)
 - Skip non numeric value in SUMIF - [#618](https://github.com/PHPOffice/PhpSpreadsheet/pull/618)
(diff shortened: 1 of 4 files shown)


PowerKiKi:
Thanks !

PR #726, opened 2018-10-19
Add output encoding option to CSV writer

biplobice:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

There is an option to set encoding on Reader. Let's add it to the writer as well.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Csv.php
+++ b/src/PhpSpreadsheet/Writer/Csv.php
@@ -106,7 +106,6 @@ class Csv extends BaseWriter
         return $this->outputEncoding;
     }
 
-
     /**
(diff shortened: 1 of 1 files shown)


PowerKiKi:
UTF-8 is the current existing encoding for CSV writer. It seems counter-productive to allow for a less inter-operable encoding than that. Especially since you can easily convert your file after it was written out. Unless you can come up with a strong use-case, this won't be merged.

kompadre:
Even with most up to date Excel, when opening a csv file it expects it to be encoded in regions encoding. For instance, cp-1252 here in Spain and it is quite a painful process to change this default option or even open a utf encoded csv with microsoft excel. Just saying.

Issue #729, opened 2018-10-21
Html Reader created invalid spreadsheets if <title> tag is too long


phinor:
This is:.

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
When reading an HTML source file to create a spreadsheet, PhpSpreadsheet will truncate a lengthy (longer than 31 characters) <title> when setting the name of the current worksheet.

### What is the current behavior?
PhpSpreadsheet creates a file with a worksheet name that is too long, prompting Excel to perform "data rescue" operation on opening the file. The worksheet name is truncated by Excel.

### What are the steps to reproduce?
```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create HTML File
$html = "<html><head><title>1234567890123456789012345678901234567890</title></head><body>Content</body></html>";
file_put_contents ("temp.html", $html);

// Create HTML Reader
$objHTMLReader = IOFactory::createReader ("Html");
$excel = $objHTMLReader->load ("temp.html");

// Output File
header ('Content-Type: application/vnd.ms-excel');
header ('Content-Disposition: attachment;filename="Test.xlsx"');
$objWriter = IOFactory::createWriter ($excel, "Xlsx");
$objWriter->save ('php://output');
```

When opening the file produced, a warning message is displayed: "We found problems with some content in 'Test.xlsx'. Do you want us to try to recover as much as we can? If you trust the source of this workbook, click 'Yes'."

The Worksheet name is truncated to "123456789012345678901".

### Which versions of PhpSpreadsheet and PHP are affected?
PHP 7.2, PhpSpreadsheet 1.4.0

kifni41:
I check the 'bug' and it's confirmed.
I tried to the sheet title on Libre Office, and it's seem that they support longer characters. I haven't check it on MS Office.

The title length limitation in on this _Worksheet/Worksheet.php_ in _setTitle_ function

I'm sure this is intended by anyone that code this function back then, but i'm not sure if this limitation is no longer needed now. 

Can anyone with deeper understanding about the spreadsheet reference confirm this matter?


```php
    public function setTitle($pValue, $updateFormulaCellReferences = true, $validate = true)
    {
        // Is this a 'rename' or not?
        if ($this->getTitle() == $pValue) {
            return $this;
        }

[...]

PR #730, opened 2018-10-21
Fix html reader issue :  invalid spreadsheets if <title> tag is too long #729

kifni41:
This is:
Fix for this html reader bug : https://github.com/PHPOffice/PhpSpreadsheet/issues/729
Also while fixed I notices, code in Worksheet.php that use fixed number instead of constant that already defined SHEET_TITLE_MAXIMUM_LENGTH
```
- [v] a bugfix
- [ ] a new feature
```

Checklist:

- [v ] Changes are covered by unit tests
- [v] Code style is respected
- [v] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)

Bug fix


kifni41:
I think the travis ci bit strange, even after i fix the code style, it still failed, maybe caused by last github server problem, and the test not getting the latest commit.
i think it's better to submit another PR

PR #735, opened 2018-10-23
Call static method with static keyword instead of self in DefaultValueBinder

midlan:
When I extended `DefaultValueBinder` and used my custom value binder, I overriden static method `dataTypeForValue` to examine all cells as strings. But my method is not called, because in `DefaultValueBinder` it is called with `self` keyword. Since the class is designed to be extended (not final), I think the use of `self` keyword is bad design, and `static` keyword should be used. When used with `static`, my overriden method `dataTypeForValue` is called properly.

Maybe same problem exists in other classes in project, but I only discovered this one.

### DIFF
--- a/src/PhpSpreadsheet/Cell/DefaultValueBinder.php
+++ b/src/PhpSpreadsheet/Cell/DefaultValueBinder.php
@@ -31,7 +31,7 @@ class DefaultValueBinder implements IValueBinder
         }
 
         // Set value explicit
-        $cell->setValueExplicit($value, self::dataTypeForValue($value));
+        $cell->setValueExplicit($value, static::dataTypeForValue($value));
 
         // Done!
         return true;


PowerKiKi:
Not sure about this. You can easily work around it by overriding `bindValue()` and possibly bypass `dataTypeForValue()` entirely.

@MarkBaker any opinion on this ?

MarkBaker:
The easiest if your extending binder overrides is to call `self::dataTypeForValue()` rather than `parent::dataTypeForValue()`; but there's no reason why the base `bindValue()` shouldn't use late static binding... though it does mean that any extending class has to ensure that ll edge cases are covered by its own implementation

midlan:
@PowerKiKi I would like to keep `bindValue` method unchanged (inherited) for possible future updates. I only wanted change the PHP internal types to document types mapping. Which is done in `dataTypeForValue` method, so it is the only method I wanted override. Now I need override both methods, to _workaround_ (as you said) it. I thinked, when I need to _workaround_ something, it means that something is wrong.

@MarkBaker I dont understand yours 

> it does mean that any extending class has to ensure that ll edge cases are covered by its own implementation

Calling method with static doesn't mean the method must be implemented in it's childs. When calling method with static keyword, the parent method is called when local overriden is not implemented. Let's look on my example: https://www.tehplayground.com/kKvtHVz7qOLVk8PB
(the C class does not contain print method)

MarkBaker:
I mean that if you extend `DefaultValueBinder` with your own value binder, and you implement your own version of the `dataTypeForValue()` method, and that is called from the default `bindValue()` method, then you must guarantee that your version of `dataTypeForValue()` will always return a valid DataType... no returning void, null or boolen false in the event of the script dropping through your own logic, no returning an arbitrary value that isn't defined in `PhpOffice\PhpSpreadsheet\Cell\DataType`

PowerKiKi:
Thanks for the contribution, I merged it

PR #740, opened 2018-10-24
Fix 729 html reader create invalid spreasheet if <title> too long

kifni41:
This is:
Fix for this html reader bug : https://github.com/PHPOffice/PhpSpreadsheet/issues/729
Also while fixed I notices, code in Worksheet.php that use fixed number instead of constant that already defined SHEET_TITLE_MAXIMUM_LENGTH

- [ ] a new feature

- [x] a bugfix

Checklist:

- [x] Code style is respected

Bug fix


PowerKiKi:
The current behavior is by design. The files written are valid, but Excel has artificial limitations. See #177 for details. 

So this won't be merged. It is up to the users to handle sheet titles appropriately if Excel is expected to be used.

PR #742, opened 2018-10-26
[XLSX] Fix sheet title extract function

### LINKED ISSUE #739: Spreadsheet getNamedRange returns null (1.5.0)
This is:

```
- [X] a bug report
- [ ] a feature request
```

Retriveing named ranges from xlsx file.

namedRanges array in Spreadsheet object is empty

Works fine in 1.4.1. 
Breaking change on 1.5.0.

```php
<?php

$spreadsheet = IOFactory::load('My File.xlsx');
$range = $spreadsheet->getNamedRange('Range Name');
// 'My File.xlsx' contains a named range 'Range Name'
// $range is null
 
```

PhpSpreadsheet v1.5
Php  5.6.33

guillaume-ro-fr:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

XLSX extractSheetTitle returned the quote character in sheet name

Fix #739

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/)
 and this project adheres to [Semantic Versioning](http://semver.org/).
 
+## [1.5.1] - 
+
+### Fixed
+
+- Fix sheet title extract function in XLSX Worksheet - [#742](https://github.com/PHPOffice/PhpSpreadsheet/pull/742)
+
 ## [1.5.0] - 2018-10-21
 
 ### Added
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -2734,7 +2734,14 @@ class Worksheet implements IComparable
         }
 
         if ($returnRange) {
-            return [substr($pRange, 0, $sep), substr($pRange, $sep + 1)];
+            $sheet = substr($pRange, 0, $sep);
+            if (($pos = strpos($sheet, "'")) !== false) {
+                $sheet = substr_replace($sheet, '', $pos, 1);
+            }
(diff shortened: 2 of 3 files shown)


guillaume-ro-fr:
There is a delta between the worksheet name and the title extractor, and this impacts several functions, including [`Coordinate::absoluteCoordinate`](https://github.com/PHPOffice/PhpSpreadsheet/blob/2dfd06c59825914a1a325f2a2ed13634b9d8c411/src/PhpSpreadsheet/Cell/Coordinate.php#L98).
I think it is therefore necessary to modify the [test condition](https://github.com/PHPOffice/PhpSpreadsheet/blob/90bb4df777fdaab17310333e87e306b9a1fed97d/src/PhpSpreadsheet/Spreadsheet.php#L724) to exclude quotes, or to modify the storage of sheet names.

PowerKiKi:
I suppose storing in memory only non-quoted title would make sense. But I am not quite sure when quotes _must_ be used... when writing to files only ?

PowerKiKi:
@MarkBaker do you think it would be wise to move to have only non-quoted title in memory and quote them only when writing to files ?

MarkBaker:
> @MarkBaker do you think it would be wise to move to have only non-quoted title in memory and quote them only when writing to files ?
I think that this is going to cause problems for any formula that references a spreadsheet title that contains quotes

PowerKiKi:
I'm concerned about that too. But I think it should be rather easy to unit test quoted and non-quoted sheet name in formulas to prove that it would work. If existing tests are fixed, and new tests cover our concerns, then I'd merge it.

PowerKiKi:
This PR has been automatically closed because the [`develop` branch was permanently deleted](https://github.com/PHPOffice/PhpSpreadsheet/commit/0e8fde9be6d3918ffe030bed0ba0297bfe330c4d).

If this is still important for you, please re-submit a PR against `master` and mention this one in the description.

PR #743, opened 2018-10-26
Fix 337 : Reader\Csv::inferSeparator() generates a notice when the file is empty

slaur:
This is:

```
- [ x ] a bugfix
- [ ] a new feature
```

Checklist:

- [    ] CHANGELOG.md contains a short summary of the change
- [    ] Documentation is updated as necessary

When loading an empty file with `Reader\Csv`, a notice is emitted because the `inferSeparator()` function tries to access an undefined array index. This can lead to an exception depending on your configuration.
Note that to detect the error in the first place, I changed the config to make phpunit fail on notices.

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -5,6 +5,7 @@
     syntaxCheck="true"
     disallowTestOutput="true">
     <php>
+        <ini name="error_reporting" value="-1"/>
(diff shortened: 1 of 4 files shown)


PowerKiKi:
Thanks !

PR #746, opened 2018-10-29
fix iconv E_NOTICE error when import chinese xls

klgd:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

I import a Chinese excel(xls), iconv report E_NOTICE error, and the same content excel(xlsx) is normal
when i use Error Control Operators, it's ok, and the imported data is ok too


PowerKiKi:
Ignoring warning is usually not a good idea, instead the reason for the warning should be investigated and fixed at the source. This won't be merged as is.

klgd:
@PowerKiKi  ok， but i don't know how to fix it with other ways

PR #753, opened 2018-10-30
Stop binding numerically formatted strings as TYPE_NUMERIC

### LINKED ISSUE #752: Please consider changing DefaultValueBinder to always bind strings as TYPE_STRING
This is a feature request.

When writing a document, please consider changing the `Cell::setValue()` behavior to _always_ treat strings as strings. Right now, strings that are formatted as a number are coerced and treated like a number. IMO, this causes confusing behavior. If I want to store a number, I would pass a number. But as I pass a string, I expect PhpSpreadsheet to store a string. There are many use cases where strings that look like numbers should be treated as strings: telephone numbers, employee ID numbers, purchase order numbers, etc. I've run into this enough times, that I now use wrappers around PhpSpreadsheet to workaround this. However, I think this is fundamental enough of a concern that PhpSpreadsheet should consider changing the behavior.

I realize this would be a backwards incompatible change. Some people may be relying on this implicit coercion. I'd argue they should be using numerical types for numerical data. If that is too big of a concern, would it be OK to emit a deprecation warning in `DefaultValueBinder::dataTypeForValue()` to declare that this behavior will change in a future version?

When creating and writing a document, `$cell->setValue('1234')` is stored with type `TYPE_STRING`.

`$cell->setValue('1234')` is stored with type `TYPE_NUMERIC`.

```php
<?php

$cell = $worksheet->getCell('A1');
$cell->setValue('1234');
echo $cell->getDataType() . "\n";
```

develop branch

jdufresne:
Now, string values are always treated as TYPE_STRING. To store a numeric value using the type auto-detection logic, one should a numeric type.

Fixes #752

This is a new feature

Checklist:

- [x] Code style is respected

See issue #752


src/PhpSpreadsheet/Cell/DefaultValueBinder.php
```diff
@@ -58,15 +58,6 @@ public static function dataTypeForValue($pValue)
         } elseif (is_bool($pValue)) {
             return DataType::TYPE_BOOL;
         } elseif (is_float($pValue) || is_int($pValue)) {
-            return DataType::TYPE_NUMERIC;
-        } elseif (preg_match('/^[\+\-]?(\d+\\.?\d*|\d*\\.?\d+)([Ee][\-\+]?[0-2]?\d{1,3})?$/', $pValue)) {
```
jdufresne: Maybe instead of removing it, this should be moved to `AdvancedValueBinder`.

MarkBaker:
I really am not comfortable with this change of behaviour.... we created the concept of binders to allow advanced users to control datatyping, while providing the default as something that simplified datatype handling for less advance users.... and I spend more than enough time explaining to people that `'2018-11-12` is not a date, but a string representing a date... if we force all datatypes to be string by default, then it will simply increase that level of difficulty for people who aren't as understanding of datatypes

MarkBaker:
The logic being removed here assesses the value of strings containing numeric values to see if they begin with a leading zero; and if so, treats them as strings.... the "non-greedy" evaluation of numeric values is supposed to handle telephone numbers, social security numbers, etc such as '0123456' as strings rather than setting them as a numeric type.
It was deliberately adopted as a compromise behaviour; to satisfy those who wanted to retain string datatyping for certain types of numeric values, while retaining the simplicity of automatic typing for those who wanted ease of use.

jdufresne:
> if we force all datatypes to be string by default ...

That is not what I'm suggesting. I'm suggesting that _string_ types be treated as strings. Integers, floating point numbers, date objects, etc. should still be stored in the document as the more obvious type.

MarkBaker:
OK, my bad.... but that does not change my opinion on this issue. That vast majority of PHP developers using this library do not fully understand datatyping and the differences between `123` and `'123'`. They take data from an on-screen form or from a csv file, which comes as a string (I've even seen databases that use varchar for everything) and expect it to work as a number. That is the reason for the default. For the more advanced user, who does understand those differences, then it is a single line of code to switch to using a different value binder.
I'm more than happy to include a new binder in the library that gives you string numerics as strings, but not as the default behaviour.

jdufresne:
Thanks for taking the time to respond. I won't have the time to continue working on this so I'll close this PR for now. Cheers.

PR #760, opened 2018-11-04
Use dedicated PHPUnit assertions

carusogabriel:
Checklist:

- [x] Code style is respected

Using dedicated assertions will give us better error messages and simplify some checks.


PowerKiKi:
Thanks !

PR #773, opened 2018-11-14
Cache readFilter access and skip inner loops for empty/default read filter

### LINKED ISSUE #772: Excessive readFilter->readCell() increase runtime notably when parsing XSLX files
This is:

```
- [x] a bug report
```

Parsing with styles enabled or disabled should not have a considerable different parsing time.

A rather large file with ~6000 rows and 10 columns takes about 2 seconds to parse with `$reader->setReadDataOnly(false)` but takes about 30 seconds with `$reader->setReadDataOnly(true)`.

```php
<?php

// Takes 30 seconds
$reader = \PhpOffice\PhpSpreadsheet\IOFactory::createReaderForFile('large_spreadsheet.xlsx');
$document = $reader->load('large_spreadsheet.xlsx');

// Takes 2 seconds
$reader = \PhpOffice\PhpSpreadsheet\IOFactory::createReaderForFile('large_spreadsheet.xlsx');
$reader->setReadDataOnly(true);
$document = $reader->load('large_spreadsheet.xlsx');
```

PHP 5.6 and PHPSpreadsheet 1.5.0

DennisBirkholz:
This is:

```
- [x] a bugfix
```

Checklist:

- [x] Code style is respected

Loading large XSLX files and using styles has a severe performance penalty (2 seconds vs. 30 seconds for ~6k rows). The performance penalty is mitigated when using no custom read filter

Fixed #772

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -2562,13 +2562,15 @@ class Xlsx extends BaseReader
             }
         }
 
+        $readFilter = (\get_class($this->getReadFilter()) !== DefaultReadFilter::class ? $this->getReadFilter() : null);
+
         // set columns/rows attributes
         $columnsAttributesSet = [];
         $rowsAttributesSet = [];
         foreach ($columnsAttributes as $coordColumn => $columnAttributes) {
-            foreach ($rowsAttributes as $coordRow => $rowAttributes) {
-                if ($this->getReadFilter() !== null) {
-                    if (!$this->getReadFilter()->readCell($coordColumn, $coordRow, $docSheet->getTitle())) {
+            if ($readFilter !== null) {
+                foreach ($rowsAttributes as $coordRow => $rowAttributes) {
+                    if (!$readFilter->readCell($coordColumn, $coordRow, $docSheet->getTitle())) {
                         continue 2;
                     }
                 }
@@ -2581,9 +2583,9 @@ class Xlsx extends BaseReader
         }
 
         foreach ($rowsAttributes as $coordRow => $rowAttributes) {
-            foreach ($columnsAttributes as $coordColumn => $columnAttributes) {
-                if ($this->getReadFilter() !== null) {
-                    if (!$this->getReadFilter()->readCell($coordColumn, $coordRow, $docSheet->getTitle())) {
+            if ($readFilter !== null) {
+                foreach ($columnsAttributes as $coordColumn => $columnAttributes) {
+                    if (!$readFilter->readCell($coordColumn, $coordRow, $docSheet->getTitle())) {
                         continue 2;
                     }
                 }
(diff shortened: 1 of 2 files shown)


MarkBaker:
Are there similar performance gains to be made when loading other formats than xlsx?

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -2562,13 +2562,15 @@ private function readColumnsAndRowsAttributes(SimpleXMLElement $xmlSheet, Worksh
...
         // set columns/rows attributes
         $columnsAttributesSet = [];
         $rowsAttributesSet = [];
         foreach ($columnsAttributes as $coordColumn => $columnAttributes) {
-            foreach ($rowsAttributes as $coordRow => $rowAttributes) {
-                if ($this->getReadFilter() !== null) {
-                    if (!$this->getReadFilter()->readCell($coordColumn, $coordRow, $docSheet->getTitle())) {
+            if ($readFilter !== null) {
+                foreach ($rowsAttributes as $coordRow => $rowAttributes) {
+                    if (!$readFilter->readCell($coordColumn, $coordRow, $docSheet->getTitle())) {
```
MarkBaker: Is there any measurable benefit to caching the title as well?

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -2562,13 +2562,15 @@ private function readColumnsAndRowsAttributes(SimpleXMLElement $xmlSheet, Worksh
             }
         }
 
+        $readFilter = (\get_class($this->getReadFilter()) !== DefaultReadFilter::class ? $this->getReadFilter() : null);
+
```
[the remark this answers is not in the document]
DennisBirkholz: After discussing this with my colleague we came to the conclusion that using `get_class()` is the safe call here. If `DefaultReadFilter` would be final, `instanceof` would suffice here.

src/PhpSpreadsheet/Reader/Xlsx.php
DennisBirkholz: Good question, my xdebug tracing did not show excessive `getTitle()` calls, but maybe I just overlooked them.

DennisBirkholz:
> Are there similar performance gains to be made when loading other formats than xlsx?

- [x] Csv has no styles and `$this->readFilter` is used instead of `$this->getReadFilter()`
- [x] Gnumeric uses the read filter only when reading the cells but `$this->getReadFilter()` is used
- [x] Html does not use the read filter at all
- [x] Ods uses the read filter only when reading the cells but `$this->getReadFilter()` is used
- [x] Slk does not use the read filter at all
- [x] Xls seems to only use read filter when reading a cell but `$this->getReadFilter()` is used
- [x] Xml uses the read filter only when reading the cells but `$this->getReadFilter()` is used

So I would say there is no such large performance hit for other formats.

PR #780, opened 2018-11-20
WIP: Xxe

MarkBaker:
This is:

```
- [ ] a bugfix
- [X] a security fix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

Changes to the xml security scanner to use `libxml_disable_entity_loader()` when cleanly supported and thread-safe, and to handle UTF-7 charset which otherwise permits an XXE exploit


MarkBaker:
Need to enforce  "doctrine/instantiator": "1.0.5" in composer.json (together with PHPUnit) for PHP 5.6; doctrine/instantiator 1.1.0 (if allowed) will break unit tests against PHP 5.6

PR #802, opened 2018-12-04
Fixes #787 libxml_disable_entity_loader() is changed

### LINKED ISSUE #787: PHP speadsheet modifies libxml_disable_entity_loader value
This is:

```
- [X] a bug report
- [ ] a feature request
```

The value of `libxml_disable_entity_loader` stay the same after I use PhpSpreadsheet

Since 1.5.1 and https://github.com/PHPOffice/PhpSpreadsheet/pull/780
The value is changed by : 
libxml_disable_entity_loader(true);

(where test.xlsx is any valid xlsx file)
```php
<?php

$previousValue =  libxml_disable_entity_loader(false);
print_r('previous value: '.$previousValue."\n");

\PhpOffice\PhpSpreadsheet\IOFactory::load('test.xlsx');

$newValue =  libxml_disable_entity_loader(false);
print_r('new value: '.$newValue."\n");
```
With 1.5.1 : 
previous value: 
new value: 1

With 1.5.0
previous value: 
new value: 

PhpSpreadsheet 1.5.1
PHP 7.2

stephanedelprat:
This would be one way to fix https://github.com/PHPOffice/PhpSpreadsheet/issues/787

Somehow when I try to load a PDF the `XmlScanner::getInstance()` is called many times:
A new `getInstance()` is called before the `__destruct()` of the previous one, causing the `previousLibxmlDisableEntityLoaderValue` to contain `false` even when the original value was `true`.

My solution was to share `$previousLibxmlDisableEntityLoaderValue` accross all instances, and be sure to not overwrite the original value.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Security/XmlScanner.php
+++ b/src/PhpSpreadsheet/Reader/Security/XmlScanner.php
@@ -14,11 +14,11 @@ class XmlScanner
     private $libxmlDisableEntityLoader = false;
 
     /**
-     * Store the initial setting of libxmlDisableEntityLoader so that we can resore t later.
+     * Stores the initial setting of libxmlDisableEntityLoader so that we can restore it later.
      *
      * @var bool
      */
-    private $previousLibxmlDisableEntityLoaderValue;
+    static private $previousLibxmlDisableEntityLoaderValue;
 
     /**
      * String used to identify risky xml elements.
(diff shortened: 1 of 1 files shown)


philipp-kolesnikov:
We also faced issues new XmlScanner behavior and I'm afraid that proposed solution will not work in our case.

Code below will print "failed to load external entity "/data/test.xml"" for any valid test.xml file 
```php

libxml_use_internal_errors(true);
libxml_clear_errors();

$document = new DOMDocument();
$document->load("test.xml");

echo libxml_get_last_error()->message. "\n";
```

Because libxml_disable_entity_loader(true) disables loading from file.

Both issues can be fixed by moving all enabling/disabling logic out of constructor/desctructor to `XmlScanner::scan($xml)` function like this
``` php
public function scan($xml)
    {
        if ($this->libxmlDisableEntityLoader) {
            $this->previousLibxmlDisableEntityLoaderValue = libxml_disable_entity_loader(true);
        }
// all old code here 		
        if ($this->libxmlDisableEntityLoader) {
            libxml_disable_entity_loader($this->previousLibxmlDisableEntityLoaderValue);
        }
    return $xml;
}

```

What do you think?

stephanedelprat:
Hello,

Yes, moving the logic to scan() seams looks good to me (works for my case
too), changing global parameters for the shorter time possible is always a
good idea.

Just be sure to never call a scan() within a scan() !


Sincerely,

Stéphane Delprat

philipp-kolesnikov:
Should I submit a new PR, or you will update this one?

> Hello, Yes, moving the logic to scan() seams looks good to me (works for my case too), changing global parameters for the shorter time possible is always a good idea. Just be sure to never call a scan() within a scan()

PowerKiKi:
I think the point of using the constructor/destructor is to be certain that it will be done. So if we move the logic to the scan method, it should use a `try...finally` block.

@philipp-kolesnikov feel free to open a new PR, but be sure to include unit tests, in order to be merged faster.

Also on a sidenote, `XmlScanner::getInstance()` gives the wrong impression that it is a singleton of some sort. But it's not the case at all. I'd be in favor of removing that static method, and making the constructor accept a `IReader` and use a private method to detect the pattern needed. But that's slightly out of topic and could be done in a separate PR.

PR #809, opened 2018-12-10
Exact match in VLOOKUP now returns first match

frantzmiccoli:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

I stumbled upon a case where the value of a VLOOKUP was incorrectly returned compared with Excel.

I have also observed that PHP associative sorting does not preserve key order (for keys which have the same value), which explain the extra test on line 729. See https://gist.github.com/frantzmiccoli/bd03395b8fc831c67828718192c214b8 if you wish to test for yourself with `asort`. `uasort` behaves the same way but the gist was quicker to write with asort.


PowerKiKi:
Thanks ! I merged with a few modifications

PR #814, opened 2018-12-13
Xlsx Reader now ignores blank cells

phinor:
The "readEmptyCells" attribute is ignored when reading spreadsheets, resulting in memory bloat.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected
- [n/a] Documentation is updated as necessary

When reading spreadsheets with large numbers of blank cells, it is very easy for PhpSpreadsheet to run our of memory. The `setReadEmptyCells` should fix this, but the setting is ignored when reading the cells.


src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -972,6 +972,10 @@ public function load($pFilename)
                                                 break;
                                         }
 
+                                        if (!$this->readEmptyCells && ($value === null || $value === '')) {
+                                            continue;
```
PowerKiKi: `continue`ing here will make you skip the `$rowIndex += 1;` which is certainly not what we want. Update tests to cover that non-empty cells coordinates (and values) are correct even when skipping empty cells.

tests/PhpSpreadsheetTests/Reader/XlsxTest.php
```diff
@@ -32,4 +32,18 @@ public function testLoadWithReadFilter()
             $this->assertEquals($ref, \array_slice($data[$i], 0, 10, true));
         }
     }
+
+    /**
+     * Test load Xlsx file while ignoring blank cells.
+     */
+    public function testLoadIgnoringBlankCells()
+    {
+        $filename = './data/Reader/XLSX/99_empty_cells.xlsx';
```
PowerKiKi: Instead of adding a binary file to the repo, create it from scratch, on the fly, when testing. You probably want to use `\PhpOffice\PhpSpreadsheetTests\Functional\AbstractFunctional` for that. And this should also cover XLS to prove that the behavior is similar for both formats. See other functional tests for inspiration.

PR #815, opened 2018-12-13
FormulaR1C1 referencing

jeandelcapitaine:
This is:

```
-
- [x ] a new feature
```

Checklist:

Necessary to correctly insert cells values when looping.

Current absolute or relative references are not adapted.


MarkBaker:
I see a mass of other changes here, but nothing that changes anything to do with R1C1 references in formulae

PowerKiKi:
I'll be closing this, as it should not be a PR to be merged, but rather a feature request.

@jeandelcapitaine please take time to write down a complete example of code. What you provided does not "compile" and we cannot understand what you want to do.

jeandelcapitaine:
Hi,

It is quite simple actually. In Excel, there is this very handy formula to
locate the cell you want:
Example of FormulaRC
<https://stackoverflow.com/questions/12958668/date-change-with-vba-excel-add-subtract-in-two-different-cells/12959457#12959457>

The same does not seem to exist in PHPSpreadhseet. While the row can be
touched easily, let's say to change its dimension
$spreadsheet->getActiveSheet()->getRowDimension('10')->setRowHeight(100);

The same cannot be done for columns where you need to actually use the
letter to set the size
$spreadsheet->getActiveSheet()->getColumnDimension('E')->setCollapsed(true);

My ask is to be able to use also numbers for columns as in FormulaR1C1.
That would enable the use of in variables to loop easly through the
spreasheet and change, get, set whatever you want in the columns.

Today, I do that with the following example. Let's say I want to autosize
the columns:
$alpha = range('A', 'Z');
$col = 1;
$spreadsheet->getActiveSheet()->getColumnDimension($alpha[$col])->setAutoSize(true);

By incrementing $col, I scroll through the sheet. Of course, here I am
limited to the 26 first columns.

So at minimum, be able to set the columns properties addressing them with
numbers, at best using total RC like for example:
$spreadsheet->getActiveSheet()->getDimensionRC('10','5')->setRCSize(100,10);

I hope my request is clearer now

BTW, I have another issue with xlsx saving
<https://stackoverflow.com/questions/53759040/how-to-avoid-excel-bad-format-error-when-saving-spreadsheet/53780933#53780933>.
It appears to be related to
application/vnd.openxmlformats-officedocument.spreadsheetml.sheet and
headers.

I defaulted to xls legacy format and it worked but not optimal.

Best

J.

Le sam. 15 déc. 2018 à 08:35, Adrien Crivelli <notifications@github.com> a
écrit :

MarkBaker:
You do realise that creating your own lookup array of column numbers to letters is redundant; PHPSPreadsheet already provides that for you using the `stringFromColumnIndex` method in the `PhpOffice\PhpSpreadsheet\Cell\Coordinate` class.

```
$spreadsheet->getActiveSheet()
    ->getColumnDimension(Coordinate::stringFromColumnIndex($col))
    ->setAutoSize(true);
```

jeandelcapitaine:
Ah !

Ok, very good then ;) ...

but not very practical to use nevertheless.

Thanks for answering !

Le sam. 15 déc. 2018 à 10:46, Mark Baker <notifications@github.com> a
écrit :

jeandelcapitaine:
Sorry to disturb again.

Could you look at this
<https://stackoverflow.com/questions/53815308/how-to-change-the-sheet-title-color-background>
?

I did not see anything in the docs.

Thanks

J.

jeandelcapitaine:
Hello,

Is there a way to resize the comments boxes with PhpSpreadsheets ?
I see that "Alignment" is not supported but I cannot find any mention of
sizing ?

Thanks in advance for your answer.

J.

>

Issue #648, opened 2018-08-27
Suboptimal memory usage in PhpSpreadsheet\Reader\Xlsx->readColumnsAndRowsAttributes()


tomi-heiskanen:
This is: a bug report & suggested patch

### What is the expected behavior?
Not to consume double amount memory which would be needed for processing.

### What is the current behavior?
The memory usage is unnecessarily doubled as row&column attributes are first read into arrays and then set into the worksheet attributes.

### What are the steps to reproduce?
1. Attempt to read a very large Excel file (hundreds of columns and 1M+ rows)
2. load() runs out of memory (unless gigabytes of memory is allowed for processing)

### Which versions of PhpSpreadsheet and PHP are affected?
At least from 1.4.0 until current develop-branch.

### Suggestion how to fix it
I could not create a pull request due to Github a permission issue, but please see patched version of the function which does not read everything into arrays first, but rather sets the attributes while processing. Patched Xlsx.php: https://gist.github.com/tomi-heiskanen/a1d0e3d376d1b019f6072eda33bd0c11

PowerKiKi:
That would be a very welcome improvement. Could you please give it another try at creating a PR ? As long as you have a GitHub account, you should be able to create PR for any projects. See https://help.github.com/articles/creating-a-pull-request-from-a-fork/

PowerKiKi:
I tried to apply your suggestion on current develop branch. I ended up with the following patch, but I could not measure any memory gain. So unless someone can proof that it is useful, it won't be merged.

```diff
From 962b5fcde082973886f99adb844715245a8d2e78 Mon Sep 17 00:00:00 2001
Date: Mon, 10 Dec 2018 23:22:29 +1100
Subject: [PATCH] wip from
 https://github.com/PHPOffice/PhpSpreadsheet/issues/648

---
 src/PhpSpreadsheet/Reader/Xlsx.php | 78 +++++++++++++-----------------
 t.php                              | 11 +++++
 2 files changed, 44 insertions(+), 45 deletions(-)
 create mode 100644 t.php

diff --git src/PhpSpreadsheet/Reader/Xlsx.php src/PhpSpreadsheet/Reader/Xlsx.php
index 335f5d7e..6abe4adb 100644
--- src/PhpSpreadsheet/Reader/Xlsx.php
+++ src/PhpSpreadsheet/Reader/Xlsx.php
@@ -2520,24 +2520,38 @@ class Xlsx extends BaseReader
      */
[...]

Issue #629, opened 2018-08-07
Reading large File increases Execution time & Memory (E.g file with 500000 records)


kumarkal:
This is: a feature request 

```
- [ ] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
- To Read large files quickly

### What is the current behavior?
- Takes more time to Read

### What are the steps to reproduce?
- Try loading csv file of 500000 records and split it into smaller set of files of about 100000 records each using chunks which will be 5 files.
- As the size of file keeps reducing, performance increases

Please provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

// add code that show the issue here...
```

bferrantino:
I would like to add a bit of clarity to this as well, since we are working together on this project.

We are processing rather large files, that are upwards of 300k lines. In order to scale this and not load the whole file into memory at once, we are using the chunkFilter functionality as laid out here: https://phpspreadsheet.readthedocs.io/en/develop/topics/reading-files/#loading-a-spreadsheet-file under **Reading Only Specific Columns and Rows from a File (Read Filters)** . 

We did some baseline testing of our script and we're noticing that once we pass the threshold of about 250k records, the `$reader->load($inputFileName);` portion of the code takes longer as the file gets larger. No matter how large of a chunk we are using, each time it tries to load the file, it takes about 20-30 seconds. This is compared to a file that is 100k records, that only takes on average 10 seconds to read each chunk. This number seems to increase/decrease significantly as we use a larger/smaller file, respectively.

What we're basically looking to determine here is how we can proceed further. Is there a known limitation with reading a certain sized file? Are there any steps we can take to make the reader perform better? 

Below is the code we're using for our Chunk IReadFilter:
```php
<?php

use PhpOffice\PhpSpreadsheet\Reader\IReadFilter;

/**  Define a Read Filter class implementing IReadFilter  */
class Chunk implements IReadFilter
{
    private $startRow = 0;

    private $endRow = 0;

    /**
[...]

PR #822, opened 2018-12-18
Issue 629 - Use a generator for Cells::getAllCacheKeys to improve performance

yuloh:
Using a generator reduces memory usage and improves performance
when loading large spreadsheets.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

PHPSpreadsheet currently uses a lot of memory when loading large spreadsheets (see #648, #629).  All of the coordinates in the spreadsheet are copied into a new array when the method `Cells::getAllCacheKeys` is called.  Since the coordinates are concatenated with a new string this results in new strings being created in memory.

This result of this method is only ever passed to `Psr\SimpleCache\Cacheinterface::getMultiple` and `Psr\SimpleCache\Cacheinterface::setMultiple`.  Since both of these methods accept an `iterable` we can use a generator instead.  Using a generator means we don't need to build the entire array in memory and instead can return one value at a time as needed.

I benchmarked this with a ~100K row spreadsheet and saw a 16% improvement in run time and 20% improvement in memory usage.  You can view a comparison [here](https://blackfire.io/profiles/compare/914a6eb7-434e-4a98-99fb-077fe9dee6f8...8babce2f-42b1-4d3c-b855-c71aaf07161c/graph?settings%5Bdimension%5D=wt&settings%5Bdisplay%5D=landscape&settings%5BtabPane%5D=nodes&selected=PhpOffice%5CPhpSpreadsheet%5CWorksheet%5CWorksheet%3A%3AgetCell&callname=main()).

You can also view the individual profiles here:

- [baseline profile](https://blackfire.io/profiles/914a6eb7-434e-4a98-99fb-077fe9dee6f8/graph)
- [with generator](https://blackfire.io/profiles/8babce2f-42b1-4d3c-b855-c71aaf07161c/graph)

This is the benchmark script I used:

```php
<?php

$spreadsheet = $reader->load(__DIR__ . '/my_export_100k.xlsx');
```

I also attached the [xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/2691239/my_export_100k.xlsx) if you would like to run the benchmark yourself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 - Fix column names if read filter calls in XLSX reader skip columns - [#777](https://github.com/PHPOffice/PhpSpreadsheet/pull/777)
 - Fix LOOKUP function which was breaking on edge cases - [#796](https://github.com/PHPOffice/PhpSpreadsheet/issues/796)
 - Fix VLOOKUP with exact matches
+- Improved memory usage and performance when loading large spreadsheets - [#822](https://github.com/PHPOffice/PhpSpreadsheet/pull/822)
 
 ## [1.5.2] - 2018-11-25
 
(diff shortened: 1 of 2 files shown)


PowerKiKi:
Excellent PR, thanks !

PR #825, opened 2018-12-18
Issue 629 - Remove duplicate strtoupper

yuloh:
Removing the duplicate strtoupper call has a meaningful impact on
performance since this method is called at least once per cell.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

`Worksheet::getCells` currently calls `strtoupper` twice.  `strtoupper` is kind of expensive and this method is called at least once for every cell in the spreadsheet.  By removing the unnecessary second call the runtime decreases by 18% when importing a ~100K cell spreadsheet.

I _think_ the line could just be changed to `$pCoordinate = strtoupper($pCoordinate)` but I'm not familiar enough with the worksheet references and named ranges to be confident that won't break anything.

- [baseline benchmark](https://blackfire.io/profiles/914a6eb7-434e-4a98-99fb-077fe9dee6f8/graph)
- [this PR benchmark](https://blackfire.io/profiles/bc463091-06fc-43d9-8480-df90753603b7/graph)
- [comparison](https://blackfire.io/profiles/compare/5aa199cc-1bbd-4459-bf6d-402ae00a1d31/graph)

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -1191,8 +1191,11 @@ class Worksheet implements IComparable
      */
     public function getCell($pCoordinate, $createIfNotExists = true)
     {
+        // Uppercase coordinate
+        $pCoordinateUpper = strtoupper($pCoordinate);
+
         // Check cell collection
-        if ($this->cellCollection->has(strtoupper($pCoordinate))) {
+        if ($this->cellCollection->has($pCoordinateUpper)) {
             return $this->cellCollection->get($pCoordinate);
         }
 
@@ -1214,9 +1217,6 @@ class Worksheet implements IComparable
             }
         }
 
-        // Uppercase coordinate
-        $pCoordinate = strtoupper($pCoordinate);
-
         if (Coordinate::coordinateIsRange($pCoordinate)) {
             throw new Exception('Cell coordinate can not be a range of cells.');
         } elseif (strpos($pCoordinate, '$') !== false) {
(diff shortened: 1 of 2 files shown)


src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -1191,8 +1191,11 @@ public function setCellValueExplicitByColumnAndRow($columnIndex, $row, $value, $
      */
     public function getCell($pCoordinate, $createIfNotExists = true)
     {
+        // Uppercase coordinate
+        $pCoordinateUpper = strtoupper($pCoordinate);
+
         // Check cell collection
-        if ($this->cellCollection->has(strtoupper($pCoordinate))) {
+        if ($this->cellCollection->has($pCoordinateUpper)) {
```
yuloh: This seems like a bug to me.  Shouldn't `get` use the same key we checked with `has`?  Based on the code in `Cells` I think this can return null.

src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -1214,17 +1217,14 @@ public function getCell($pCoordinate, $createIfNotExists = true)
             }
         }
 
-        // Uppercase coordinate
-        $pCoordinate = strtoupper($pCoordinate);
-
         if (Coordinate::coordinateIsRange($pCoordinate)) {
```
yuloh: None of these checks actually need to use the uppercase version as they are only checking for symbols, not letters.

PR #826, opened 2018-12-18
Named ranges sheet name (XLSX Reader fix)

frantzmiccoli:
In some files I have observed that names ranges are attached to strangely named sheets. For example if a sheet is named "Hello" the range can be associate with "[1]Hello", since the brackets are forbidden in Excel sheet names I am certain that this has no side effects.

I haven't been able to complete the current unit test as I don't know how to reproduce it in an Excel file, I have just observed it with files I have received.

I have read in some pull request comments, concerns about the speed of regular expressions, since this is happening only on file load, I think it's an ok thing.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

It was causing a bug with some xslx files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 - Fix column names if read filter calls in XLSX reader skip columns - [#777](https://github.com/PHPOffice/PhpSpreadsheet/pull/777)
 - Fix LOOKUP function which was breaking on edge cases - [#796](https://github.com/PHPOffice/PhpSpreadsheet/issues/796)
 - Fix VLOOKUP with exact matches
+- Fix named range resolution
 
 ## [1.5.2] - 2018-11-25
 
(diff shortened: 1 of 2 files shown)


PowerKiKi:
According to http://homepages.ecs.vuw.ac.nz/~elvis/db/Excel.shtml#Workbook the prefix surrounded by `[]` is used to specify the workbook file on disk. But the example you mentioned, and the PR, assume only a single digit. It does not seem like a filename at all and make me think that your files are invalid. 

Also PhpSpreadsheet does not support references to other files (and probably won't support it for a long time, if ever).

So unless you can find official documentation about the syntax you see in your example, this won't be merged.

frantzmiccoli:
Thanks for the feedback, this sheet was not linked to external files. I will try to find reference documentation and post it there if successful.

frantzmiccoli:
I was thinking about this, if I can provide a sample broken file would it be relevant to accept this change?

PowerKiKi:
A sample broken file won't help. What we need the official specification that describe what it s and how it should be handled.

PR #831, opened 2018-12-22
fix color from css in reading from html - add missing '{'

vipmaa1:
This is:

```
- [ x ] a bugfix
- [ ] a new feature
```

Checklist:

in case we generate Spreadsheet from html file and the code in file have text color in css "color:#FF00FF" it will showing as black color because it will render like rgb content with } "FF00FF}"
So, we fix it by adding missing bracket "{"


PowerKiKi:
Thanks !

PR #844, opened 2019-01-07
Feature: branch pruning

frantzmiccoli:
Repost of https://github.com/PHPOffice/PhpSpreadsheet/pull/818 after a rebase on `master`.

This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

Calculation engine was resolving every function by first resolving its arguments including IFs, this was causing significant over evaluation when IFs were used as it meant for every case to be evaluated.

I have tested against 5 files made by 4 different people to ensure that this was not introducing regression, I have observed none. It generates speed improvement from 0% to 80% on those files.

---

**EDIT** Completing from the discussion I had with @PowerKiKi 

Yes, this is touching the core of the calculation engine, so, it would not be to surprising to introduce regression. I did my best to thoroughly test it, if you see extra tests to perform let me know.

As per the extra public methods:

* `Stack::getStackItem()`: enables code factorization as `Calculation::_parseFormula()` doesn't manipulate tokens only through a stack and was also manually creating the arrays representing tokens.
* `Stack::__toString()`: was really convenient to debug I would recommend to leave it but I could truncate it.
* I followed the result caching logic which introduced a few extra methods in `Calculation`: `setBranchPruningEnabled()`, `enableBranchPruning()`, `disableBranchPruning()` and `clearBranchStore()`.
* Nothing to do with this pull request but I am wondering if some functions are not public just to enable testing like `processTokenStack()`, I usually prefer to use the reflection API in my unit tests.

A side point about `CalculationTest::testBranchPruningFormulaParsing`, this test was pretty thick. I could not use a data provider as the expected result testing is way non trivial. So I did split the test in many different functions.

As per the code splitting in the calculation engine itself, the `_parseFormula()` and `processTokenStack()` functions are not using context objects that could be pass around to subrountines. I could spend some time splitting those huge methods as I think I have gained some understanding of the calculation engine inner workings but I think this should be for another pull request. I don't think it would deeply impact execution speed.


frantzmiccoli:
Did a rebase

PowerKiKi:
@MarkBaker this PR still look quite promising should we merge it now ? or would it conflict with your work on the new calculation engine ?

frantzmiccoli:
Side note: we have been using in production since a few weeks now on some pretty advanced sheet without observing any bugs

PowerKiKi:
Thank you for your work and patience. I finally squashed and merged it as 0b387e767e99a9042886ea673d9b9905c7951e87

PR #846, opened 2019-01-08
Implement borders for HTML reader

nscyclone:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

`border` shorthand property from inline styles on `td` and `th` tags can now be used to set `allBorders` on separate cells:
* Border style depends on the `border-width` and `border-style` parts of  `border` using conversion logic similar to HTML Writer.
* Border color depends on the `border-color` of the `border` and allows to use RGB, HEX or named colors. Earlier all the other supported inline styles worked only with HEX-formatted colors, now they support RGB and named colors as well. In order to
achieve this visibility of several functions of the HTML Helper used for color format conversion has been changed from `protected` to `public static`. The only place of their previous usage has been refactored correspondingly. PHPDoc annotations have been added to these functions since they have become exposed to public.

### DIFF
--- a/src/PhpSpreadsheet/Helper/Html.php
+++ b/src/PhpSpreadsheet/Helper/Html.php
@@ -684,7 +684,14 @@ class Html
         $this->stringData = '';
     }
 
-    protected function rgbToColour($rgb)
+    /**
+     * Convert CSS RGB color string to HEX color.
+     *
+     * @param string $rgb a string containing integer RGB values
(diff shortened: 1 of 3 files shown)


patrickbrouwers:
Implemented in #884

nscyclone:
Thanks, @patrickbrouwers!

PR #848, opened 2019-01-14
implement RATE algorithm from ODFF

### LINKED ISSUE #324: RATE function doesn't work
This function doesn't work very well, ie in this case:

RATE(360, -570.3, 93550)

it returns this value: 3.19311841954E-16

and Excel returns a very different value (0,51300% in percentage)

Can you help me please ?

markkimsal:
This is:

```
- [*] a bugfix
```
Fixes #324

Implement the ODFF algorithm for RATE() from LibreOffice.  Add test data that tests floating point $nper.


markkimsal:
The scrutinizer checks is flagging too many things as unused code. 
```
$found = false
while (!$found) {
  if ($check) {
      $found =true;
  }
}
```
It says that $found=true is unused.  That and a few other variables that are definitely not unused.

oleibman:
Superseded by 4107.

PR #854, opened 2019-01-18
Fix VLOOKUP whose return was impacted by order for a $not_exact_match

frantzmiccoli:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

A bug was noticed where the VLOOKUP was impacted by the ordering of inputs. I noticed this after a call to LOOKUP.


frantzmiccoli:
I am still investigating the behavior in Excel and trying to be sure that I am fixing this the right way. I will let you know what I am done.

frantzmiccoli:
I have observed inconsistencies between what is implemented and two line examples (mentioned in the unit tests). I don't understand why Microsoft Excel for Mac 15.35 (170610) behaves how it does. LibreOffice 6.0.3.2 however is consistent with this implementation.

I attached a file 
[lookup-reproduction.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/2773812/lookup-reproduction.xlsx) if you wish to give a look to this.

src/PhpSpreadsheet/Calculation/LookupRef.php
```diff
@@ -719,7 +719,6 @@ public static function VLOOKUP($lookup_value, $lookup_array, $index_number, $not
                 (!is_numeric($lookup_value) && !is_numeric($rowData[$firstColumn]))) {
                 if ($not_exact_match) {
                     $rowNumber = $rowKey;
-                    $rowValue = $rowData[$firstColumn];
 
                     continue;
                 } elseif ((strtolower($rowData[$firstColumn]) == strtolower($lookup_value))
```
MarkBaker: Not related to the bug that you're fixing, but it's probably worth changing these from `strtolower()` to use the `StringHelper::strToLower()` to make them multibyte-safe

frantzmiccoli: Will try to do that over the week-end.

frantzmiccoli:
@MarkBaker done, this fixes the inconsistencies with Excel.

Issue #700, opened 2018-10-02
Wrong calculation of highest column with specified row


michael-roth:
This is:

```
[X] a bug report
```

### What is the expected behavior?
When reading an empty spreadsheet file, the highest (data) column for a given row should be `A` or `null`.

### What is the current behavior?
When calling `\PhpOffice\PhpSpreadsheet\Collection\Cells::getHighestColumn()` with `$row !== null` the method returns the column 1 to high.

### What are the steps to reproduce?
1. Create an empty spreadsheet file (xlsx)
1. Load it with `\PhpOffice\PhpSpreadsheet\Reader\Xlsx::load()`
1. Call `\PhpOffice\PhpSpreadsheet\Collection\Cells::getHighestColumn(1)`
1. The return value will be `'B'`

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$filePath = '/path/to/empty/file';

$reader = IOFactory::createReaderForFile($filePath);
$spreadsheet = $reader->load($filePath);

$highestColumn = $spreadsheet->getActiveSheet()->getHighestColumn(1);
var_dump($highestColumn);
```

### Which versions of PhpSpreadsheet and PHP are affected?
- PhpSpreadsheet: `1.4`
- PHP: `7.2`

michael-roth:
I think the error lies in `\PhpOffice\PhpSpreadsheet\Collection\Cells::getHighestColumn()`:

```php
public function getHighestColumn($row = null)
{
    if ($row == null) {
        $colRow = $this->getHighestRowAndColumn();
        return $colRow['column'];
    }

    $columnList = [1];

    // ...

    return Coordinate::stringFromColumnIndex(max($columnList) + 1);
}
```
`$columnList` always contains `1` and returns its max value +1

poisa:
Same here except sometimes the max value is higher than 1. I sometimes just change an unrelated -lesser- column, hit save in Excel and the `getHighestColumn()` returns a number that suggest 2 or more extra columns.

poisa:
For future reference, the way I went around this problem is by checking the values for each column using some heuristics to detect when there was no more data. For example, if you detect 50 empty columns then most likely your data has already finished. This is definitely not perfect but for my very basic use will suffice.

BenKewell:
The wrong value is caused by two parts:

\PhpOffice\PhpSpreadsheet\Collection\Cells::getHighestColumn()

Calculated value is +1 before return.
[...]

PR #856, opened 2019-01-20
Fix wrong calculation of highest column

BenKewell:
issue #700

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

As described in issue #700, calculation of highest column is wrong.
Returned value is +1 higher than the actual column count.
The wrong value is caused by two parts:

\PhpOffice\PhpSpreadsheet\Collection\Cells::getHighestColumn()
   - Calculated value is +1 before return.

\PhpOffice\PhpSpreadsheet\Reader\Xlsx::readColumnsAndRowsAttributes()
   - A wrong default value is cached when loading file.

This commit fixes both parts and is tested on a production website.

On the production website, the produced XLSX files contain many blank pages when printed, which is due to the wrong +1 column count as described in issue #700. This commit fixes the printing issue with correct count value returned.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -2525,6 +2525,10 @@ class Xlsx extends BaseReader
         if (isset($xmlSheet->cols) && !$this->readDataOnly) {
             foreach ($xmlSheet->cols->col as $col) {
                 for ($i = (int) ($col['min']); $i <= (int) ($col['max']); ++$i) {
+                    if ((int) ($col['max']) == 16384) {
+                        break;
+                    }
+                    
                     if ($col['style'] && !$this->readDataOnly) {
                         $columnsAttributes[Coordinate::stringFromColumnIndex($i)]['xfIndex'] = (int) $col['style'];
                     }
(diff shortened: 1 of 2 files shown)


src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -2525,6 +2525,10 @@ private function readColumnsAndRowsAttributes(SimpleXMLElement $xmlSheet, Worksh
         if (isset($xmlSheet->cols) && !$this->readDataOnly) {
             foreach ($xmlSheet->cols->col as $col) {
                 for ($i = (int) ($col['min']); $i <= (int) ($col['max']); ++$i) {
+                    if ((int) ($col['max']) == 16384) {
```
MarkBaker: Shouldn't this be `> 16384`, otherwise we don't read attributes for the last possible column in the worksheet?

BenKewell: This derives from my debugging attempt to solve the loading issue when I try to load a XLSX file created by MS Office.

The file only uses columns 1 to 3, but the xmlSheet of the file contains two sets of cols:
1. for column 1 to 3
2. for column 4 to 16384

Seems like MS Office creates an extra set to replesent all remaining unused columns.

In the old code, $columnsAttributes is touched for the second set with $col['min'] == 4
and then the loop is terminated since $col['max'] == 16384
Since column 4 is flagged, setColumnAttributes() is later called on column 4 and thus triggers column 4 to be cached as the highest column in \PhpOffice\PhpSpreadsheet\Worksheet

"if ((int) ($col['max']) == 16384)" is the old code's behavior and I have no idea why it's written as that. Looking for a fix to my issue with minimum change to code, I decided to simply terminate the loop early.

BenKewell: Confirmed that the issue in \PhpOffice\PhpSpreadsheet\Reader\Xlsx::readColumnsAndRowsAttributes() is XLSX file dependent and should not be considered as issue #700, changes reverted.

PR #878, opened 2019-02-11
fix error in rendering multiple rowspan

vipmaa1:
This is: fixing bug in rendering html file having multiple rowspan.
Error message 

> A non well formed numeric value encountered

```
- [X] a bugfix
```

Checklist:

- [x] Code style is respected

fixing bug in rendering html file having multiple rowspan, in normal case it's showing error "A non well formed numeric value encountered" because variable `$attributeArray['rowspan']` convert to string and not able to run for loop without recasting it as integer


PowerKiKi:
Thank you for your PR and sorry for the long wait. It has been partially merged in 785705b71258c72f9678f3c2da5fd4f7a1d04c97

PR #908, opened 2019-02-26
Update .gitattributes to remove more non-essential files

JulienChavee:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [ ] Code style is respected

I think that a lot a non-essential files are included in the zip file that Composer create.
I let the folder `samples` in place but IMO this folder should be excluded too, just because this folder shouldn't be there in a production environment.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,16 @@
+# Exclude non-essential files
+
+/.github export-ignore
+/docs export-ignore
 /tests export-ignore
-README.md export-ignore
+/.gitattributes export-ignore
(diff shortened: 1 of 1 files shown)


JulienChavee:
Up to avoid this merge request being closed.

PowerKiKi:
Duplicate of #139. See my answer https://github.com/PHPOffice/PhpSpreadsheet/pull/139#issuecomment-294266691

@MarkBaker would you have a different opinion about excluding docs from releases ?

PR #964, opened 2019-04-16
func getSheetByName update to throw exception instead return null

andrei91gache:
This is:

```
- [] a bugfix
- [*] a new feature
```

Checklist:

- [*] Code style is respected

First the function getsheetByIndex throw a exception when don t find the index
Second not everytime we treat null return, i work with it on prestashop and i use that library for parse xls and csv files automaticaly so when a error come like our supplier write the name of page not good then is more better to have a exception than a null for treat, is more clean.
Thank you to take in consideration.

### DIFF
--- a/src/PhpSpreadsheet/Spreadsheet.php
+++ b/src/PhpSpreadsheet/Spreadsheet.php
@@ -712,10 +712,9 @@ class Spreadsheet
 
     /**
      * Get sheet by name.
-     *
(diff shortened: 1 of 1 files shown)


MarkBaker:
Rather than creating a BC break, which will affect every user of PHPSpreadsheet that already tests for a null return, why don't you just change your code to test for null returns, and have that throw an exception?

PR #1019, opened 2019-06-18
add ability to set codepage explicitly for BIFF5

boltunoreh:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

If excel 95 file doesn't contain codepage record, cyrillic text can't be decoded well, because CP1252 charset is used by default and it can't be changed explicitly. Setting codepage to CP1251 does the job.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xls.php
+++ b/src/PhpSpreadsheet/Reader/Xls.php
@@ -440,6 +440,14 @@ class Xls extends BaseReader
     }
 
     /**
+     * @param string $codepage
+     */
+    public function setCodepage($codepage)
+    {
+        $this->codepage = $codepage;
+    }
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Reader/Xls.php
```diff
@@ -439,6 +439,14 @@ public function canRead($pFilename)
         }
     }
 
+    /**
+     * @param string $codepage
+     */
+    public function setCodepage($codepage)
```
PowerKiKi: The method should have a comment explaining what it does.

oleibman:
Equivalent functionality added by PR #1484 in June 2020.

PR #858, opened 2019-01-23
Dupplicate call of strtoupper and multiple calls auf preg_replace optimizing

fglueck:
strtoupper also in setSelectedCells()

This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Changes are covered by unit tests
- [X] Code style is respected
- [X] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [x] CHANGELOG.md contains a short summary of the change
- [x] Documentation is updated as necessary

### Why this change is needed?
Just for Performance

look at:
    public function setSelectedCells($pCoordinate)
    {
        // Uppercase coordinate
        $pCoordinate = **strtoupper($pCoordinate);**

PR #1040, opened 2019-06-28
Performance optimization (dupplicate call of strtoupper and multiple calls auf preg_replace)

fglueck:
Performance optimizing of dupplicate call strtoupper and preg_replace.

```
- [x ] a bugfix
```

same like:
[https://github.com/PHPOffice/PhpSpreadsheet/pull/825](https://github.com/PHPOffice/PhpSpreadsheet/pull/825)


oleibman:
This appears to be a duplicate of #858. As noted in that ticket, getStyle no longer performs strtoupper, and the Validations class, new since this PR was introduced, combines its preg_replace calls into 1. Removing stale label, but keeping PR closed.

PR #1048, opened 2019-07-02
Fix register mismatch

Thereisnobugs:
This is:

```
- [ x ] a bugfix
```

Some xml files contains lowercase string encoding="utf-8"


MarkBaker:
Implemented, thank you

PR #1050, opened 2019-07-02
Allow XmlScanner to correctly restore libxml entity_loader setting

### LINKED ISSUE #1045: 1.8 fix breaks xmlreader when opening multiple files
This is a bug report.

```
<?php

use PhpOffice\PhpSpreadsheet\Reader\Xlsx;
use PHPUnit\Framework\TestCase;

class BugTest extends TestCase
{
    public function testDev(): void
    {
        $reader = new Xlsx();
        $reader->load('tests/test1.xlsx');
        unset($reader);

        $reader = new \XmlReader();
        $reader->open('tests/test2.xlsx');
    }
}
```

```
PHPUnit 6.5.14 by Sebastian Bergmann and contributors.

E                                                                   1 / 1 (100%)

Time: 462 ms, Memory: 10.00MB

There was 1 error:

1) BugTest::testDev
XMLReader::open(): Unable to open source data

tests\BugTest.php:20
```

`1.8, 7.2`

### Cause

XmlScanner will not leave scope until the script terminates since its reference is stored in the shutdown handler callback. Resetting of the xml entity loader setting has to be done in a function and not in a method. 

https://github.com/PHPOffice/PhpSpreadsheet/commit/0e6238c69e863b58aeece61e48ea032696c6dccd#diff-5c5eea600b92ba6e320a104de4fbb16eR35

rtek:
XmlScanner was not restoring libxml_disable_entity_loader since
destruct was not being called until script shutdown. This is because
the shutdown handler required an XmlScanner instance.

Also fix an unrelated bug where the UTF-8 encoding test was
case sensitive.

This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

Fix #1045

### DIFF
--- a/src/PhpSpreadsheet/Reader/BaseReader.php
+++ b/src/PhpSpreadsheet/Reader/BaseReader.php
@@ -58,21 +58,6 @@ abstract class BaseReader implements IReader
     public function __construct()
     {
         $this->readFilter = new DefaultReadFilter();
-
-        // A fatal error will bypass the destructor, so we register a shutdown here
-        register_shutdown_function([$this, '__destruct']);
-    }
-
-    private function shutdown()
-    {
-        if ($this->securityScanner !== null) {
-            $this->securityScanner = null;
-        }
-    }
-
-    public function __destruct()
-    {
-        $this->shutdown();
     }
 
     public function getReadDataOnly()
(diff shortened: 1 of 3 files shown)


MarkBaker:
Thanks for the submission; I'd been struggling to work out how I could implement this for the thread-safe and non-thread-safe use cases, and for the situation where a fatal error terminates the script.
I wastrying to over-engineer; but it's really quite simple.

PR #1053, opened 2019-07-03
Added condition to nullify cells when merging cells.

ReDev1L:
This is:

- [x] a bugfix

Checklist:

- [x] Code style is respected

When we merge cells, we nullify their values as it does Excel for windows.
But this breaks autofilter work on merged cells.
This PR adds parameter to Worksheet->mergeCells($pRange, $withAutoFilter = false) method.
When $withAutoFilter is true -> method wont nullify cells, so autofilter will work on merged cells.

I've made changes backwards - compatible, but i dunno how removed cell nullifying will affect on other parts of library. We have tested that PR on our excel sheets, and we are satisfied by that fix, using this in production.

Sorry, but i'm not ready to work on this PR after submitting.

No filters, merged cells:
no_filters

Applying filters:
filters

Filters applied - BROKEN, before this PR:
(2 rows, but 10 needed as in next screenshot, 2 rows because of collapse, there will be 1 row if not using collapse)
Annotation 2019-07-03 135803

Filters applied - WORKING, after this PR:
filters_applied

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -1679,11 +1679,12 @@ class Worksheet implements IComparable
      *
      * @param string $pRange Cell range (e.g. A1:E1)
      *
-     * @throws Exception
+     * @param bool   $withAutofilter Fill merged cells for autofilter
      *
      * @return Worksheet
+     * @throws Exception
      */
-    public function mergeCells($pRange)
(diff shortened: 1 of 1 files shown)


ReDev1L:
No filters, merged cells:
no_filters

Applying filters:
filters

Filters applied - BROKEN, before this PR:
broken_filters_applied

Filters applied - WORKING, after this PR:
filters_applied

PR #1054, opened 2019-07-03
Use case-insensitive comparison for encoding

crishoj:
The recent change incorrectly assumes the encoding to be uppercase (`UTF-8`), but [IANA](https://www.iana.org/assignments/character-sets/character-sets.xhtml) says _no distinction is made
between use of upper and lower case letters_.

This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

The recent change to address CVE-2019-12331 breaks parsing of spreadsheets with lowercase XML, such as the following:

```xml
<?xml version="1.0" encoding="utf-8" standalone="yes"?>
...
```

### DIFF
--- a/src/PhpSpreadsheet/Reader/Security/XmlScanner.php
+++ b/src/PhpSpreadsheet/Reader/Security/XmlScanner.php
@@ -96,12 +96,12 @@ class XmlScanner
         $result = preg_match($pattern, $xml, $matches);
         $charset = strtoupper($result ? $matches[1] : 'UTF-8');
 
-        if ($charset !== 'UTF-8') {
+        if (strcasecmp($charset, 'UTF-8')) {
             $xml = mb_convert_encoding($xml, 'UTF-8', $charset);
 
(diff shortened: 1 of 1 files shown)


MarkBaker:
Thank you; this issue should already have been resolved in master

crishoj:
Ah, good.

I find the following form easier to comprehend:

```php
$charset = preg_match($pattern, $xml, $matches) 
    ? strtoupper($matches[1]) 
    : 'UTF-8';
```

src/PhpSpreadsheet/Reader/Security/XmlScanner.php
```diff
@@ -96,7 +96,7 @@ private function toUtf8($xml)
         $result = preg_match($pattern, $xml, $matches);
         $charset = strtoupper($result ? $matches[1] : 'UTF-8');
```
PowerKiKi: `strtoupper` here make the new `strcasecmp` unnecessary

PR #1076, opened 2019-07-14
Fix 1055

Juluan:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

To fix the issue 1055, which make PhpSpreadsheet Hlookup function work only if the list is ordered if range_lookup is false, which isn't the behaviour in Excel and Calc documentation


MarkBaker:
Thank you for the contribution

PR #1109, opened 2019-07-25
Php74 preparatory work

MarkBaker:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

Fixes for stricter offset access type checking in PHP 7.4


kornrunner:
Instead of opening another PR, and since this one already deals with the topic:

On:
https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php#L1108 
`$attributes['t']` might not be set, and in that case it issues a deprecation notice under 7.4 (see https://twitter.com/nikita_ppv/status/1148931277216800769)

Thanks!

PR #1115, opened 2019-07-30
fix getCalculatedValue

st875427:
This is:

```
- [o] a bugfix
```

Checklist:

- [x] Code style is respected
- Documentation is updated as necessary

Because when the formula got two or more INDIRECT in one line, the file will NOT be generated. It is a very dangerous for output a file. So we need to fix it.


st875427:
Hi, I've added unit test to demonstrate the bug. Please review it, thanks.

PowerKiKi:
Thanks ! it's been merged

PR #1133, opened 2019-08-07
Fix issue 1132: import CSV values as strings

ondrej-bouda:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

Fixes issue 1132.


PowerKiKi:
Interestingly both LibreOffice and Excel will guess the type of data when reading CSV. Not only will they interpret `165008E4` as a number, but they will also interpret `=A3*2` as a valid formula. This is the exact behavior of PhpSpreadsheet currently. And that's why I don't think this PR should be merged.

But to solve your use-case, you can easily use a custom `IValueBinder`. [Read more here](https://phpspreadsheet.readthedocs.io/en/latest/topics/accessing-cells/#using-value-binders-to-facilitate-data-entry).

ondrej-bouda:
Cool, thank you for your suggestion on using a custom `IValueBinder`, should do the job for me.

I must agree that PhpSpreadsheet is working as expected since both the spreadsheet programs do the same. It's just a pity they do IMHO, but once they do, I also think PhpSpreadsheet should adhere to it.

PR #1136, opened 2019-08-09
Allow HTML Reader to load from string

gnat42:
We often want to export a table as an excel sheet. The system renders the html and it seems like a waste of time to write it to the file system to use the reader. This allows us to render the html and then just pass it to a reader

This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

Just allows Reader\Html to accept html content as a string instead of just a file.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Html.php
+++ b/src/PhpSpreadsheet/Reader/Html.php
@@ -592,12 +592,6 @@ class Html extends BaseReader
             throw new Exception($pFilename . ' is an Invalid HTML file.');
         }
 
-        // Create new sheet
-        while ($spreadsheet->getSheetCount() <= $this->sheetIndex) {
-            $spreadsheet->createSheet();
-        }
-        $spreadsheet->setActiveSheetIndex($this->sheetIndex);
-
         //    Create a new DOM object
         $dom = new DOMDocument();
         //    Reload the HTML file into the DOM object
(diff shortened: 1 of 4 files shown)


gnat42:
I can add the rest of the checklist items if this is something that you think you would merge

PowerKiKi:
It sounds like it could be useful, but to be merged the checklist will have to be completed and the code should not be duplicated, instead create a common private method used by both `load` and `loadFromString`. Most likely that new method would receive a `DOMDocument`. Also be sure to make use of `$this->securityScanner`

gnat42:
@PowerKiKi  Sounds good, I just didn't want to go to a whole lot of trouble if the basic idea wouldn't be accepted. I've rebased and re-worked the PR to include all checklist items. Let me know if there are any issues that need tweaking.

codysnider:
@gnat42 Let me know if you need help with this. I was planning to put something similar in today but looks like you are 90% there.

gnat42:
What's missing? I added tests and docs and added to the changelog?

src/PhpSpreadsheet/Reader/Html.php
```diff
@@ -606,14 +600,56 @@ public function loadIntoExisting($pFilename, Spreadsheet $spreadsheet)
...
+    {
+        //    Create a new DOM object
+        $dom = new DOMDocument();
+        //    Reload the HTML file into the DOM object
+        $loaded = $dom->loadHTML(mb_convert_encoding($this->securityScanner->scan($content), 'HTML-ENTITIES', 'UTF-8'));
+        if ($loaded === false) {
+            throw new Exception('Failed to load content as a DOM Document');
+        }
+
+        return $this->loadDom($dom, new Spreadsheet());
```
codysnider:
```suggestion
        return $this->loadDocument($dom, new Spreadsheet());
```

codysnider:
As far as I can tell, just that (picked up by scrutinizer).

gnat42:
So I'm a bit confused as to how the code coverage decreased. I have a test that tests loadFromString and it passes, but the code coverage makes it out as if loadFromString isn't tested at all.

codysnider:
You may want to assign @PowerKiKi as both an assignee and a reviewer. I'm not a maintainer and haven't the means to approve or deny this PR.

gnat42:
Hey @codysnider I'm not sure I'm able to assign anyone. Hopefully @PowerKiKi sees and assigns himself.

PowerKiKi:
Thanks ! it has been merged as 95c8bb9918b43a1f088cab54c4684e1ef40caaaa (with minor modifications)

Issue #804, opened 2018-12-05
PhpSpreadsheet content.xml in ODS file must be minified only


akimsullec:
This is:

```
- [*] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Read ODS files without errors
### What is the current behavior?
Fatal error: Uncaught Error: Call to undefined method DOMText::hasAttributeNS() in {{some path prefix}}/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Ods.php:467 Stack trace: #0 

### What are the steps to reproduce?
Some ODS generators as Trassir devices for example, makes ODS format reports with unminified content.xml file in ODS archive.
Just unminify content.xml in any ODS archive and you'll catch the error.

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

// add code that show the issue here...
```

### Which versions of PhpSpreadsheet and PHP are affected?
php 7.0
phpss - latest

cqix:
A test file
[test.ods.zip](https://github.com/PHPOffice/PhpSpreadsheet/files/3493532/test.ods.zip)

oleibman:
Reopening, expect fix soon.

PR #1137, opened 2019-08-12
Fix Error in Ods.php Reader for some files

cqix:
#804 Catch Uncaught Error Call to undefined method DOMText::hasAttributeNS() in Ods.php Reader

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Some ODS files have a childNode of type DOMText, instead of DOMElement only. This tiny fix allows to read these files.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Ods.php
+++ b/src/PhpSpreadsheet/Reader/Ods.php
@@ -377,6 +377,10 @@ class Ods extends BaseReader
                                     }
                                 }
 
+                                if (!method_exists($cellData, 'hasAttributeNS')) {
+                                    continue;
+                                }
+                                
                                 // Initialize variables
                                 $formatting = $hyperlink = null;
                                 $hasCalculatedValue = false;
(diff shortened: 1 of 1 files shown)


cqix:
> Instead of loosely testing the existence of a method, it would be best to accurately test the expected type of child, and also extract DomElement specific code into a private method, to make it easier to understand what's going on.
> 
> Also tests and changelog would be nice too.

Yes, that whole code is a little complicated to read. I did not event find out what the rest of the code was dooing, so I catched for the obvious.
Anyways, I also want to push some other stuff soon (NamedRanges for ODS Writer), but I need to find some time for tests etc before.

rsevero:
Faced the same _Uncaught Error Call to undefined method DOMText::hasAttributeNS() in Ods.php Reader_ during some imports.

Applied the fix provided at https://github.com/PHPOffice/PhpSpreadsheet/pull/1137#issue-306581193 and my imports are working again.

oleibman:
Superseded by PR #4087.

PR #1143, opened 2019-08-19
fixbug:the setCellValue second parameter can't accept int

ruibin:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Spreadsheet.setCellValue method cannot accept int parameter. I got fatal error:"Trying to access array offset on value of type int at /vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/DefaultValueBinder.php:58"


oleibman:
Code in question now tests for float/int first.

PR #1148, opened 2019-08-28
Set Branch pruning Off by default 

rolandsusans:
This is:

- [x] a bugfix

https://github.com/PHPOffice/PhpSpreadsheet/issues/1149

Checklist:

- [ ] Code style is respected


frantzmiccoli:
I think we could do better than disabling by default by fixing the bug you may have discovered.

The branch pruning is not supposed to have side effects, but as you noticed I was not sure of this.

You add a test with a dependency on an Excel file that contains thousands of lines, could you consider updating `CalculationTest.php` with an minimal reproducible example that shows your point ? You have an example of how created a spreadsheet without an object on line 122 https://github.com/PHPOffice/PhpSpreadsheet/blob/master/tests/PhpSpreadsheetTests/Calculation/CalculationTest.php#L122

If that's ok, once you are done I will investigate on that.

rolandsusans:
@frantzmiccoli I have simplified the excel file. ( reduced amount of rows, left only one worksheet )
1 test is meant to be failing  - it highlights the case where branch pruning is braking the calculation.

rolandsusans:
Hope it helps you to find the bug in-branch pruning.

rolandsusans:
@frantzmiccoli have you made any progress on this investigation ?

frantzmiccoli:
@rolandsusans No, I will try to spend time on it this week-end.

frantzmiccoli:
Sorry I have been lacking time to work on that, I put it in my agenda for next sunday.

PowerKiKi:
@frantzmiccoli thank you for investigating this one. I took the liberty to assign you the issue...

frantzmiccoli:
This is not a bug related with the branch pruning but with IF formula resolution.

```php
$workSheet = $spreadsheet->getActiveSheet();

$cellA1 = $workSheet->getCell('A1');
$cellA1->setValue('=IF(NA(),1,0)');
$this->assertEquals('#N/A', $cellA1->getCalculatedValue());
// instead of '#N/A' we get 1 here.
```
In the sample file, instead of 1, the resolved value is '#N/A'. I am working on a fix.

frantzmiccoli:
See https://github.com/PHPOffice/PhpSpreadsheet/pull/1165

This fix is not enough to fix inconsistencies though.

frantzmiccoli:
https://github.com/PHPOffice/PhpSpreadsheet/pull/1167 fixes the remaining problem.

rolandsusans:
@frantzmiccoli Tnx for investigation. I checked those PR's and left some comments. Will close this PR.

PR #1166, opened 2019-09-22
scrutinizer do not analyze test code

diegosouza:
This is:

```
- [ ] a bugfix
- [x] a CI adjustment
- [ ] a new feature
```

Checklist:

- [x] Changes are covered by unit tests (**does not apply**)
- [x] Code style is respected (**does not apply**)
- [x] CHANGELOG.md contains a short summary of the change (**does not apply**)
- [x] Documentation is updated as necessary (**does not apply**)

The scrutinizer score is low, but it is lower because it is inspecting code under `tests` too.

[Here is an example of a relevant project](https://github.com/laravel-zero/laravel-zero/blob/stable/.scrutinizer.yml) doing the way I suggest you.

I think the `tests` directory should be ignored by scrutinizer. Do you agree?

### DIFF
--- a/.scrutinizer.yml
+++ b/.scrutinizer.yml
@@ -9,6 +9,10 @@ coding_style:
             around_operators:
                 concatenation: true
 
+filter:
+    excluded_paths:
+        - "tests/"
+
 build:
     nodes:
         analysis:


PowerKiKi:
From PhpSpreadsheet point of view test code is our code too and its quality should not be neglected. So I think Scrutinizer should analyse it too.

PR #1168, opened 2019-09-23
Fixes pointed by static analysis

diegosouza:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

To have a little less technical debt.

### DIFF
--- a/src/PhpSpreadsheet/Reader/IReadFilter.php
+++ b/src/PhpSpreadsheet/Reader/IReadFilter.php
@@ -9,9 +9,8 @@ interface IReadFilter
      *
      * @param string $column Column address (as a string value like "A", or "IV")
      * @param int $row Row number
-     * @param string $worksheetName Optional worksheet name
      *
      * @return bool
      */
-    public function readCell($column, $row, $worksheetName = '');
+    public function readCell($column, $row);
 }
--- a/src/PhpSpreadsheet/Writer/Ods/Content.php
+++ b/src/PhpSpreadsheet/Writer/Ods/Content.php
@@ -12,7 +12,6 @@ use PhpOffice\PhpSpreadsheet\Style\Font;
 use PhpOffice\PhpSpreadsheet\Worksheet\Row;
 use PhpOffice\PhpSpreadsheet\Worksheet\Worksheet;
 use PhpOffice\PhpSpreadsheet\Writer\Exception;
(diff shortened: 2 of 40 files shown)


src/PhpSpreadsheet/Reader/IReadFilter.php
```diff
@@ -9,9 +9,8 @@ interface IReadFilter
      *
      * @param string $column Column address (as a string value like "A", or "IV")
      * @param int $row Row number
-     * @param string $worksheetName Optional worksheet name
      *
      * @return bool
      */
-    public function readCell($column, $row, $worksheetName = '');
+    public function readCell($column, $row);
```
PowerKiKi: This is a major breaking change in public interfaces and should be reverted, along with all other implementing classes

src/PhpSpreadsheet/Writer/Ods/Content.php
```diff
@@ -12,7 +12,6 @@
 use PhpOffice\PhpSpreadsheet\Worksheet\Row;
 use PhpOffice\PhpSpreadsheet\Worksheet\Worksheet;
 use PhpOffice\PhpSpreadsheet\Writer\Exception;
-use PhpOffice\PhpSpreadsheet\Writer\Ods;
```
PowerKiKi: This is required and must be reverted.

src/PhpSpreadsheet/Writer/Xlsx.php
```diff
@@ -232,6 +232,7 @@ public function save($pFilename)
                         //signed macros ?
                         // Yes : add the certificate file and the related rels file
                         $zip->addFromString('xl/vbaProjectSignature.bin', $this->spreadSheet->getMacrosCertificate());
+                        // @scrutinizer ignore-call
```
PowerKiKi: A better way to fix this is to introduced dedicated properties for each parts, instead of using a loosely type array. So this should be reverted, and introducing properties for each part should be another dedicated PR.

diegosouza:
> Thank you for the PR, some commits are interesting, but a few of them should be improved before merging. Will you work on that ? or should I cherry-pick only the interesting commits ?

It's better you cherry pick if you want. I won't make additional changes. The codebase is too messy/weak.

PowerKiKi:
Cherry picked what was needed, thanks

PR #1177, opened 2019-09-26
Validate XIRR inputs and return correct error values

### LINKED ISSUE #1120: XIRR returning warnings and incorrect result
This is:

```
- [X] a bug report
- [ ] a feature request
```

I would expect the result to be 0.13796 (as calculated using XIRR in Excel)
However, this is because my dates are given in non ascending order and the document is clear that this is wrong. I believe a VALUE error should be returned.

Lots of warnings given
```
Warning: A non-numeric value encountered in /vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Calculation/Financial.php on line 2288
```
The result given is 0.16448

```php
<?php

$values = [
	1893.67,
	139947.43,
	52573.25,
	48849.74,
	26369.16,
	-273029.18
];

$dates = [
	43643,
	43636,
	43637,
	43640,
	43643,
	43673,
];

$result = \PhpOffice\PhpSpreadsheet\Calculation\Financial::XIRR($values,$dates);

echo $result;
```

PHP 7.2
Latest PHPSpreadsheet 1.8.2

blacknell:
Fix: Return #NUM! if values and dates contain a different number of values
Fix: Return #NUM! if there is not at least one positive cash flow and one negative cash flow
Fix: Return #NUM! if any number in dates precedes the starting date fixes #1120 
Fix: Return #NUM! if a result that works cannot be found after max iteration tries
Fix: Correct DocBlocks for XIRR & XNPV
Add: Validate XIRR with unit tests

This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

If the date order is non ascending then numerous warnings are generated and the wrong result returned. In addition, a number of input cases result in the wrong error value being returned.


PowerKiKi:
Thanks !

PR #1181, opened 2019-10-01
Dealing with invalid xml tag ("<br>" without closing) in vml comments

rdoering:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

- Excel writes nie-lines VML-comments as "<br>"-tag without a closing "<\br>"-tag 
- This is no valid xml code
- therefore simplexml_load_string() is failing and phpShreadsheet is failing to open the file


rdoering:
If the missing php 7.4 support is a problem, I could try to fix it.

rdoering:
Any questions?

rdoering:
Is here any progress?

rdoering:
Travis is only failing on PHP 7.4.

rdoering:
Any news here?

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -916,8 +900,10 @@ public function load($pFilename)
                                     $relPath = File::realpath(dirname("$dir/$fileWorksheet") . '/' . $relPath);
 
                                     try {
+                                        $secureXml = $this->securityScanner->scan($this->getFromZipArchive($zip, $relPath));
+                                        $cleanedXml = str_replace('<br>', '<br/>', $secureXml);
```
PowerKiKi: This is way too dangerous and cannot be merged as is.

rdoering: You're right, this solves my problem could turn an correct code like "<br></br>" into an unwanted code like "<br/></br>". Thanks for this hint I will prepare another solution.

rdoering:
May I should change the name of the second method.

rdoering:
Please don't hesitate to contact me, if there still is a change requested.

rdoering:
Do you need more information?

rdoering:
Please let me know, if I could help to push this topic forward.

rdoering:
@PowerKiKi  Is there an issue I can help you with?

rdoering:
I would like to support you with this issue.

rdoering:
I am already here :-)

rdoering:
Anything new here?

rdoering:
Please tell me, if I can help you.

rdoering:
@PowerKiKi 
Thanks for reviewing the code.

The difficulty was, that the function simplexml_load_string() wasn't throw an exception in default config. That is why we added the new wrapper method  simplexml_load_string().

If the method simplexml_load_string() is fails by calling the same function simplexml_load_string() coded as before, then and only then we try to parse the code as HTML-Code instead of pure XML.

`$dom = new DOMDocument;
$rc = $dom->loadHTML($string, $options);`

Unfortunately we rethrow the exception. Thats why you are right, the new code could effect the behaviour and we should remove the rethrow-statement.

> IMHO this is going too far only to allow to load an invalid file. I'd suggest fixing invalid files before loading them with PhpSpreadsheet, or maintain your fork if you encounter many of those files.

Such Excel-files with new-lines in a VML commands are written by Excel 2013 and not really invalid. A VML command contains HTML-code instead pure XML-code. This is mostly the same but not at all. (1)


(1) - https://tools.ietf.org/html/rfc7992#section-9.12

PowerKiKi:
> then and only then we try to parse the code as HTML

That's my problem, one time we parse as XML, but sometimes we parse as HTML. This might create new issues that might be hard to resolve.

Also [Excel 2013 is EOL since 2018-10-04](https://support.microsoft.com/en-us/lifecycle/search/918) for most people. I don't think it is worth adding a risk to our codebase for something that will naturally disappear.

But again, I'll let @MarkBaker have the final word on this.

rdoering:
Any news here?

rdoering:
Don't hasited to ask.

rdoering:
Any progress on this front?

PowerKiKi:
I know this might be frustrating after all the effort you put into this, but I'll close this PR, to avoid giving hope that it might get merged.

The technical solution suggested in this PR brings too much risk and potential maintenance nightmare.

Again, if @MarkBaker has a different opinion, I'll let him re-open/merge it. But for now I'll reject it.

rdoering:
Hi @PowerKiKi @MarkBaker 

I totally agree, it is a big step to switch form parsing comments as HTML instead of XML but, if Microsoft is writing HTML-comments, IMHO it seems to be the right approach :-).

The only risk I see is to not fail to load a file with a malformed HTML-comment, if PHP-method DOMDocument::loadHTML has is buggy. The reason for this is:

- The new approach is limited to VML-handling, it only comes to play if simplexml_load_string() is failing while loading comments.
-  We currently processed tons of file with applied changes, without any error

If you see a potential maintenance nightmare, I would like to support you with this.

Maybe it would be better create an issue-ticket first to get a better view about what how problem looks like?

PR #1190, opened 2019-10-07
Use configured decimal- and thousandsseparator

MacDisein:
This is:

```
- [x ] a bugfix
- [ ] a new feature
```

Checklist:

- [x ] Changes are covered by unit tests
- [x ] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)

Because the formula function FIXED should return a correct result.


MarkBaker:
Than you for the change. Implemented manually to modify style, otherwise fixing the PR was more difficult because it was directly from the master branch

Issue #953, opened 2019-04-03
removeColumn() does not remove last column / removes the wrong column


tomas-eklund:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

When using [removeColumn()](https://phpoffice.github.io/PhpSpreadsheet/master/PhpOffice/PhpSpreadsheet/Worksheet/Worksheet.html#method_removeColumn) to remove the last column in a worksheet I would expect the last column to be removed.

### What is the current behavior?

It is not possible to remove the last column in a worksheet. The last column is emptied, but not removed. 

Additionally, if you index beyond the last column using `removeColumn()` then the last column will be emptied.

### What are the steps to reproduce?

Consider the following example sheet:

.|A|B|C
-|-|-|-
**1**|A1|B1|C1
**2**|A2|B2|C2

Performing `removeColumn('C')` on this sheet will result in:

.|A|B|C
-|-|-|-
**1**|A1|B1|
**2**|A2|B2|

Also `removeColumn('D')` yields the same result, which is even worse as column C should not have been touched at all in this case. 

However, `removeColumn('A')` (or 'B') works as expected:

.|A|B
-|-|-
**1**|B1|C1|
**2**|B2|C2|

I consider this to be two separate bugs, even though they may be related:

1. It should be possible to remove the last column in a worksheet. As of now it is not.
2. Indexing out of bounds should have a well defined and unsurprising behavior. Throwing an exception, returning an error code, issuing a warning, or doing nothing are all acceptable (to me) as long as the behavior is defined and documented. Removing (the content of) a different column than the one requested is not.

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php
require_once 'vendor/autoload.php';
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();
$html_writer = new \PhpOffice\PhpSpreadsheet\Writer\Html( $spreadsheet );
$sheet = $spreadsheet->getActiveSheet();
$remove_column = 'C';
$data = [
    ['A1', 'B1', 'C1'],
    ['A2', 'B2', 'C2'],
];
$sheet->fromArray( $data );
[...]

PR #1197, opened 2019-10-11
Bugfix/remove column out of range

darenas31415:
Fix remove a column out of range removes the last column
```
Given:
+---+---+
| A | B |
+---+---+
When: Attempting to remove 'D'
Then: The worksheet should not be altered
```
The current behaviour is:
```
Given:
+---+---+
| A | B |
+---+---+
When: Attempting to remove 'D'
Then:
+---+
| A |
+---+
```

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

It fixes issue #953

### DIFF
--- a/tests/PhpSpreadsheetTests/Worksheet/WorksheetTest.php
+++ b/tests/PhpSpreadsheetTests/Worksheet/WorksheetTest.php
@@ -181,4 +181,65 @@ class WorksheetTest extends TestCase
             $worksheet->getCell('C1')->getValue()
         );
     }
+
+    public function removeColumnProvider(): array
+    {
+        return [
+            'Remove first column' => [
+                [
+                    ['A1', 'B1', 'C1'],
+                    ['A2', 'B2', 'C2'],
+                ],
+                'A',
+                [
+                    ['B1', 'C1'],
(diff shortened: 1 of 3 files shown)


tests/PhpSpreadsheetTests/Worksheet/WorksheetTest.php
```diff
@@ -181,4 +181,95 @@ public function testRemoveCellsCorrectlyWhenRemovingRow()
             $worksheet->getCell('C1')->getValue()
         );
     }
+
+    public function removeColumnProvider(): array
+    {
+        return [
+            'Remove first column' => [
```
MarkBaker: Nice use of index labelling for test data; I'd only recently discovered this feature of PHPUnit for labelling data from providers myself, and started to apply it in a few places, so it's good to see somebody else using it

darenas31415: Yes, very handy indeed! Not only to quickly see which data set failed in case of any errors but also to know what's being tested. Looking at a random array of values makes it hard for your future self or contributors in this case.

PR #1199, opened 2019-10-14
support "showZeros" setting in Excel advanced worksheet options

### LINKED ISSUE #1196: getCalculatedValue() always returns 0 for NULL references
This is:

```
- [x] a bug report
```

In my spreadsheet I have a reference formula "=A10". 

If A10 doesn't have a value, null should be returned.

0 is returned.

The reason is the code on lines 2815-2817 in Calculation.php:

```
if ($result === null) {
            return 0;
...
```
Why is this line there? Is this correct behaviour?

current

lukey78:
This is:

```
- [x] a bugfix
- [x] a new feature
```
Fixes #1196 

Checklist:

- [x] Code style is respected

If you reference a cell with "=A2" and A2 contains nothing, the result of this calculation is a 0 (number) and not NULL (empty cell) as you would expect.

The default Excel behaviour is to show 0 when a calculation or reference returns null.

This behaviour can be changed with the view setting "Show a zero in cells that have zero value" (see: https://support.office.com/en-us/article/display-or-hide-zero-values-3ec7a433-46b8-4516-8085-a00e9e476b03). 

If this option is set to false (0), then Excel shows an empty cell in these cases.

This pull requests changes the following:

- The normal behaviour is not changed at all
- If you set "showZeros" to false (through WorkheetView->setShowZeros(false)) - OR by working with an Excel file containing this option - then this will happen:
  - if the result of "getCalculatedValue" is null, then null is returned
  - the "showZeros" setting is also saved to the generated XLSX.

It might be debatable if the current behaviour of getCalculatedValue (always convert NULL to 0) is correct.

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/SheetView.php
+++ b/src/PhpSpreadsheet/Worksheet/SheetView.php
@@ -36,6 +36,16 @@ class SheetView
     private $zoomScaleNormal = 100;
 
     /**
+     * ShowZeros.
+     *
+     * If true, "null" values from a calculation will be shown as "0". This is the default Excel behaviour and can be changed
+     * with the advanced worksheet option "Show a zero in cells that have zero value"
+     *
+     * @var bool
+     */
+    private $showZeros = true;
+
+    /**
      * View.
      *
      * Valid values range from 10 to 400.
@@ -113,6 +123,24 @@ class SheetView
         }
 
         return $this;
+    }
+
+    /**
+     * Set ShowZeroes setting.
+     *
+     * @param bool $pValue
+     */
+    public function setShowZeros($pValue)
+    {
+        $this->showZeros = $pValue;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isShowZeros()
+    {
+        return $this->showZeros;
     }
 
     /**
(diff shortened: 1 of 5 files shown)


src/PhpSpreadsheet/Worksheet/SheetView.php
```diff
@@ -115,6 +125,24 @@ public function setZoomScaleNormal($pValue)
...
+     */
+    public function setShowZeros($pValue)
+    {
+        $this->showZeros = $pValue;
+    }
+
+    /**
+     * @return bool
+     */
+    public function isShowZeros()
```
MarkBaker: While the use of `isShowZeroes()` is semantically cleaner; for historic reasons, all equivalent methods use `get`, so the use of `getShowZeros()` is better from the perspective of consistency within the library

lukey78:
> While I haven't yet checked for myself, how does this affect Excel functions like `ISBLANK()` and `ISNUMBER()`. Do they need to temporarily disable the worksheet view setting to return the correct result?

if A1 is a reference to an empty field (e.g. B1, "=B1") then A1 shows a "0" in the default Excel setting.

ISBLANK(A1) returns FALSE
ISNUMBER(A1) returns TRUE

With deactivating the Excel setting "Show a zero in cells that have zero value" the cell A1 shows "" (nothing). The formulas above return the same output.

PR #1206, opened 2019-10-15
Fixed Functions::ifCondition for allowing empty condition ("") and different than empty condition ("<>")

jorgecasas:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected
- [] CHANGELOG.md contains a short summary of the change if necessary
- [] Documentation is updated as necessary

In cells with formulas containing conditions like `=IFSUM(A1:A3;"";B1:B3)` to sum cells from range A1:A3 with empty value in range B1:B3, the function `Functions::ifCondition()` create in this case the code `=""""` instead of `=""`, so it didn't work.

Added test to check condition empty (`""`) and also condition different value than empty (`"<>"`)


PowerKiKi:
Thanks for the PR !

PR #1208, opened 2019-10-15
Remove shrikRangeToFit call for Reader in Data Validations and fix alerts

yurigeorgiev4:
This is:

```
- [ x ] a bugfix
- [ ] a new feature
```

Checklist:

### Background
When trying to populate an **xlsx** template downloaded from Amazon SellerCentral, the Data Validations are not setup correctly on the result file.

 https://sellercentral.amazon.com/gp/help/external/202094740?language=en_US&ref=efph_202094740_cont_201576440

Fixes issue that the Data Validation Prompt and Error messages are not displayed when the spreadsheet is being read and then saved.

Fixes issue that when a Data Validation has been setup for a range of rows let's say B2:B25, if the content of the sheet is to row 5, the validations from B5:B25 would be wiped.


oleibman:
Equivalent change was merged with PR #2225 in July 2021.

PR #1211, opened 2019-10-17
Do not export .github folder

ravage84:
Just came across this. I do not think this is necessary. Keeps the built artifacts smaller and tidy.


MarkBaker:
You're right, thanks for the PR

PR #1240, opened 2019-11-14
Update IOFactory.php

Lablnet:
- [ x] The key of `reader` and `writer` should be lowercase

Checklist:

- [ x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)

Yesterday i integrated this with laravel, and i the `xslx` file was uploaded by user, i just use laravel build in function to get extension, but i have to use `ucfirst()` function to make first letter upper case, its not good, now whenever user provide the key, it converted to lowercase, and in keys table keys are lowercase so no problem will happen.

### DIFF
--- a/src/PhpSpreadsheet/IOFactory.php
+++ b/src/PhpSpreadsheet/IOFactory.php
@@ -13,25 +13,25 @@ use PhpOffice\PhpSpreadsheet\Shared\File;
 abstract class IOFactory
 {
     private static $readers = [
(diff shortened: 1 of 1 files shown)


PowerKiKi:
Being case-sensitive is a feature here. We do not want to allow for perceived magic. The writer is called "Xlsx" not "xlsx", nor "xLSx".

I suggest you consider using `IOFactory::load()` to have the simplest most flexible way to load a file.

PR #1242, opened 2019-11-14
Update Csv.php

SwithFr:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

Occasionally, a temporary Csv file doesn't have extension and mime type returned is 'text/x-Algol68'.


oleibman:
See https://stackoverflow.com/questions/70004695/codeigniter-csv-file-upload-fails-due-to-mime-type-text-x-algol68 for a better description of the problem.

Php7.3- identified the file as text/x-Algol68. However Php7.4+ identify the file as text/plain. So, although the fix is easy, it is no longer required (PhpSpreadsheet currently requires Php8.1+).

Issue #1241, opened 2019-11-14
syntax error  Calculation.php 


sevgicorumlu:
Parse error: syntax error, unexpected '=' in 
koolreport/excel/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Calculation/Calculation.php on line 2460
->   [$language] = explode('_', $locale);

MAKS-dev:
Current online version code is only PHP 7.1 compatible:
```
            [$language] = explode('_', $locale);
```
Older version was (which is still supported in PHP):
```
            list($language) = explode('_', $locale);
```
https://www.php.net/manual/en/function.list.php
> **Symmetric array destructuring**
> The shorthand array syntax ([]) may now be used to destructure arrays for assignments (including within foreach), as an alternative to the existing list() syntax, which is still supported.

I don't like that this makes the code only support PHP7.1+ there is no actual need. Yes I know that support for older PHP versions are long dropped, but it is still wide spread.

Anyway @sevgicorumlu if this is the only error, you can change the line of code for now. But best is to upgrade to PHP7.3 all together ;-)

sevgicorumlu:
thank you :)

PR #1243, opened 2019-11-14
Save the PHP 7.0.x compatibility

Thaldos:
This is:

```
- [ ] a bugfix
```

Checklist:

- [ ] Code style is respected

In PHP 7.0.x the Calculation.php file throws the error :
Parse error: syntax error, unexpected '=' in phpspreadsheet/src/PhpSpreadsheet/Calculation/Calculation.php on line 2460

You can see on this snippet : http://sandbox.onlinephpfunctions.com/code/12d20ff992200b59c2a62341ba74eec5a2b951d2

With my modifications, it works gracefully on PHP 7.0.x

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,5 +1,5 @@
 {
-    "name": "phpoffice/phpspreadsheet",
+    "name": "thaldos/phpspreadsheet",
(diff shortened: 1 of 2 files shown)


MAKS-dev:
Why not use the original code? `list` is a better solution here.
For details see: https://github.com/PHPOffice/PhpSpreadsheet/issues/1241#issuecomment-553842593

PR #1245, opened 2019-11-17
adjust to excel floor func behavior

eltociear:
Excel FLOOR() function accept negative number and negative significance, but this function accept only positive significance.

```
- [ ] a bugfix
```

Checklist:

- [x] Code style is respected


eltociear:
thanks reply! and may be test case have mistake.

PR #1255, opened 2019-11-25
Calculation formula correction

CSXing:
When the formula cell returns an error value, throwing an exception causes the read to terminate, hoping to return a null value instead of the exception hint.

This is:

```
- [ ] a bugfix
```

Checklist:

- [ ] Code style is respected


CSXing:
When the formula cell returns an error value, throwing an exception causes the read to terminate, hoping to return a null value instead of the exception hint.

Issue #1250, opened 2019-11-22
Xls writer, on save, sets selected cell in sheet to last parsed cell with formula and cannot set active sheet (selects multiple sheets when available).


jimmy4o4:
Xls writer, on save, sets selected cell in sheet to last cell with formula and cannot set active sheet (selects multiple sheets when available).

This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
1. On save() with Xls writer, selected cell(s) should remain selected and selection should not change.
2. On save() with Xls writer, active sheet should remain active. Only one (active), not multiple sheets should be selected in output file.

### What is the current behavior?
1. On save() with Xls writer, cell(s) selection changes to last parsed sheet cell that contain formula;

This is caused by running \PhpOffice\PhpSpreadsheet\Cell\Cell::getStyle() method (in fact \PhpOffice\PhpSpreadsheet\Worksheet\Worksheet::getStyle()) in \PhpOffice\PhpSpreadsheet\Calculation\Calculation::_calculateFormulaValue() method.

2. On save() with Xls writer, output file has selected multiple sheets and active is not the one that was made active before save. Using LibreOffice Calc 6.3.2.2 x86 to open output file.

### What are the steps to reproduce?

```php
<?php

require_once 'vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

$spreadsheet->setActiveSheetIndex(0)
    ->setTitle('Test1')
    ->setCellValue('D1', 1)
    ->setCellValue('D2', 2)
    ->setCellValue('D3', 3)
    ->setCellValue('D4', 4)
    ->setCellValue('D5', '=SUM(D1:D4)')
    ->setSelectedCell('A1');

$spreadsheet->createSheet(1);

$spreadsheet->setActiveSheetIndex(1)
    ->setTitle('Test2')
    ->setCellValue('D1', 4)
    ->setCellValue('E1', 3)
    ->setCellValue('F1', 2)
    ->setCellValue('G1', 1)
    ->setCellValue('H1', '=SUM(D1:G4)')
    ->setSelectedCell('A1');

$spreadsheet->setActiveSheetIndex(0);

[...]

PR #1256, opened 2019-11-25
Fix for Xls writer wrong cells and sheet selection.

jimmy4o4:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

See #1250 .


jimmy4o4:
Fix for Xls BIFF window2 structure was needed too.

I've added some tests too.

All of this covers only problems with Xls. Ods and Xlsx are still affected, I'm not sure, if there is any possible solution for these file types.

PR #1262, opened 2019-11-25
Modify XLSX RW to keep decimal for floats with a zero decimal part

rtek:
This is:
- [x] a bugfix

```
Prior to 1.10, all numeric values where read as floats. In 1.10
numeric values are read using 0 + x, which relies on PHP type
juggling rules. As a result, float(0.0) is written as string('0'),
then read back as int(0). This fix causes the writer to retain the
the decimal for float values such that a reader can differentiate
floats from ints.
```

Checklist:

- [x] Code style is respected


PowerKiKi:
Thanks for this great PR !

PR #1270, opened 2019-12-03
fix customer conditional format parse bug.

ponyhu:
This is:

```
- [* ] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

in Xlsx.php row:570 parse and set customer conditional format, styleXml->dxfs->dxf is SimpleXMLElement array, so this need process it.


ponyhu:
I think this bug need fix in other place, so close this pull request and push new.

PR #1277, opened 2019-12-07
Detect if cell is text format

wvfeng:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Text format cells do not need to be converted. Automatic recognition of the type will cause some cells set to text type to be distorted, which is not common sense.


oleibman:
Justification does not contain enough detail - how will a cell be distorted? On the other hand, this change may well cause problems for cells which contain a formula or rich text.

Issue #1281, opened 2019-12-10
Xlsx Writer corrupts decimal values with commas in them


makedin:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Decimal values should be represented correctly (like "1.1") in Xlsx files.

### What is the current behavior?
If PHP has been set to use a locale that uses comma as decimal separator, any decimal values (floats, real numbers) with a decimal part become corrupted: "1,1" (1.1) for instance becomes "1.1.0".

Note that LibreOffice Calc silently fixes values like these. Neither Microsoft Excel nor Google Sheets handle these values as decimals (Excel gives a warning and converts the value to a string. Sheets drops the value silently). PhpSpreadsheet itself silently converts the value to a string.

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php
require __DIR__ . '/vendor/autoload.php';

if (!setlocale(LC_ALL, 'fi_FI.UTF-8')) {
  die("Locale not available, this isn't going to work\n"); 
} 

$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();
$spreadsheet->getActiveSheet()->setCellValue('A1', 1.1);

$filename = 'decimalcomma.xlsx';
$writer = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet);
$writer->save($filename);          

$reader = new \PhpOffice\PhpSpreadsheet\Reader\Xlsx();
$spreadsheet = $reader->load($filename);
echo($spreadsheet->getActiveSheet()->getCell('A1')->getValue());

```

### Which versions of PhpSpreadsheet and PHP are affected?
The bug seems to have been introduced by cf30c2a82420789a5ff5cce37082584d27e204e4. It should manifest with any supported PHP version.

MarkBaker:
Thank you for highlighting this issue, and for the resolution

PR #1282, opened 2019-12-10
Fix Xlsx Writer's handling of decimal commas

makedin:
Perform the comma -> period substitution in Xlsx Writer writeCell()
before checking if the decimal value has a period in it. The previous
behavior led to decimals of the form "1,1" to become "1.1.0".

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

See #1281


MarkBaker:
Thanks you for this solution to the locale problem

Issue #667, opened 2018-09-10
Writing to and saving existing xlsx file.


unknown-opensource:
This is a bug report for 1.4.0

### What is the expected behavior?
      Libreoffice (Version: 6.0.3.2 Build ID: 1:6.0.3-0ubuntu) opens file without warning.

### What is the current behavior?
      Warning loading document output.xlsx: The data could not be loaded completly because the maximum number of columns per sheet was exceeded.

### What are the steps to reproduce?
        $template = "template.xlsx"; 
        // (https://github.com/PHPOffice/PhpSpreadsheet/files/2365395/template.xlsx) - empty file
        $spreadsheet = \PhpOffice\PhpSpreadsheet\IOFactory::load($template);
        $writer = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet);
        $writer->save("output.xlsx");

GreatHumorist:
@unknown-opensource @MAKS-dev I tried in my computer, there is no error occurred, and the output.xlsx file is successfully generated.

MAKS-dev:
Sorry ... I'm not conttibuting here on the issue nor (tried to) reproduce it.
Just pointed out where to find the file ;-)

damijanc:
This is not a library bug. Libre office supports max 1024 columns while Excel support 16384.  You can read all about it https://bugs.documentfoundation.org/show_bug.cgi?id=50916.

unknown-opensource:
@damijanc so my empty file has too many columns?

damijanc:
No, there is nothing wrong with your file. You will get the warning because excel supports more columns than open office does. As long as you do not generate excel files with more than 256 columns of data your open office users will be fine.

unknown-opensource:
@damijanc it is interesting how template file (https://github.com/PHPOffice/PhpSpreadsheet/files/2365395/template.xlsx) opens without error before processing with PhpSpreadsheet

damijanc:
Maybe the file metadata is wrong. It would be interesting to check.

PowerKiKi:
Not a PhpSpreadsheet bug, but a limitation of LibreOffice

Yopai:
I don't understand why you say "this is a limitation of LibreOffice"

dericlima:
I agree with @Yopai , while exporting with PHPExcel something is changing and causing this warning.

[...]

PR #1289, opened 2019-12-13
fix for `maximum number of columns per sheet was exceeded` in LibreOffice.

bnowak:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

XLSX files saved in LibreOffice are not correctly loaded by the reader.  
LibreOffice saves files with metadata for 1025 columns and reader consider only 16384 columns from MS Office. It leads to mapping all of trailing empty columns in Libre's file. Next, when we've saved the file by writer and trying to open it in Libre the warning `The data could not be loaded completely because the maximum number of columns per sheet was exceeded.` occurs. It's because Libre handle only 1024 columns.  
This fix added handling for trailing metadata columns from LibreOffice files.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx/ColumnAndRowAttributes.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx/ColumnAndRowAttributes.php
@@ -136,7 +136,7 @@ class ColumnAndRowAttributes extends BaseParserClass
             for ($columnAddress = $startColumn; $columnAddress !== $endColumn; ++$columnAddress) {
                 $columnAttributes[$columnAddress] = $this->readColumnRangeAttributes($column, $readDataOnly);
 
-                /**
+                /*
                  * LibreOffice saves xlsx files with 1025 columns mapping,
                  * while MS Office saves it with 16384 columns mapping.
                  */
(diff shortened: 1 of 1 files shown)


bnowak:
It's related to #584 and #667

giuliotripi:
Your commit worked for me too

bnowak:
There is any chance to merge it?

stfl:
+1

sunsetsystems:
I had this problem and the fix worked for me (manually ported in to 1.10.1).

bnowak:
up

cylosh:
Was this merged?

bnowak:
nope

eclimov:
Up

PowerKiKi:
This cannot be merged, because it would be a loss of data for valid Excel files that happens to have one metadata ending on columns 1025 and then have some more. We should not limit data that are read if PhpSpreadsheet model can support it. Instead we should limit data that are written according to the format specification.

And because XLSX format can support more than 1025 columns, there is no reason to have this limit in our writer either. Instead if you think your files might be read by LibreOffice, then you should manually delete the extra data before writing the file.

stuaxo:
This seems like a bug in libreoffice, is there an upstream bug for it ?

PR #1292, opened 2019-12-16
Support writing to resource handles in all IWriter implementations

basbl:
**Looking for feedback**

This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [ ] Code style is respected

Some writers already had support for writing to resource handles passed through `IWriter::save`. The ones preventing this from becoming a supported feature among all of them were the XLSX and the ODS writers which are zipped collections of files and `\ZipArchive` does not support writing to streams.

By taking on an extra dependency, namely `maennchen/zipstream-php`, this could be supported.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Csv.php
+++ b/src/PhpSpreadsheet/Writer/Csv.php
@@ -4,6 +4,7 @@ namespace PhpOffice\PhpSpreadsheet\Writer;
 
 use PhpOffice\PhpSpreadsheet\Calculation\Calculation;
 use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Writer\Exception as WriterException;
 
 class Csv extends BaseWriter
 {
@@ -77,7 +78,7 @@ class Csv extends BaseWriter
     /**
      * Save PhpSpreadsheet to file.
      *
-     * @param string $pFilename
+     * @param resource|string $pFilename
      *
      * @throws Exception
      */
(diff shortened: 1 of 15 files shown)


rosstuck:
Just wanted to throw in that this would be super useful since it would allow folks to generate XLSX and other formats without needing write access to a temp directory anymore. If there's anything we can do to help nudge this along, please say the word, we're happy to put in the work on it. :heart:

PowerKiKi:
This seems to be a move in the right direction. @basbl, you created this PR as a draft. What else did you plan to do before completion ?

basbl:
Mainly just gathering feedback and fixing the build, @PowerKiKi

PowerKiKi:
Let us know when you feel it can be reviewed, or if you need help from the community for something.

Also, feel free to add typings in existing code when it is not breaking, and definitely add them for all new code.

basbl:
@PowerKiKi I could use some help with getting the final Scrutinizer issues resolved. There's also a false positive in there like:

IssueId: 42375536  Message: It seems like ``$pFilename`` can also be of type ``resource``; however, parameter ``$filename`` of ``fopen()`` does only seem to accept ``string``, maybe add an additional type check?  Filename: src/PhpSpreadsheet/Writer/Ods.php  LineNumber: 108  Link: https://scrutinizer-ci.com/g/PHPOffice/PhpSpreadsheet/inspections/609cbf0c-f57a-46f7-b13b-2b06d52f5e5e/issues/files/src/PhpSpreadsheet/Writer/Ods.php?status=new&orderField=path&order=asc&honorSelectedPaths=0&issueId=42375536

src/PhpSpreadsheet/Writer/Csv.php
```diff
@@ -126,6 +132,8 @@ public function save($pFilename)
         }
 
         // Close file
+        rewind($fileHandle);
```
PowerKiKi, line 135: Why did you choose to rewind here and in other writers ?

PowerKiKi:
Thanks ! I completed it a bit and merge it.

basbl:
Thanks a bunch!

PR #1316, opened 2020-01-13
Changes to WEEKNUM and YEARFRAC

oleibman:
The optional second parameter for WEEKNUM can take any of 10 values
(1, 2, 11-17, and 21), but currently only 1 and 2 are supported.
This change adds support for the other 8 possibilities.

YEARFRAC in Excel does not require that end date be before start date,
but PhpSpreadsheet was returning an error in that situation.

YEARFRAC third parameter (method) of 1 was not correctly implemented.
I was able to find a description of the algorithm, and documented
that location in the code, and implemented according to that spec.
PHPExcel had a (failing) test to assert the result of
YEARFRAC("1960-12-19", "2008-06-28", 1). This test had been dropped
from PhpSpreadsheet, and is now restored; several new tests have
been added, and verified against Excel.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

As described above, extend support for WEEKNUM, and correct some problems in YEARFRAC.


src/PhpSpreadsheet/Calculation/DateTime.php
```diff
@@ -1301,31 +1313,34 @@ public static function WEEKNUM($dateValue = 1, $method = 1)
 
         if (!is_numeric($method)) {
             return Functions::VALUE();
-        } elseif (($method < 1) || ($method > 2)) {
-            return Functions::NAN();
         }
-        $method = floor($method);
+        $method = (int) $method;
+        $methodarr = [1 => 1, 2, 11 => 2, 3, 4, 5, 6, 7, 1, 21 => 21];
```
MarkBaker: Would this array be better as a static property, or even a constant array; and perhaps with name constants for the keys to provide a human-readable explanation of the $method names that can then be use in the body of the code?

oleibman: Recoded array as constant (didn't even know that was possible), plus other named constants per this review comment.

MarkBaker:
Thanks for the fixes, and for the improvements. The contribution is much appreciated

PR #1321, opened 2020-01-14
[BC break/RFC] Fix default cell type treatment as numeric for XLSX

rtek:
SpreadsheetML defines cell type attribute t to indicate the cell type.
The attribute is optional and defaults to "n" for numeric. The XLSX
Reader incorrectly defaults the cell type to "stringy" for both
undefined t attribute and t="n".

This is:

```
- [X] a bugfix
- [] a new feature
```

Checklist:

- [x] Code style is respected

SpreadsheetML defines the default cell type as numeric, but the XLSX reader treats a missing cell type or cell type = numeric as a "stringy" value. This means that the reader is discarding the numeric type information from the XML. In most cases this is masked by the `DefaultBinder` which explicitly preg's for numeric strings and correct the reader's behavior. 

The bugs / inconsistencies are:

1. Strings that pass the numeric preg are incorrectly casted to numeric
2. The`IValueBinder` is always presented with string, even when the XML defines the type as numeric
3. A numeric cell that contains non-numeric information is not treated as an exception

This is probably a BC break as it seems like a fundamental issue with the Reader - I would expect there is lots of code relying on numeric strings being casted to a number despite the XML declaring them as strings.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -3,6 +3,7 @@
 namespace PhpOffice\PhpSpreadsheet\Reader;
 
 use PhpOffice\PhpSpreadsheet\Cell\Coordinate;
+use PhpOffice\PhpSpreadsheet\Cell\DataType;
 use PhpOffice\PhpSpreadsheet\Cell\Hyperlink;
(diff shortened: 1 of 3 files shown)


oleibman:
Superseded by PR #4139.

PR #1323, opened 2020-01-15
Active Cell Anywhere

oleibman:
When freeze pane is in use on a worksheet, PhpSpreadsheet saves to Xlsx in such
a way that the active cell is always set to the top left cell below the freeze
pane. I find it difficult to understand why:
  1. You have given users the setSelectedCells function, but then choose to
     ignore it.
  2. Excel itself does not act in this manner.
  3. PHPExcel did not act in this manner.
  4. PhpSpreadsheet when writing to Xls does not act in this manner.
     This is especially emphasized because the one test in FreezePaneTest which
     would expose the difference is the only test in that member which is
     not made for both Xls and Xlsx.
  5. It is *really* useful to be able to open a spreadsheet anywhere, even when
     it has header rows.

Rather than create regression problems by changing this behavior all the time,
I instead introduce a new method, setActiveCellAnywhere, to
Writer/Xlsx/Worksheet to control it via a private variable, with the default
(false) being the current behavior. When it is set to true, the active cell
is left as the user intended.

The documentation is updated to describe the new method.
New tests are added, and the Xlsx-only test is changed to test
both Xls and Xlsx.

This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

As described above, add the ability to open a worksheet anywhere even when freeze pane is used.

### DIFF
--- a/docs/topics/reading-and-writing-to-file.md
+++ b/docs/topics/reading-and-writing-to-file.md
@@ -150,20 +150,6 @@ $writer = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet);
 $writer->save("05featuredemo.xlsx");
 ```
 
-#### Active cell
-
-By default, this writer will set the active cell on any worksheet which
-uses freeze panes to the top left cell following the freeze pane.
-You can instead set the active cell to be anywhere on the sheet
-with the Worksheet setSelectedCells method
-in conjuction with the following:
-
-``` php
-$writer = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet);
-$writer->setActiveCellAnywhere(true);
-$writer->save("05featuredemo.xlsx");
-```
-
 #### Formula pre-calculation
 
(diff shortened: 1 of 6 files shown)


oleibman:
Change simplified per request.
The Travis CI build failed, but I do not know why. There is no explanation.
Examining the job log, I see the following message around line 10107:
Version master
  Parsing ############### 31%                                    1 error
          PhpOffice\PhpSpreadsheet\Calculation\Calculation
However, (a) it does not say what the error is (it did not stop all the other tests from passing),
and (b) this change does not touch PhpOffice\PhpSpreadsheet\Calculation\Calculation.
So, if you can let me know what needs to be done here, I can try to take care of it,
but, if not ...

oleibman:
I noticed that Calculation had been changed 3 days ago. Manually applying that simple change to my pull request eliminated the Travis CI problem. I don't know if that was the best way to solve the problem or not, but it worked.

PowerKiKi:
Thanks, merged in fb37938

Issue #1325, opened 2020-01-15
The WorksheetXml autofilter['ref'] could return a filter range containing $ as character 


astronati:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
The "$" character should be removed from the filter range.

### What is the current behavior?
The `setRange` method of the `Xlsx/AutoFilter` class expects a filter range format like "A1:E10".
The returned value from `$this->worksheetXml->autoFilter['ref']` could contain "$" and returning a value like "$A$1:$E$10".

### Which versions of PhpSpreadsheet and PHP are affected?
1.10.1

PR #1326, opened 2020-01-15
ref #1325 Removed all '$' values from the auto filter range

astronati:
See https://github.com/PHPOffice/PhpSpreadsheet/issues/1325

This is:

```
- [x] a bug report
- [ ] a feature request
```

The "$" character should be removed from the filter range.

The `setRange` method of the `Xlsx/AutoFilter` class expects a filter range format like "A1:E10".
The returned value from `$this->worksheetXml->autoFilter['ref']` could contain "$" and returning a value like "$A$1:$E$10".

1.10.1


astronati:
@PowerKiKi Added test

Issue #1161, opened 2019-09-16
RESOLVING + RESOLVING2 constants (Calculation/Calculation.php:4148)


otuhacek:
Hi,
I just went into some problem when exporting some content on AC99 field. It complains of redefining RESOLVING2 constant multiple times, which triggered PHP Notice. Is this piece of code meant to be there or is it just something from previous testing?

https://github.com/PHPOffice/PhpSpreadsheet/blob/a972943ac56895905d9cbaf103d17e9e25fb8b50/src/PhpSpreadsheet/Calculation/Calculation.php#L4148

Thanks

![resolving](https://user-images.githubusercontent.com/55390565/64955374-5da8b480-d888-11e9-88a4-4e8d13b515b9.png)

pedrofsantoscom:
I also have the same notice. Besides the notice it runs without a problem.
This happens on a file created that has 22-29 sheets, and on each of those sheets column A has ~1000 rows and each row on col A has a very long expression (7000 characters long).

![image](https://user-images.githubusercontent.com/2325978/68167768-91d25480-ff5e-11e9-8f24-dcac66fcd9c1.png)

jonnylink:
This is still an issue

jonnylink:
This is from this commit: https://github.com/jonnylink/PhpSpreadsheet/commit/0b387e767e99a9042886ea673d9b9905c7951e87

It does in fact look like code that was simply there to test, since there are no other references to the constant RESOLVING or RESOLVING2 in the codebase. I'm going to make a PR to remove that broken code.

Benjiiii:
Facing same issue

woganmay:
Just had this come up in production now for the first time. Definitely still an issue.

jonnylink:
Might want to comment or thumbs up the PR if you're having this issue. Not sure anyone is checking closed issues.

https://github.com/PHPOffice/PhpSpreadsheet/pull/1328

oleibman:
Fixed by PR #1328 in February 2020.

PR #1328, opened 2020-01-16
BUGFIX for issue #1161

jonnylink:
This is: 

```
- [x] a bugfix (#1161)
- [ ] a new feature
```

Checklist:

- [x] Changes are covered by unit tests — No. Removed code that was never covered in tests and shouldn't have been there in the first place. Nothing to test.
- [x] Code style is respected
- [x] Documentation is updated as necessary — no not necessary

As noted in [#1161](https://github.com/PHPOffice/PhpSpreadsheet/issues/1161) there was some test code left in the calculation.php file. This would only show up on larger files, so it hasn't been prevalent, but it should be removed. There are no references to the constants being set. Even if they were, it's bad practice to set constants in code. Here it's in a loop and also in a simple if/else block.


mostafatalebi:
May I know when we are going to have this commit released? (it is merged with master but not release-tagged).

jonnylink:
Yeah, my team is running into this bug again. Would be great to have it released

PowerKiKi:
That was released in https://github.com/PHPOffice/PhpSpreadsheet/releases/tag/1.11.0 back in March.

PR #1332, opened 2020-01-18
Bugfix for Issue 1271

oleibman:
Summarizing:
MATCH with a static array should return the position of the found value based on the values submitted.
Returns #N/A, unless the element searched for is at the end of the array.

The problem is in Calculation.php line 4231:
                    if (!is_array($functionCall)) {
                        foreach ($args as &$arg) {
                            $arg = Functions::flattenSingleValue($arg);
                        }
                        unset($arg);
                    }

I believe this code is intended to handle functions where PhpSpreadsheet just passes
the call on to PHP without implementing the code on its own, e.g. for atan or acos.
In the bug report, the following code fails:
  $flat_rate = "=MATCH(6,{4,5,6,2}, 0)";
  $sheet->getCell('A1')->setValue($flat_rate);
The expected value is 3, but the actual result is "#N/A".
The reason for this result is that the parser replaces the braces with calls
to the MKMATRIX internal function, whose value for functioncall was:
'self::MKMATRIX'. Since this isn't an array, the flattening code is executed,
and the unintended result occurs. The fix is to change the definition for
functioncall in that case to [__CLASS__, 'mkMatrix'], avoiding the flattening.

However, there is also another part to this bug. The flattening should be
returning the first entry in the array, but is in fact returning the last.
This explains why the bug report specified "unless ... end of the array".
I confirmed that Excel does use the first item in the array rather than the last,
e.g. =atan({1,2,3}) entered into a cell will return atan(1), not atan(3).
The problem here is that flattenSingleValue, which says in its comments that
it is supposed to be returning the first item, uses array_pop rather than array_shift.
I have changed that as well. The same mistake was also present in
Cell.php function getCalculatedValue. The correct behavior can be verified
by entering =minverse({-2.5,1.5;2,-1}) into an Excel cell'
Excel flattens the result ({2,3;4,5}) to 2, and so should PhpSpreadsheet.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Bugfix, as described.

### DIFF
--- a/src/PhpSpreadsheet/Calculation/Functions.php
+++ b/src/PhpSpreadsheet/Calculation/Functions.php
@@ -646,7 +646,7 @@ class Functions
     public static function flattenSingleValue($value = '')
     {
         while (is_array($value)) {
-            $value = array_pop($value);
+            $value = array_shift($value); // was array_pop
         }
 
         return $value;
--- a/src/PhpSpreadsheet/Calculation/Functions.php
+++ b/src/PhpSpreadsheet/Calculation/Functions.php
@@ -646,7 +646,7 @@ class Functions
     public static function flattenSingleValue($value = '')
     {
         while (is_array($value)) {
-            $value = array_shift($value); // was array_pop
+            $value = array_shift($value);
         }
 
         return $value;
--- a/src/PhpSpreadsheet/Cell/Cell.php
+++ b/src/PhpSpreadsheet/Cell/Cell.php
@@ -266,7 +266,7 @@ class Cell
                 //    We don't yet handle array returns
                 if (is_array($result)) {
                     while (is_array($result)) {
-                        $result = array_pop($result);
+                        $result = array_shift($result); // was array_pop
                     }
                 }
             } catch (Exception $ex) {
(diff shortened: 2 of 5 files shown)


oleibman:
I don't entirely understand the process here. As I have submitted other pull requests which have been merged, perhaps something is wrong with this one, but I don't know what. There is a definite bug in the existing code, and this change fixes it. All tests have passed, and there have been no comments requesting changes. If you could let me know what is missing, I'll be glad to try and fix it.

src/PhpSpreadsheet/Cell/Cell.php
```diff
@@ -266,7 +266,7 @@ public function getCalculatedValue($resetLog = true)
                 //    We don't yet handle array returns
                 if (is_array($result)) {
                     while (is_array($result)) {
-                        $result = array_pop($result);
+                        $result = array_shift($result); // was array_pop
```
PowerKiKi, line 269:
```suggestion
                        $result = array_shift($result);
```

src/PhpSpreadsheet/Calculation/Functions.php
```diff
@@ -646,7 +646,7 @@ public static function flattenArrayIndexed($array)
     public static function flattenSingleValue($value = '')
     {
         while (is_array($value)) {
-            $value = array_pop($value);
+            $value = array_shift($value); // was array_pop
```
PowerKiKi, line 649:
```suggestion
            $value = array_shift($value);
```

PowerKiKi:
Merged in 3fa624a1c9df19c181b0dae78d9fcd4f7edba0b2

PR #1340, opened 2020-01-30
update doc on how to run the migrate tool

Drivez:
as per this https://stackoverflow.com/questions/53215313/phpspreadsheet-how-do-i-use-auto-migration-tool clarify how to run the tool

This is:

```
- [ X ] Update doc
- [ ] a bugfix
- [ ] a new feature
```

Checklist:

- [ - ] Changes are covered by unit tests
- [ X ] Code style is respected
- [ X ] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ X ] CHANGELOG.md contains a short summary of the change
- [ X ] Documentation is updated as necessary

To better clarify how to run the migration tool

### DIFF
--- a/docs/topics/migration-from-PHPExcel.md
+++ b/docs/topics/migration-from-PHPExcel.md
@@ -14,7 +14,7 @@ installed with composer, it can be run like so:
 
 ``` sh
 cd /project/to/migrate/src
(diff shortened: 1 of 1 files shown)


PowerKiKi:
Because the script contains a shebang, prepending the command with `php` should not be necessary. But it doesn't hurt either and might help in some cases...

Thanks !

PR #1343, opened 2020-02-03
Added support for the ARABIC excel function

paulkned:
This is:

```
- [ ] a bugfix
- [ x ] a new feature
```

Checklist:

- [ x ] CHANGELOG.md contains a short summary of the change


MarkBaker:
Note that I've merged to master "as is". However, this code doesn't handle valid values like `MIM` for `1999`

paulkned:
@MarkBaker good point. I'll submit a new pull request for that. If I'm building something, why not build it correctly :)

MarkBaker:
There's other unusual short-forms; MS Excel's ROMAN() function has a second argument of type (which we don't actually support in our implementation), but which does allow you to see the short-form values

PR #1354, opened 2020-02-07
BIFF8 E_NOTICE undefined offset fix

aternowy:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Sheet View Settings Block should be 8-14 bytes long in BIFF8 Excel 97 according to the open office file format documentation. However access to byte 10 and 12  is not possible when record data is malformed, so  `getUInt2d`  throws notice.
image

### Summary 

Test access to byte offsets for "cached magnification factor" before using it. Use default values otherwise to avoid notices

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xls.php
+++ b/src/PhpSpreadsheet/Reader/Xls.php
@@ -4377,11 +4377,22 @@ class Xls extends BaseReader
             // offset: 10; size: 2; cached magnification factor in page break preview (in percent); 0 = Default (60%)
             // offset: 12; size: 2; cached magnification factor in normal view (in percent); 0 = Default (100%)
             // offset: 14; size: 4; not used
-            $zoomscaleInPageBreakPreview = self::getUInt2d($recordData, 10);
+            if (!isset($recordData[10])) {
+                $zoomscaleInPageBreakPreview = 0;
+            } else {
+                $zoomscaleInPageBreakPreview = self::getUInt2d($recordData, 10);
+            }
+
             if ($zoomscaleInPageBreakPreview === 0) {
                 $zoomscaleInPageBreakPreview = 60;
             }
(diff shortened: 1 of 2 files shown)


src/PhpSpreadsheet/Reader/Xls.php
```diff
@@ -4367,12 +4367,12 @@ private function readWindow2()
             // offset: 10; size: 2; cached magnification factor in page break preview (in percent); 0 = Default (60%)
             // offset: 12; size: 2; cached magnification factor in normal view (in percent); 0 = Default (100%)
             // offset: 14; size: 4; not used
-            $zoomscaleInPageBreakPreview = self::getUInt2d($recordData, 10);
-            if ($zoomscaleInPageBreakPreview === 0) {
+            $zoomscaleInPageBreakPreview = 60;
+            if (isset($recordData[10]) && ($zoomscaleInPageBreakPreview = self::getUInt2d($recordData, 10)) === 0) {
```
PowerKiKi, line 4371: Please avoid having assignment in condition and squash all the commit into one.

eth8505:
@PowerKiKi I remember we actually had the assignments outside of the the condition before but it failed the cyclomatic complexity check. To prevent having to do a larger refactoring we decided to move the assignment into the condition, epecially since this is only a small fix preventing a notice.

If you really think it's necessary we will of course try to find a solution acceptable for everyone, please let me and @aternowy know how to proceed.

PowerKiKi:
The best would be to refactor, but if we cannot do that, we should at least avoid assignment in if condition and accept a cyclomatic complexity check fail.

aternowy:
@PowerKiKi thank you for response, we will add fix to this PR ASAP

eth8505:
@PowerKiKi i finally got around to this. I have removed the assignments in the conditions which now causes the cyclomatic complexity problem as I said before.

PR #1358, opened 2020-02-11
XLSX: always write code names

### LINKED ISSUE #1357: XLSX writer: code names are not always respected
This is:

```
- [ ] a bug report
- [x] a feature request
```

If a spreadsheet has worksheets with a code names, then code names are respected and written to the generated file.

Code names are not written to the generated file unless the spreadsheet has any macros.

```php
<?php

$spreadsheet->getActiveSheet()->setCodeName($expected = 'foo');

$writer = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($spreadsheet, 'Xlsx');
$writer->save($path = tempnam(sys_get_temp_dir(), 'example'));

$reader = PhpOffice\PhpSpreadsheet\IOFactory::createReader('Xlsx');

$actual = $reader
    ->load($path)
    ->getActiveSheet()
    ->getCodeName()
;

if ($actual !== $expected) {
    printf("Expected code name of active sheet to be 'foo', got '%s'.\n", $actual);
    exit(1);
}
```

* PHP: any version
* PhpSpreadsheet: 1.10.1

----

Dear maintainer, would you consider always setting the code name regardless of whether macros exist in the spreadsheet or not? Currently [this block](https://github.com/PHPOffice/PhpSpreadsheet/blob/082266a/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php#L142-L148) controls whether or not to add code names to the generated file. I would propose removing the if-guard so that code names would *always* be written.

The specific use case I have is to provide the user with a template, which they could fill and then submit back to the same system. Using code names would make the template more flexible and resistant to changes to worksheet names (i.e. downloading a template in one locale & submitting it back in a different one).

Would you accept such a contribution, if I made a PR? Any implementation gotcha-s (i.e. other formats etc.)?

Thanks,
Kristen

kgilden:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected
- [ ] ~~Documentation is updated as necessary~~ I don't think there is anything to update in documentation.

Previously the XLSX writer would only add code names to worksheets, if the spreadsheet had any macros. However, sometimes code names are useful on their own. Closes #1357.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## [Unreleased]
 
+- Xlsx writer respects code names regardless of whether the spreadsheet contains macros [#1358](https://github.com/PHPOffice/PhpSpreadsheet/pull/1358)
 - Conditionals - Extend Support for (NOT)CONTAINSBLANKS [#1278](https://github.com/PHPOffice/PhpSpreadsheet/pull/1278)
 - Handle Error in Formula Processing Better for Xls [#1267](https://github.com/PHPOffice/PhpSpreadsheet/pull/1267)
 - Handle ConditionalStyle NumberFormat When Reading Xlsx File [#1296](https://github.com/PHPOffice/PhpSpreadsheet/pull/1296)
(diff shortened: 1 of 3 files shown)


kgilden:
Closing, see #1357 for explanation.

PR #1362, opened 2020-02-16
PHPDoc: use @return $this for fluent methods

caugner:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

* Some `@return` annotations used wrong casing (e.g. `@return CSV` instead of `@return Csv`)
* Most classes can be extended and therefore the annotation `@return static` would be more appropriate to avoid inadvertent type narrowing.
* The `@return $this` annotation is more explicit than `@return static`.


caugner:
@PowerKiKi Any chance to get this merged? This fixes errors reported by [Psalm](https://psalm.dev/) like this:

```
ERROR: InvalidClass - src/.../ExcelExport.php:135:17 - Class or interface PhpOffice\PhpSpreadsheet\Writer\CSV has wrong casing
                $writer->setDelimiter(',')
```

Issue #1364, opened 2020-02-17
Removing last row incorrect behavior


TimGa:
This is:

```
- [X] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
```
Before
+-----+                         After
|A1|B1|                         +-----+
+-----+                         |A1|B1|
|A2|B2| -> Removing last row -> +-----+
+-----+                         |A2|B2|
|A3|B3|                         +-----+
+-----+

After removing last 3rd row from the table
rest rows (1st and,2nd) should remain
```

### What is the current behavior?
```
Before
+-----+
|A1|B1|                         After
+-----+                         +-----+
|A2|B2| -> Removing last row -> |A1|B1|
+-----+                         +-----+
|A3|B3|
+-----+

But in fact when removing last 3rd row,
the 2nd row also disappears
```

### What are the steps to reproduce?
```php
<?php

require_once 'vendor/autoload.php';

use PhpOffice\PhpSpreadsheet\Spreadsheet;

$arrayData = [
    ['A1'],
    ['A2'],
    ['A3'],
];

$spreadsheet = new Spreadsheet();
$worksheet = $spreadsheet->getActiveSheet();
$worksheet->fromArray($arrayData);

$worksheet->removeRow(3, 1);

echo $worksheet->getCellByColumnAndRow(1, 1)->getValue() . PHP_EOL; // expect 'A1'
echo $worksheet->getCellByColumnAndRow(1, 2)->getValue() . PHP_EOL; // expect 'A2'
```
expected result is:
```
A1
A2
```
actual result is:
```
A1
```

### Which versions of PhpSpreadsheet and PHP are affected?
PhpSpreadsheet v1.10.1
PHP v7.4.2

TimGa:
#1365

oleibman:
Fixed by PR #1365.

PR #1365, opened 2020-02-17
Fix issue #1364: Removing last row incorrect behavior

TimGa:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

This PR fixes [#1364](https://github.com/PHPOffice/PhpSpreadsheet/issues/1364) Removing last row incorrect behavior

Description:
`$highestRow = $this->getHighestDataRow();` was calculated after `$this->getCellCollection()->removeRow($pRow + $r);` - this is the root reason for incorrect rows removal because removing last row will change '$this->getHighestDataRow()' value, but removing row from the middle will not change it. So, removing last row causes incorrect `$highestRow` value that is used for wiping out empty rows from the bottom of the table:
```php
for ($r = 0; $r < $pNumRows; ++$r) {
    $this->getCellCollection()->removeRow($highestRow);
    --$highestRow;
}
```
To prevent this incorrect behavior I've moved highest row calculation before row removal.
But this still doesn't solve another problem when trying remove non existing rows: in this case the code above will remove `$pNumRows` rows from below of the table, e.g. if `$highestRow=4` and `$pNumRows=6`, than rows 4, 3, 2, 1, 0, -1 will be deleted. Obviously, this is not good, that is why I've added `$removedRowsCounter` to fix this issue.
And finally, moved Exception to early if statement to get away from unnecessary 'if-else'.


TimGa:
Hi, @PowerKiKi 
Could you please review this PR?

minhtranahri:
Hi @TimGa , i think just remove this code is enough
```
for ($r = 0; $r < $pNumRows; ++$r) {
    $this->getCellCollection()->removeRow($highestRow);
    --$highestRow;
}
```

TimGa:
Hi @minhtranahri , you are right, removing this part of code will solve the issue, but somebody wrote it for some reason, that is why I decided to keep it. This part of code could be used for deleting cache stuff or some other implicit actions i don't know. So I decided to fix this issue without removing existing lines of code.

minhtranahri:
Hi @TimGa, thanks for clarifying and being careful. Very appreciate!

PowerKiKi:
Thank you for the well crafted PR and your patience.

PR #1389, opened 2020-02-21
Keep string as string, without converting it to a number [fixes #1387]

### LINKED ISSUE #1387: dataTypeForValue() casting string ("1.100") to float (1.1)
**This is:** a bug report;

### Description

I am using `$worksheet->fromArray()` method, that calls the `DefaultValueBinder::dataTypeForValue()` with my datas. Some of have values like `"1.100"` (as string, a street number, for instance) or `"1,100"` (as string too), but it is replaced by a float `1,1` (the "dot" or "comma" will depends of system locale).

Cell should stores a string `"1.100"`.

Cell is storing a float `1,1`. Which is an invalid street number information (that should be exactly `"1.100"`).

```php
<?php

$spreadsheet = new Spreadsheet;

$sheet->fromArray([
    [ '1.100' ]
]);

$writer->save('issue-1387.xlsx');
```

[issue-1387.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/4236980/issue-1387.xlsx)

### Debugging

I have debugged the `DefaultValueBinder::bindValue()` method, and checked the output from `$value` parameter and `static::dataTypeForValue($value)` return. It is what I received:

image

It is: `dataTypeForValue()` will return `"n"` (_number_) for `"1.100"` instead of `"s"` (_string_).

How the [`preg_match()` line](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Cell/DefaultValueBinder.php#L64) will handle this value (`"1.100"`), so I think that it should be improved to keep `"1.100"` as string, or create an alternative method to force string when using `$worksheet->fromArray()`.

For instance, we have the [`TextElement` class](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/RichText/TextElement.php), but it actually is not supported at this context.

Anyway, I think that `"1.100"` should not be treated as number, except if it `is_int()` or `is_float()`. Or if it have any `"0"` on right-side.

### Workaround

The current possible workaround is very ugly and will modify the cell data: `"1.100 "` (note the space after the number).

Other option is forcing via `$worksheet->setCellValueExplicit()`, but it is not viable when using `$worksheet->fromArray()`.

- PhpSpreadsheet 1.10.1
- PHP 7.3.14

rentalhost:
**This is:** a bugfix.

**Checklist:**

- [x] Changes are covered by all unit tests;
- [x] Code style is respected (_read the justification_);
- [x] Commit message explains **why** the change is made;
- [x] CHANGELOG.md contains a short summary of the change;
- [x] Documentation is updated as necessary (_not necessary_).
- [x] No breakchanges;

Fixes #1387. 

The string `"1.100"` was being converted to number `1.1`, which break a "street number" information, for instance, if put in its own column (_as my real case_). If user wants that the column receives a real number it must cast `(int)`. 

Currently there not a good way to keep this value "as is" by using `$worksheet->fromArray()` method. But this commit will allow the numeric version works by casting `(int)` if user prefer the previous mode, but it need be done implicity.

I don't believe that it is a breakchanging, but a bugfix due that the string is modified from original content to be represented as a number where the original value was a real string.

### Commits explanation

The fix itself is done on the [**first commit 849f36f**](https://github.com/PHPOffice/PhpSpreadsheet/commit/849f36f901918f02d302c0ba0ef46e9b7cccfe0f). If only it cares, merge only it. 

**I have additional notes about this commit:**

* It consider as _string-type_ (`"s"`) any string lower than `PHP_INT_MAX` that matches with `/\.\d*0$/`. It is: strings like `"1.0"` should be keeped as string and not be converted to number automatically. It is need because Excel will strip leading zeros visually for numbers. So even for strings it will be displayed just like `1`, which could breaks the real information that was passed as string.
* I had to modify the [`Worksheet::writeCell()`](https://github.com/rentalhost/PhpSpreadsheet/blob/38bc2a0ca49a396b26e40d20163f4873504a8f66/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php#L1074) method to writes the `c[t]` tag attribute even for _numeric-type_ (`"n"`). It was needed because the main change will concludes that `1.0` is string during the reading too, so now having `<c t="n">` will garantees that `1.0` is in fact a number during the reading process;
* I had added [9 new tests](https://github.com/rentalhost/PhpSpreadsheet/blob/38bc2a0ca49a396b26e40d20163f4873504a8f66/tests/data/Cell/DefaultValueBinder.php#L76-L111) to `DefaultValueBinder`, so we can make sure that this commit will not break nothing and will fix the issue itself;

In case of the user really needs converts `1.0` into number, it should be casted by `(int)`. In that way, the cell type will be _number_ and not _string_.

---

Ok, now let me talk about the [**second commit c532572**](https://github.com/PHPOffice/PhpSpreadsheet/commit/c532572dc1a38a2874b8d942a210d7c95f9711be). I did a bunch of optimizations and improvements to code reading, but I believe that I do not broken the _code-style rules_, but too will not add none real performance advantage (I am considering no gain during tests, a real world case I belive that it could be valid).

**It was what I did:**

* I had removed the `elseif()` for where it is not need. When a `if() return;` occur, `elseif()` is redundant and could be replaced safety by just an `if()`;
* [DefaultValueBinder.php#L72](https://github.com/rentalhost/PhpSpreadsheet/blob/c532572dc1a38a2874b8d942a210d7c95f9711be/src/PhpSpreadsheet/Cell/DefaultValueBinder.php#L72): originally it check if `is_string()` before check if first character is `=` (two conditionals). I had dropped the `is_string()` once that previous code will discard non-string types (eg. `is_null()`, `is_float()`...). It improves performance in this point about _11.8%_;
* [DefaultValueBinder.php#L76-L83](https://github.com/rentalhost/PhpSpreadsheet/blob/c532572dc1a38a2874b8d942a210d7c95f9711be/src/PhpSpreadsheet/Cell/DefaultValueBinder.php#L76-L83): originally the new commit comes inside of `preg_match()` step, but it is not need. So I have moved to before it due to use less resources, and the new `preg_match()` will consumes less than before;
* [DefaultValueBinder.php#L85](https://github.com/rentalhost/PhpSpreadsheet/blob/c532572dc1a38a2874b8d942a210d7c95f9711be/src/PhpSpreadsheet/Cell/DefaultValueBinder.php#L85): the previous `preg_match()` (that was keeped) have a lot of needless escapes, so I dropped it and I had normalized the order of `-+` signals (where when it is applied, `-` will be more common than `+`);
* [DefaultValueBinder.php#L85](https://github.com/rentalhost/PhpSpreadsheet/blob/c532572dc1a38a2874b8d942a210d7c95f9711be/src/PhpSpreadsheet/Cell/DefaultValueBinder.php#L85): previously, when a possible number is matched by the `preg_match()` a procedure do a `ltrim(..., "+-")`. So basically it could return a string-type for `"01"` string, for instance. But it could be done with a negative lookahead on this same `preg_match()` with the same effect;
* [DefaultValueBinder.php#L89](https://github.com/rentalhost/PhpSpreadsheet/blob/c532572dc1a38a2874b8d942a210d7c95f9711be/src/PhpSpreadsheet/Cell/DefaultValueBinder.php#L89): previous there are a check if string could be a error code. It was keeped, but the way of how it is done have changed. I had created a new method `DataType::isErrorCode()` that will check it. No performance advantages, just moving the responsability to the `DataType` class.


src/PhpSpreadsheet/Cell/DefaultValueBinder.php
```diff
@@ -51,30 +51,43 @@ public static function dataTypeForValue($pValue)
...
-                return DataType::TYPE_STRING;
-            } elseif ((strpos($pValue, '.') === false) && ($pValue > PHP_INT_MAX)) {
+        }
+
+        if ($pValue > PHP_INT_MAX) {
+            if (strpos($pValue, '.') === false) {
                 return DataType::TYPE_STRING;
             }
+        }
+        else if (preg_match('/\.\d*0$/', $pValue)) {
```
MichaelPFrey: unnecessary new line.
the `}` should be on the same line as the `else if (`
(see also https://travis-ci.org/PHPOffice/PhpSpreadsheet/jobs/653614524)

rentalhost: Fixed.

rentalhost:
@MichaelPFrey could you check if now it is fine and merge to master?

There are some ETA to next patch version?

MichaelPFrey:
> @MichaelPFrey could you check if now it is fine and merge to master?

I am **not** a project member/contributor, just a random person. I just looked into the project for personal education, saw the failed automated test (seeing where/how test fails is the quickest way to learn what they do) and gave you a hint how to pass the test.

rentalhost:
Oh, sorry. Anyway, thanks for help me here. ^^

@PowerKiKi @MarkBaker could you check that?

rentalhost:
@PowerKiKi thanks for reply.

The case you cited did not exist, until then, in the tests, so I declared it by myself. But I imagine that this change has indeed occurred, and it is precisely the problem that is being corrected.

Note: when you input `string ("123.000")` we must have an output like `123.000` on XLS, because it could be an address number, like my case. And currently is impossible to get this output to file via `fromArray()`. Actually it is converted to `123.0`, which breaks data and make it unrecoverable from XLS. It seems a bug to me, you don't think? The write-and-read process should return the same wrote value during reading, I guess.

In case of user wants really to keep as number, he just should cast it `(float) "123.000"` and it will be done.

So basically what I did was identify the input type and set it as-is to output. And during reading it could be returned as original source was.

Please, I ask you to try to reconsider or try to imagine a real situation where this could be an even bigger problem than what is being dealt with here.

Thanks! 😃

PowerKiKi:
The behavior that you describe as a bug, is actually an intentional feature of `DefaultValueBinder` as is proven in https://github.com/PHPOffice/PhpSpreadsheet/blob/b07cd2028dd2454cbcfeff9a70d41761bc77f054/tests/data/Cell/DefaultValueBinder.php#L36-L39

> And currently is impossible to get this output

This is incorrect. See the following doc to learn more on how to implement the behavior you want: https://phpspreadsheet.readthedocs.io/en/latest/topics/accessing-cells/#using-value-binders-to-facilitate-data-entry

rentalhost:
But this code was not modified and was keep as-is:

rentalhost:
But this code was not modified and was keep as-is:

https://github.com/PHPOffice/PhpSpreadsheet/blob/a732ca0331917d02edf5c7a194164715931256b4/tests/data/Cell/DefaultValueBinder.php#L36-L39

Basically, numeric strings are casted to number, for instance, `string ("123")` will keep casts to `int (123)`. The unique difference is when your string contains multiple zeroes on the right like `string ("123.00")`, so we mantain the `.00` as-is without converts to number.

Your solution using bindings was basically what I did on my side to avoid this problem, but it requires two loopings because I am using `fromArray()` (that is the problem itself), while the `setCellValue()` works as expected due to bind support.

If you disagree with the application of the data in this way, you can consider closing the PR, as I have no idea how to resolve it in any other way.

Anyway, thanks for the support and attention!

PowerKiKi:
Closing since this would go against the intention of DefaultValueBinder and alternative, non-hack solutions exist.

Issue #1375, opened 2020-02-19
PDF-Generation don't respect condtional formatting


Collie-IT:
This is:

```
- [x ] a bug report
- [x ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
PDF generation should respect the conditonal formating rules. The PDF should look like following Picture.
![2020-02-19_18h20_21](https://user-images.githubusercontent.com/40590185/74857901-b0f5c600-5344-11ea-98e8-434b595b94c0.png)

The Examplefile is:

[ExampleFormatting.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/4226371/ExampleFormatting.xlsx)

### What is the current behavior?
The PDF don't has the condtional bold formatting. Like shown in the following image:

![2020-02-19_18h25_21](https://user-images.githubusercontent.com/40590185/74858199-36797600-5345-11ea-8469-09932bbea712.png)

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$reader =  \PhpOffice\PhpSpreadsheet\IOFactory::createReader("Xlsx");
$spreadsheet = $reader->load(<PATH TO XLSX>);
$writer = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($spreadsheet, 'Tcpdf');
$writer->save(<TARGET PATH>);
// add code that show the issue here...
```

### Which versions of PhpSpreadsheet and PHP are affected?
current PhpSpreadsheet php 7.3

oleibman:
Reopening. This will be addressed by PR #4412, and will be closed again once that is complete.

PR #1393, opened 2020-02-23
Is condtional style active

Collie-IT:
This is:

```
- [ ] a bugfix
- [x] a new feature

```

Checklist:

- [x] Code style is respected
- [x] Documentation is updated as necessary -not needed

PDF (tcpdf) printing does not respect condtional formatting  #1375.  It is necessary to have a function to check if is there an active conditional formatting and  then to set the cell style appropriatly.

### Limitations
All CELLIS condtion type with there operater types/functions are covered. The text operators (STARTSWITH, ENDSWITH, CONTAINS AND NOT CONTAINS) aren't covered because they are currently not implemented in Worksheet.php writeCFRule.


Collie-IT:
Dear @PowerKiKi, I added your feedback

> $multuseCol and $multuseRow are not explained, and their name are not self-describing

- I changed the names

>     don't use binary file for test, and instead create spreadsheet in memory from scratch

- I added a new test file 
- I added a in memory method because of that excel can do some strange things in the conditional formatting I don't delete the excel test case with a real excel.

Excel manages to have conditions that looks like for e.g. =$E$5:$E$14 NOT Between -$D$5-0.4 And $D$5+0.4 That means the condition will return -$D$5-0.4 and $D$5+0.4 everytime, but this must be adjusted in case of cell $E$6 to -$D$6-0.4 And $D$6+0.4  

> the new method should be useful somewhere in PhpSpreadsheet, otherwise there is no value to add it. The only place I can think of would be in Writer/Html, so Html and PDFs would support conditional styles.

- Indeed there should be an integration to the writer. I would prefer the html writter, too.  Can you suggest a good place? 
- For Tcpdf I have already implemented the code

> to be really useful, we should also have a new Worksheet::getStyle

- Can you tell me how this should look like? I don't understand what I should do in this case.

>  why did you choose bccomp, instead of standard operators ?

- bccomp is used for floats comparison with a precision in number of decimalplaces. (Excel rounds to displayed precision before comparision)

> Do not disable the cache by default, it would be a performance hit. Instead do nothing at all, and let end-users do that manually if they really requires it

- I have excluded the disable cache by default
- But there is an issue at the second condtion. Every time the value of the first condition will be given back from the cache, e.g. condtion1 = 17.5 and condtion2 = -17.5 the return is everytime 17.5. 

I strongly suggest to use   $calcer->disableCalculationCache(); or find the reason for wrong cache return.

src/PhpSpreadsheet/Style/Conditional.php
```diff
@@ -271,4 +274,148 @@ public function __clone()
...
+                    $return = (!(bccomp($calcVal, $coords0Val, $precision) <= 0 || bccomp($calcVal, $coords1Val, $precision) <= 0));
+                } else {
+                    $return = (!(bccomp($calcVal, $coords1Val, $precision) <= 0 || bccomp($calcVal, $coords0Val, $precision) <= 0));
+                }
+
+                break;
+            case '':
+                break;
+            default:
+                throw new \Exception('Unknown ' . $this->getOperatorType() . ' during active check of conditional formatting.');
```
PowerKiKi, line 355:
```suggestion
                throw new \Exception('Unknown ' . $this->getOperatorType() . ' during check of active conditional formatting.');
```

src/PhpSpreadsheet/Style/Conditional.php
```diff
@@ -271,4 +274,148 @@ public function __clone()
...
+     * @param string $cellName
+     * @param int $precision
+     * @param int $updateColumnBy
+     * @param int $updateRowBy
+     *
+     * @throws Exception
+     *
+     * @return bool
+     */
+    public function isActive(Calculation $calcer, Cell $cell, int $precision = 8, int $updateColumnBy = 0, int $updateRowBy = 0)
```
PowerKiKi, line 299:
```suggestion
    public function isActive(Calculation $calcer, Cell $cell, int $precision = 8, int $updateColumnBy = 0, int $updateRowBy = 0): bool
```

oleibman:
Conditional formatting for html (and therefore pdf) was added by PR #4412. Removing stale label, leaving issue closed.

PR #1398, opened 2020-02-28
Fix return type docblock for the Cells::get()

artemmolotov:
This is:

```
- [+] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

PHPStorm inspections and PHPStan display an incorrect hint or error if the special type `null` is not specified.


PowerKiKi:
Thanks ! merged in  8eaceb0

PR #1409, opened 2020-03-06
Added support for the WEBSERVICE function

paulkned:
This is:

```
- [ ] a bugfix
- [ x ] a new feature
```

Checklist:

- [ x ]  CHANGELOG.md contains a short summary of the change

The WEBSERVICE function was not supported, and now it is. The only thing I'm unsure of are the unit tests. Obviously this function needs to contact an external website. I chose a generic API which seems like it will stay in the air for a while. However, there is no way of knowing this for sure, neither is there any way of knowing if the API will keep returning the same results. This will break the unit tests without any code changes. But not including this test will make sure the function is not properly tested ... dilemma's, dilemma's.


PowerKiKi:
What about using something like guzzle ? We can then assume guzzle is tested and working as expected and we may be able to mock it for our own tests ?

paulkned:
@PowerKiKi Good point. I didn't want to add packages willy-nilly but with your blessing I added guzzle. This made mocking and testing a bit robuster.

src/PhpSpreadsheet/Calculation/Web.php
```diff
@@ -0,0 +1,69 @@
...
+{
+    protected static $client;
+
+    /**
+     * Override the default webclient.
+     * Used for test purposes.
+     *
+     * @param Client $client
+     */
+    public static function setClient(Client $client)
```
PowerKiKi, line 19: This should live in `\PhpOffice\PhpSpreadsheet\Settings`, like the similar `setCache/getCache`, so settings are centralized and could _maybe_ later be transformed into a PSR-11.

src/PhpSpreadsheet/Calculation/Web.php
```diff
@@ -0,0 +1,69 @@
...
+     * Excel Function:
+     *        Webservice(url)
+     *
+     * @category Web Functions
+     *
+     * @param string $url
+     *
+     * @return string the output resulting from a call to the webservice
+     */
+    public static function WEBSERVICE($url)
```
PowerKiKi, line 38: Use proper typing for all new code

```suggestion
    public static function WEBSERVICE(string $url)
```

tests/PhpSpreadsheetTests/Calculation/Functions/Web/WebServiceTest.php
```diff
@@ -0,0 +1,56 @@
...
+use GuzzleHttp\Psr7\Response;
+use PhpOffice\PhpSpreadsheet\Calculation\Functions;
+use PhpOffice\PhpSpreadsheet\Calculation\Web;
+use PHPUnit\Framework\TestCase;
+
+class WebServiceTest extends TestCase
+{
+    protected static $client;
+
+    public static function setUpBeforeClass()
```
PowerKiKi, line 20:
```suggestion
    public static function setUpBeforeClass(): void
```

tests/PhpSpreadsheetTests/Calculation/Functions/Web/WebServiceTest.php
```diff
@@ -0,0 +1,56 @@
...
+            new ConnectException('This is a 404 error', new Request('GET', 'test')),
+            new Response('200', [], str_repeat('a', 40000)),
+            new Response('200', [], 'This is a test'),
+        ]);
+
+        $handlerStack = HandlerStack::create($mock);
+        self::$client = new Client(['handler' => $handlerStack]);
+    }
+
+    public function setUp()
```
PowerKiKi, line 34:
```suggestion
    public function setUp(): void
```

tests/PhpSpreadsheetTests/Calculation/Functions/Web/WebServiceTest.php
```diff
@@ -0,0 +1,56 @@
...
+        Functions::setCompatibilityMode(Functions::COMPATIBILITY_EXCEL);
+    }
+
+    /**
+     * @dataProvider providerWEBSERVICE
+     *
+     * @param mixed $expectedResult
+     * @param string $url
+     */
+    public function testWEBSERVICE($expectedResult, $url)
```
PowerKiKi, line 45:
```suggestion
    public function testWEBSERVICE(string $expectedResult, string $url): void
```

tests/PhpSpreadsheetTests/Calculation/Functions/Web/WebServiceTest.php
```diff
@@ -0,0 +1,56 @@
...
+    {
+        Web::setClient(self::$client);
+        $result = Web::WEBSERVICE($url);
+        $this->assertEquals($expectedResult, $result);
+    }
+
+    public function providerWEBSERVICE()
```
PowerKiKi, line 52:
```suggestion
    public function providerWEBSERVICE(): array
```

patrickbrouwers:
@PowerKiKi @paulkned @MarkBaker  is Guzzle 7 a hard requirement for this to work or could we loosen up the requirement to `^6.3.1|^7.0`?  

Installing in a clean Laravel 7.0 install currently doesn't allow 1.14 to be installed, which I'm afraid will cause a lot of confusing among the Laravel Excel users who want to upgrade to use the backwards compatible fix for the empty enclosures which was fixed in 1.14.

mfn:
See also https://github.com/PHPOffice/PhpSpreadsheet/issues/1562

I did see the new release but wondering why it didn't upgrade => because my project uses Guzzle ^6

PR #1424, opened 2020-03-24
Limit composer dist to src/, bin/ and a few informative files

spawnia:
While the docs and samples are great, i do not need them in my vendor directory. The same is true for the many configuration files.

A few select files such as the readme, changelog and contributing information were deliberately left in. We can remove those too if you think that is better.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,14 @@
-/tests export-ignore
-README.md export-ignore
-*.min.js binary
 /.github export-ignore
+/docs export-ignore
+/samples export-ignore
+/tests export-ignore
+/.gitattributes export-ignore
+/.gitignore export-ignore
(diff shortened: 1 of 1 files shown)


chrisidakwo:
I'd always thought it was too much having samples and some other "unnecessary" files in a vendor, however they would usually come really handy for someone using the package for the first couple of times. I speak from experience. I can't count the number of times I looked at the sample files to understand how to go about certain things.

To that end, I'd suggest why not have those samples as a standalone package? 

Too much? I can try to help with that

spawnia:
@chrisidakwo

    composer install --prefer-source

chrisidakwo:
Mad! 😂 Actually didn't even think of that! Damn!

PR #1437, opened 2020-04-05
fix: Don't remove rows twice

drola:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

While migrating from PHPExcel to PHPSpreadsheet, I tracked down a bug to removeRow() method. The method mistakenly removes the rows twice, resulting in too many rows being removed. This PR fixes the issue


drola:
This issue has been fixed by #1364. Thank you for the review. Closing, because this fix is not needed anymore.

PR #1442, opened 2020-04-13
Add support for IFS() logical function

Takitaf:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

PHPSpreadsheet missing Excel IFS logical function support


MarkBaker:
Thank you for this submission; I'l take a look at it this weekend, as it would be good to support the IFS() function

Takitaf:
@MarkBaker Thanks for a response. 

There is one detail in my implementation that probably need to be changed. In theory I want to use existing functions if I can and also want to stop executing on a first condition that returned true. It works that way, but I use here the `Functions::DUMMY()` for a "false value" placeholder. As far as I can see, although by name it seems to be perfectly fine, it is rather used to mark not-implemented features yet.

Additionally, there is a small chance, that user's "true value" would be exactly the same as `Functions::DUMMY()` value - and in that case it will not behave correctly.

I could calculate IFS in reverse order - calculate last condition, get previous condition and use last condition value as "false value" of previous condition, and repeat that until I reach the very first condition. It, however, makes this feature calculate always all conditions in each IFS statement. 

Maybe good enough solution is just introduce dummy value locally, that is even harder to collide with in real world (php_spreadsheet prefix + hash?)

src/PhpSpreadsheet/Calculation/Logical.php
```diff
@@ -372,4 +372,40 @@ public static function IFNA($testValue = '', $napart = '')
...
+        if (count($arguments) % 2 != 0) {
+            return Functions::NA();
+        }
+
+        for ($i = 0; $i < count($arguments); $i += 2) {
+            $testValue = ($arguments[$i] === null) ? '' : Functions::flattenSingleValue($arguments[$i]);
+            $returnIfTrue = ($arguments[$i + 1] === null) ? '' : Functions::flattenSingleValue($arguments[$i + 1]);
+            $result = self::statementIf($testValue, $returnIfTrue, Functions::DUMMY());
+
+            if ($result !== Functions::DUMMY()) {
```
MarkBaker, line 404: Yes, I understand your concern that there is a possibility that the returned result may be a valid case of the string `'#Not Yet Implemented'`; unlikely, but not impossible.
I might be inclined to create something purely PHP detectable, that cannot exist in a spreadsheet, such as an exception. Maybe create a new EvaluateIfFalseException which can be returned as `$result` rather than thrown; and directly type tested rather than simply value tested.

Takitaf: That's a good idea. Can I use just simple instance of generic Exception in that case: `$falseValueException = new Exception();`? Not sure if we need special Exception for that.

MarkBaker: Yes you could use a standard Exception, I was simply thinking about readabilityand providing a named exception where the name provided contextual meaning

PR #1444, opened 2020-04-14
S3 streaming support

andonovn:
Not sure if this is considered a bug fix or a new feature.

Checklist:

- [X] Code style is respected

Currently, it's impossible to stream the file to s3. The following is the error when someone attempts to do so:

`ZipArchive::close()`
`/var/www/application/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Writer/Xlsx.php:409`

The solution provided in this PR is very naive, and not elegant at all, but it works.
Another option would be to extract `strtolower($pFilename) == 'php://output' || strtolower($pFilename) == 'php://stdout'` to a protected method, which could be overriden by the consumers.
There are more complex options as well, fe creating different adapters, but unfortunately I don't really have time to work on them.

But the main goal of this PR is to open a discussion around this. I am curious to know if you are willing to add S3 streaming support at all.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Xlsx.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx.php
@@ -178,7 +178,7 @@ class Xlsx extends BaseWriter
 
             // If $pFilename is php://output or php://stdout, make it a temporary file...
             $originalFilename = $pFilename;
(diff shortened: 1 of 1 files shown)


PowerKiKi:
We have #1292 to add support for writing to resources in a more general way. I believe that would make this PR obsolete.

PR #1448, opened 2020-04-18
Chart Samples Fix

cwolcott:
Update chart parameter from 0 to DataSeries::EMPTY_AS_GAP

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

All chart examples passed the displayBlanksAs parameter as 0 instead of 'gap'. I added a constant EMPTY_AS_GAP to the DataSeries and then change all chart samples to use this new constant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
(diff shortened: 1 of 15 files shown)


oleibman:
Should samples/templates/chartSpreadsheet.php also be updated to use 'gap' rather than '0'? (Used in 33_Chart_create_bar and 34_Chart_update)

Issue #1457, opened 2020-05-01
Cell Coordinate splitRange doesn't support space delimeter


mcouillard:
This is: a new feature request without affecting existing functionality.

```
- [ ] a bug report
- [X] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
The range argument accepts a list of coordinates by comma or space

### What is the current behavior?
It incorrectly parses the list into an array of invalid coordinates

### What are the steps to reproduce?
A client of ours created an Excel spreadsheet with the following Protection rules:
```
<protectedRanges>
 <protectedRange sqref="C14:O15 C161:O1081 C16:H160 J16:O160" name="SheetName1"/>
 <protectedRange sqref="I16:I160" name="SheetName2"/>
</protectedRanges>
```
I don't know what version of Excel was used - likely the latest version.

### Which versions of PhpSpreadsheet and PHP are affected?
1.10 onward, perhaps earlier

mcouillard:
Ping to keepalive; good discussion in the PR and Mark may be able to find the right solution

oleibman:
Removing stale label and reopening issue. This is similar to how Excel treats ranges in Conditional Formulas (see PR #4042) and Data Validation (PR #4240), inverting the meaning of commas and spaces from how it treats multiple ranges in formulas. It should be possible for PhpSpreadsheet to handle this inversion for Protected Ranges as well.

PR #1458, opened 2020-05-01
Cell Coordinate splitRange doesn't support space delimeter

mcouillard:
This is: new feature to support spaced coordinates instead of just comma delimited lists. Specific to the splitRange() method.
Issue #1457 

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [X] Code style is respected

Excel can save protection ranges as a spaced list instead of just a comma list.


MarkBaker:
You're right that a space can appear in a range definition; but it isn't simply an alternative for a comma.
Both space and comma are operators in a range definition:
 - A comma is the Union operator (combines two references into a single reference)
 - A space is the Intersection operator (returns a reference to the intersection of two ranges)
So I need to take a much more detailed look at how well PHPSpreadsheet handles this difference when it is used.

mcouillard:
That makes more sense! My solution didn't feel optimal and I couldn't find the right doc from Microsoft or about OpenXML to determine what these spaces here actually meant.

I appreciate your help.

MarkBaker:
He, yes! It's pretty obscure: simple example is to fill cells A1 to C3 with the values 1 to 9, then to have a couple of formulae: `=SUM(A1:B2,B2:C3)` and `=SUM(A1:B2 B2:C3)` and look at the different results.
I'll try putting some tests together tonight to see what the effects will be.

oleibman:
Superseded by PR #4580. Removing stale label, leaving closed.

PR #1460, opened 2020-05-02
README - add link to migration path to show people easier way

TomasVotruba:
I've notice the README contains what has happened with PHPExcel, but actually doesn't tell user what to do about it.

I think most people would not assume there exists some magic powerful tool that will handle most of the migration for them. They might start to rewrite it manually, which is terrible waste of time.

Saying that, there should be at least small link to already existing documentation


TomasVotruba:
Ping

TomasVotruba:
Thanks :+1:

Issue #1246, opened 2019-11-18
Using the DAYS function produces NAME? an error after open XLSX spreadsheet in edit mode


marinbale2377:
This is:

```
- [X] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
The DAYS($endDate, $startDate) should calculate the number of days between an starting date and an endate.

### What is the current behavior?
The function works correctly after XLSX opening in view only mode, but after enabling edit mode every CELL where is used returns an NAME? error indicating that invalid characters where introduced in the formula.

If you manually edit the formula (F2) and click enter, the correct value is recalculated for the cell.

As a workaround, you can substitute the formula by a simple subtract of  the start_date from the end_date, but using the formula will be great, as it sort of self-document the purpose of the operation.

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Enable AdvancedValueBinder and RETURNDATE_EXCEL
\PhpOffice\PhpSpreadsheet\Cell\Cell::setValueBinder( new \PhpOffice\PhpSpreadsheet\Cell\AdvancedValueBinder() );        
\PhpOffice\PhpSpreadsheet\Calculation\Functions::setReturnDateType(\PhpOffice\PhpSpreadsheet\Calculation\Functions::RETURNDATE_EXCEL);

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

// add code that show the issue here...
// $credit->plazo_meses = 120;
for ($x = 0; $x < $credit->plazo_meses; $x++) {
// Interes: =IF(ISBLANK(B12),($B$7/12)*K11,(((DAYS(B12,B11))/365)*$B$7)*K11)
    $fInteres = '=IF(ISBLANK(B'.(12+$x).'),($B$7/12)*K'.(11+$x).',(((DAYS(B'.(12+$x).',B'.(11+$x).'))/365)*$B$7)*K'.(11+$x).')';
    $sheet->setCellValue('I'.(12+$x), $fInteres);
    $sheet->getStyle('I'.(12+$x))
[...]

PR #1461, opened 2020-05-03
Save Excel 2010+ Functions Properly

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

For functions introduced in Excel 2010 and beyond, Excel saves them
in formulas with the xlfn_ prefix. PhpSpreadsheet does not do this;
as a result, when a spreadsheet so created is opened, the cells
which use the new functions display a #NAME? error.
This the cause of bug report 1246:
https://github.com/PHPOffice/PhpSpreadsheet/issues/1246
This change corrects that problem when the Xlsx writer encounters
a 2010+ formula for a cell or a conditional style. A new class
Writer/Xlsx/Xlfn, with 2 static methods,
is introduced to facilitate this change.

As part of the testing for this, I found some additional problems.
When an unknown function name is used, Excel generates a #NAME? error.
However, when an unknown function is used in PhpSpreadsheet:
  - if there are no parameters, it returns #VALUE!, which is wrong
  - if there are parameters, it throws an exception, which is horrible
Both of these situations will now return #NAME?
Tests have been added for these situations.

The MODE (and MODE.SNGL) function is not quite in alignment with Excel.
MODE(3, 3, 4, 4) returns 3 in both Excel and PhpSpreadsheet.
However, MODE(4, 3, 3, 4) returns 4 in Excel, but 3 in PhpSpreadsheet.
Both situations will now match Excel's result.
Also, Excel allows its parameters for MODE to be an array,
but PhpSpreadsheet did not; it now will.
There had not been any tests for MODE. Now there are.

The SHEET and SHEETS functions were introduced in Excel 2013,
but were not introduced in PhpSpreadsheet. They are now introduced
as DUMMY functions so that they can be parsed appropriately. 

Finally, in common with the "rate" changes for which I am
creating a pull request at the same time as this one:
samples/Basic/13_CalculationCyclicFormulae
PhpUnit started reporting an error like "too much regression".
The test deals with an infinite cyclic formula, and allowed
the calculation engine to run for 100 cycles. The actual number of cycles
seems irrelevant for the purpose of this test. I changed it to 15,
and PhpUnit no longer complains.


PowerKiKi:
Merged in 4f6d4af39626d5c1cef76f446ee743896d7d6982, thanks !

tyteen4a03:
This fixes `ISOWEEKNUM()` but doesn't fix `DAYS()`. Is `DAYS()` related to this change?

oleibman:
Yes, DAYS is covered by this change. The following code produces the expected result for me when run against a freshly downloaded copy of PhpSpreadsheet:
        $sheet->setCellValue('A1', '=DAYS("2017-01-01","2016-01-01")');
        $writer->save('outfile.xlsx');
What kind of problem are you seeing?

PR #1464, opened 2020-05-11
Improve HTML Writer

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

There are a number of situations where HTML write was producing
HTML which could not be validated. These include:
  - inconsistent use of backslash terminating META, IMG, and COL tags
  - @page style tags in body rather than header. Aside from being
    non-standard, HTML Reader treats those as spreadsheet data.
  - `<div style="page-break-before:always" />`, a construct which is
    usually better handled through css anyhow.
  - no alt tag for images (drawings and charts)

Other problems:
  - Windows file names not handled correctly for images
  - Memory drawings not handled in extendRowsForChartsAndImages
  - No handling of different values for showing gridlines
    for screen and print
  - Mpdf and Dompdf do not require the use of inline css.
    Tcpdf remains a holdout in the use of this inferior approach.
  - no need to chunk base64 encoding of embedded images
  - support for colors in number format was buggy (html tags
    run through htmlspecialchars)

Code has been refactored when practical to reduce the number of
very large functions.

Coverage is now 100% for the entire HTML Writer module,
from 75% lines and 39% methods beforehand.

All functions dealing only with charts
are bypassed for coverage because the version of Jpgraph available in
Composer is not suitable for PHP7. The code will, nevertheless,
run successfully, but with warning messages. I have confirmed that
the code is entirely covered, without warnings, when the current
version of Jpgraph is used in lieu of the one available in Composer.
I will be glad to revisit this when the Jpgraph problem is resolved.

Directory PhpSpreadsheetTests/Writer/Html was created to house
the new tests. It seemed logical to move HtmlCommentsTest to
the new directory from PhpSpreadsheetTests/Functional.

A function to generate all the HTML is useful, especially for testing,
but also in lieu of the multiple other generate* functions. I have
added and documented generateHTMLAll.

The documentation for the generate* functions (a) produces invalid html,
(b) produces html which cannot be handled correctly by HTML reader,
and (c) even if those were correct, does not actually affect
the display of the spreadsheet. The documentation has been replaced
by a valid, and more instructive, example.

The (undocumented) useEmbeddedCss property, and the functions
to test and set it are no longer needed. Rather than breaking
existing code by deleting them, I marked the functions deprecated.

This change borrows a change to LocaleFloatsTest from
pull request 1456, submitted a little over a week before this one.


oleibman:
I have resolved the conflicts. After that, Travis CI complained about tests which it had not complained about before - it appears that it changed what it considers the current directory. After fixing those, Scrutinizer complained about something which it had not complained about before. Everything is now fixed, all tests pass, no new issues. So I think this is ready to go. I shall take care to declare types for future changes.

PowerKiKi:
Thanks again :tada: 

Since now all new code MUST be typed. Existing code that you touch SHOULD be typed too, but we have to be careful to avoid breaking things. Because of the nature of this project, we rely a lot on type flexibility, so it is likely that we will never be able to introduce typing in some places. Typically all calculation things might be hard to type, but our "models", readers and writers should be typable.

Issue #1465, opened 2020-05-11
Caculation of DATEDIF result is partially wrong


albanduval:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

DATEDIF calculated result should return the same result as MS Excel

Diff in years ("Y" unit) :

- 2000-02-02 => 2001-02-01 : should return 0
- 2001-05-04 => 2020-05-03 : should return 18

Diff in months with years ignored ("YM' unit) :

- 2000-02-02 => 2001-02-01 : should return 11
- 2001-05-04 => 2020-05-03 : should return 11

### What is the current behavior?

Diff in years ("Y" unit) :

- 2000-02-02 => 2001-02-01 : returns -1
- 2001-05-04 => 2020-05-03 : return 17

Diff in months with years ignored ("YM' unit) :

- 2000-02-02 => 2001-02-01 : return -1
- 2001-05-04 => 2020-05-03 : return -1

### What are the steps to reproduce?

```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

$intervals = [
    [ 'start' => '2000-02-02', 'end' => '2001-02-01' ],
    [ 'start' => '2001-05-04', 'end' => '2020-05-03' ],
    ];

foreach($intervals as $interval) {
    echo "{$interval['start']} => {$interval['end']}\n";
    $startDate = new \DateTime($interval['start']);
    $endDate = new \DateTime($interval['end']);

    $diffY = \PhpOffice\PhpSpreadsheet\Calculation\DateTime::DATEDIF($startDate, $endDate, 'Y');
    echo "  >  $diffY years\n";

    $diffYM = \PhpOffice\PhpSpreadsheet\Calculation\DateTime::DATEDIF($startDate, $endDate, 'YM');
    echo "  >  $diffYM months\n";
}
```

### Which versions of PhpSpreadsheet and PHP are affected?

master branch since at least 2016

PR #1466, opened 2020-05-11
Calculation - DATEDIF - fix result for Y & YM units

albanduval:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

To solve the issue #1465 

Attached is a [zip file](https://github.com/PHPOffice/PhpSpreadsheet/files/4608441/test_datedif.zip) containing :
- a PHP CLI tool to run a batch of tests that exports the results into CSV files
- the CSV results for current master branch
- the CSV results for the code with fix

The results show differences in the results for both Y & YM units, correcting the detected wrong results.


MarkBaker:
Thank you for this submission. The original code was ported from the older PHPExcel library, with it's PHP 5.2 minimal version, so using DateInterval was never an option, and the code wasn't perfect. Now, with PHPSpreadsheet's higher minimum version, it is an option; so thank you for working on this update, and for resolving the issues that were in the original code.

Your work is appreciated.

PR #1475, opened 2020-05-17
Improve Coverage for CSV

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

I believe that both CSV Reader and Writer are 100% covered now.

There were some errors uncovered during development.

The reader specifically permits encodings other than UTF-8 to be used.
However, fgetcsv will not properly handle other encodings.
I tried replacing it with fgets/iconv/strgetcsv, but that could not
handle line breaks within a cell, even for UTF-8.
This is, I'm sure, a very rare use case.
I eventually handled it by using php://memory to hold the translated
file contents for non-UTF8. There were no tests for this situation,
and now there are (probably too many).

"Contiguous" read was not handle correctly. There is a file
in samples which uses it. It was designed to read a large sheet,
and split it into three. The first sheet was corrrect, but the
second and third were almost entirely empty. This has been corrected,
and the sample code was adapted into a formal test with assertions
to confirm that it works as designed.

I made a minor documentation change. Unlike HTML, where you never
need a BOM because you can declare the encoding in the file,
a CSV with non-ASCII characters must explicitly include a BOM
for Excel to handle it correctly. This was explained in the Reading CSV
section, but was glossed over in the Writing CSV section, which I
have updated.


PowerKiKi:
Great job on unit testing, thanks !

lxlkuc:
Hi @oleibman ,thanks for your job, it's great.

But in some situations, I don't need any enclosure for csv file. 

What I want is:
> DRAW_NO,REV
 AYA00022,1

But now I can only get when I `setEnclosure('')`:
> "DRAW_NO","REV"
 "AYA00022","1"

Please consider that.

oleibman:
@lxlkuc Sorry, I did not see your comment till now. However, I submitted another pull request to correct the problem which you mention, and the new request has also been merged. Note that it also adds a new method, setEnclosureRequired, to the Writer/Csv class. The default for enclosureRequired is true, but, if you set it to false, I believe it will do exactly what you want (without having to set enclosure to null string which I believe is a potentially undesirable option).
$writer = new Csv();
$writer->setEnclosureRequired(false)
$writer->save(...)

lxlkuc:
@oleibman Yes I just upgraged to 1.14 and it works now, thanks for your job. 
I'm using Maatwebsite/Laravel-Excel for my project, it seems did not support `setEnclosureRequired()` right now, but `setEnclosure('')` works fine.

Issue #1477, opened 2020-05-19
Support wkhtmltopdf writer


ronisaha:
This is:

```
- [ ] a bug report
- [x] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

It would be nice to have support for wkhtmltopdf

PowerKiKi:
See PR

PR #1478, opened 2020-05-19
Add Snappy as pdf writer

ronisaha:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

This will add a new Pdf writer Adapter to address #1477


PowerKiKi:
I'm not in favor of merging this because we already have three PDF writers and this bring a fourth one. This will add even more maintenance work for us for a very limited added value.

Also from personal, and anecdotal, experience wkhtmltopdf is somewhat hard to work with, because managing binaries can be surprisingly delicate. In the past releases were sparse and the lib was hard to build. Though from a quick look at GitHub this might have improved. But anyway this would not be my first choice for PDF creation.

But more importantly this new PDF writer can, and should, live in its own package. This reduce work for us, give you more freedom for updates and allow the new package to properly declare required dependencies (instead of recommendation like in this PR).

So I'll close this for now.

Issue #1441, opened 2020-04-11
isDateTime doesn't respect formula's


jverelst:
https://github.com/PHPOffice/PhpSpreadsheet/blob/87f71e1930b497b36e3b9b1522117dfa87096d2b/src/PhpSpreadsheet/Shared/Date.php#L328

When you have an excel file with a date-field which is also a formula, this method will return 'false', because it only checks 'getValue()' and not 'getCalculatedValue()'. I am not sure on how that could be fixed, I wrote a small helper function myself using this code as a baseline, but I think that PhpSpreadsheet should support this.

PowerKiKi:
Fixed in #1480

PR #1480, opened 2020-05-20
Fix Issue 1441 (isDateTime and Formulas)

oleibman:
This is:

```
- [ x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

When you have a date-field which is a formula, isDateTime returns false.
https://github.com/PHPOffice/PhpSpreadsheet/issues/1441

Report makes sense; fixed as suggested. Also fixed a few minor
related issues, and added tests so that Shared/Date and Shared/TimeZone
are now completely covered.

Date/setDefaultTimeZone and TimeZone/setTimeZone were not consistent
about what to do in event of failure - return false or throw.
They will now both return false, which is what Date's function
said it would do in its doc block anyhow. Date/validateTimeZone will
continue to throw; it was protected, but was never called outside
Date, so I changed it to private.

TimeZone/getTimeZoneAdjustment checked for 'UST' when it probably
meant 'UTC', and, as it turns out, the check is not even needed.

The most serious problem was that TimeZone/validateTimeZone does not
check the backwards-compatible time zones. The timezone project
aggressively, and very controversially, "demotes" timezones;
such timezones eventually wind up in the PHP backwards-compatible list.
We want to make sure to check that list so that our applications do not
break when this happens.

I bundled into this change two unrelated testing changes.
I think one of my Html tests could be more bulletproof in the event
that a specific assertion failed.
One of the sample CSV's used backslash before quote for escaping,
but, in fact, Excel doubles the quote. This did not invalidate the
test, but it makes the file look weird. An additional assertion
was added to ensure that the cell's contents are correct.


MarkBaker:
Thank you for the submission

Is it possible to unbundled the two unrelated testing changes, and create separate PRs for each of them?

oleibman:
Done. The testing changes have been removed from this PR. I will submit them, plus some other testing changes, through other PRs over the next couple of days.

MarkBaker:
Thank you for this, and for the unbundling as requested

PR #1493, opened 2020-05-28
Xls Writer - Correct Timestamp Bug, Improve Coverage

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

The Xls Writer sets its timestamp incorrectly. The problem is actually
in Shared/Ole::localDateToOLE, which converts its timestamp using
gmmktime; mktime is correct. If I save a file at 3:00 p.m. in San Francisco,
this bug means the time is actually recorded as 3:00 p.m. UTC.
A consequence of this is that if you use Phpspreadsheet to read the
file and save it as a new Xls, the creation timestamp goes further
and further back in time with each generation (or further forward
if east of Greenwich). One of the tests added confirms that
the creation timestamp is consistent with the start and end times
of the test.

The major change in coverage is adding tests to save GIF and BMP
images, which aren't supported in Xls, but are converted to PNG
in the PhpSpreadsheet code.


src/PhpSpreadsheet/Writer/Xls.php
```diff
@@ -389,25 +389,129 @@ private function buildWorksheetEschers(): void
...
+    private function checkForDrawings(): bool
+    {
         // any drawings in this workbook?
         $found = false;
         foreach ($this->spreadsheet->getAllSheets() as $sheet) {
             if (count($sheet->getDrawingCollection()) > 0) {
                 $found = true;
+                if ($this->simulateNoGd || !extension_loaded('gd')) {
+                    throw new RuntimeException('Saving images in xls requires gd extension');
+                }
```
PowerKiKi, line 499: This, and `setSimulateNoGd` and related, should be entirely deleted, because we now use composer to assert for us that all dependencies do exist. And thus our code can assume it is there.

```suggestion
```

oleibman: That does seem like a much better way to handle it. I have included your suggested change in my latest commit/push.

MarkBaker:
Thank you for the fix

PR #1494, opened 2020-05-28
Fix typo in PHPDocs

finagin:
If you want it's may be array of strings like `@return string[]`

This is a typofix

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -1747,7 +1747,7 @@ public function unmergeCellsByColumnAndRow($columnIndex1, $row1, $columnIndex2,
     /**
      * Get merge cells array.
      *
-     * @return array[]
+     * @return array
```
PowerKiKi, line 1750:
```suggestion
     * @return string[]
```

PR #1499, opened 2020-05-31
Add ability to save edited Html/Pdf

oleibman:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

We give users the ability to edit Html/Pdf, but it's a little cumbersome
to use the edited Html for an Html file, and difficult to use it
for a Pdf. I believe we could make it fairly painless in both cases
by allowing the user to set a callback to edit the generated Html.
This can be accomplished with fewer than a dozen lines of very simple code.
I think this would be easier than grabbing the Html in pieces,
editing it, and reassembling it. I think it would also be simpler
than an alternative I considered, namely the addition of a new method
(e.g. saveEditedHtml) to each of the Html and Pdf writers.

One edit that users might like to make when editing html is to add
fallback fonts, something that is not currently available in
PhpSpreadsheet, and might be difficult to add. A natural extension to
that idea would be the use of webfonts, something which is guaranteed
difficult to add. See samples/Basic/17b_Html for an example of this.

None of the PDF writers support webfonts yet. That doesn't mean they
won't do so in future, but, for now, samples/Pdf/21a_Pdf is a prosaic
example of something you could do with this callback. In fact, this
opens the door to letting the user replace the entire body with data
of their choosing, effectively allowing PhpSpreadsheet (where you can
set things like paper size and orientation) to be used as a front-end to
the Pdf processor without the user having to be be overly familiar with
the vagaries of the PDF processor. I think this is actually a pretty
nice idea. YMMV. See samples/Basic/21b_Pdf for an example.


docs/topics/reading-and-writing-to-file.md
```diff
@@ -703,6 +703,31 @@ echo $writer->generateSheetData();
 echo $writer->generateHTMLFooter();
 ```
 
+#### Editing HTML During Save Via a Callback
+
+You can also add a callback function to edit the generated html
+before saving. For example, you could add a webfont
+(not currently supported for Pdf) as follows:
```
PowerKiKi, line 710: We should find a better example than webfont, because they are actually useless as of now. It does not make much sense to demonstrate something that has no purpose.

docs/topics/reading-and-writing-to-file.md
```diff
@@ -703,6 +703,31 @@ echo $writer->generateSheetData();
 echo $writer->generateHTMLFooter();
 ```
 
+#### Editing HTML During Save Via a Callback
```
PowerKiKi, line 706:
```suggestion
#### Editing HTML during save via a callback
```

samples/Basic/17b_Html.php
```diff
@@ -0,0 +1,27 @@
...
+require __DIR__ . '/../Header.php';
+$spreadsheet = require __DIR__ . '/../templates/sampleSpreadsheet.php';
+
+$filename = $helper->getFilename(__FILE__, 'html');
+$writer = new Html($spreadsheet);
+
+function webfont(string $html): string
+{
+    $linktag = <<<EOF
+<link href="https://fonts.googleapis.com/css2?family=Poiret+One&display=swap" rel="stylesheet" />
```
PowerKiKi, line 14: idem, we should find better example, or remove it entirely

samples/Pdf/21a_Pdf.php
```diff
@@ -0,0 +1,47 @@
...
+    $newstyle = <<<EOF
+<style type='text/css'>
+body {
+background-color: yellow;
+}
+</style>
+
+EOF;
+
+    return preg_replace('@</head>@', "$newstyle</head>", $html);
```
PowerKiKi, line 28: Let's stick to either `~` or `/` for regexp delimiters.

```suggestion
    return preg_replace('~</head>~', "$newstyle</head>", $html);
```

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -131,6 +131,13 @@ class Html extends BaseWriter
      */
     private $generateSheetNavigationBlock = true;
 
+    /**
+     * Callback for editing generated html.
+     *
+     * @var callable
+     */
+    protected $editHtmlCallback = '';
```
PowerKiKi, line 139:
```suggestion
     * @var null|callable
     */
    protected $editHtmlCallback = null;
```

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -190,13 +197,27 @@ public function generateHtmlAll()
 
         // Write footer
         $html .= $this->generateHTMLFooter();
+        $cbk = $this->editHtmlCallback;
+        if ($cbk) {
+            $html = $cbk($html);
+        }
```
PowerKiKi, line 203:
```suggestion
        $callback = $this->editHtmlCallback;
        if ($callback) {
            $html = $callback($html);
        }
```

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -190,13 +197,27 @@ public function generateHtmlAll()
...
 
         Calculation::setArrayReturnType($saveArrayReturnType);
         Calculation::getInstance($this->spreadsheet)->getDebugLog()->setWriteDebugLog($saveDebugLog);
 
         return $html;
     }
 
+    public function setEditHtmlCallback(callable $cbk): void
```
PowerKiKi, line 211:
```suggestion
    public function setEditHtmlCallback(?callable $cbk): void
```

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -190,13 +197,27 @@ public function generateHtmlAll()
...
+    {
+        $this->editHtmlCallback = $cbk;
+    }
+
+    public function resetEditHtmlCallback(): void
+    {
+        $this->editHtmlCallback = '';
+    }
+
```
PowerKiKi, line 220: Not needed anymore

```suggestion
```

docs/topics/reading-and-writing-to-file.md
oleibman: I will certainly make the other changes you recommend, but I'm having trouble understanding your statements about webfonts. Do you find them useless in general? Or useless for this application? Or useless for this application because they can't be used for PDF? Knowing your particular objection will make it easier for me to find a better example. On the subject of which, would simply adding fallback generic fonts qualify as better?

oleibman: I have tried something else, hopefully more to your liking.

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -131,6 +131,13 @@ class Html extends BaseWriter
      */
     private $generateSheetNavigationBlock = true;
 
+    /**
+     * Callback for editing generated html.
+     *
+     * @var null|callable
+     */
+    protected $editHtmlCallback;
```
PowerKiKi, line 139:
```suggestion
    private $editHtmlCallback;
```

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -190,13 +197,22 @@ public function generateHtmlAll()
...
         Calculation::setArrayReturnType($saveArrayReturnType);
         Calculation::getInstance($this->spreadsheet)->getDebugLog()->setWriteDebugLog($saveDebugLog);
 
         return $html;
     }
 
+    public function setEditHtmlCallback(?callable $cbk): void
+    {
+        $this->editHtmlCallback = $cbk;
+    }
```
PowerKiKi, line 214:
```suggestion
    /**
     * Set a callback to edit the entire HTML.
     *
     * The callback must accept the HTML as string as first parameter,
     * and it must return the edited HTML as string.
     */
    public function setEditHtmlCallback(?callable $callback): void
    {
        $this->editHtmlCallback = $callback;
    }
```

samples/Pdf/21b_Pdf.php
```diff
@@ -0,0 +1,51 @@
...
+
+use PhpOffice\PhpSpreadsheet\Worksheet\PageSetup;
+use PhpOffice\PhpSpreadsheet\Writer\Pdf\Dompdf;
+use PhpOffice\PhpSpreadsheet\Writer\Pdf\Mpdf;
+use PhpOffice\PhpSpreadsheet\Writer\Pdf\Tcpdf;
+
+function replaceBody(string $html): string
+{
+    $lorem = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.';
+    $bodystring = '@<body>.*</body>@ms';
```
PowerKiKi, line 11:
```suggestion
    $bodystring = '~<body>.*</body>~ms';
```

tests/PhpSpreadsheetTests/Writer/Html/CallbackTest.php
```diff
@@ -0,0 +1,53 @@
...
+        $newstyle = <<<EOF
+<style type='text/css'>
+body {
+    background-color: yellow;
+}
+</style>
+
+EOF;
+
+        return preg_replace('@</head>@', "$newstyle</head>", $html);
```
PowerKiKi, line 23:
```suggestion
        return preg_replace('~</head>~', "$newstyle</head>", $html);
```

PR #1511, opened 2020-06-04
Bug: IValueBinder method bindValue can not be triggered

a443939412:
This is a bugfix


a443939412:
> Please describe here and in the commit message, why this change is necessary.

PR #1514, opened 2020-06-07
Improve Coverage for Sylk

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

I believe that both BaseReader and Sylk Reader are now 100% covered.

Documentation available for this format is sparse.
It was always incomplete, and in some cases inaccurate.
My goal was to use PhpSpreadsheet to load the test file,
save it as Xlsx, and visually compare the two, then add a test
loaded with assertions. Cell values and calculated values,
and border styles were generally handled pretty well without changes.
Other types of styling were not handled so well. I added a few cells
to exercise some previously uncovered code.

Sylk files must be ASCII. I have deprecated the use of the
setEncoding and getEncoding functions, which had no test cases.


MarkBaker:
I know SYLK is a pretty obscure format these days, but it is useful to have better coverage with assertions for the spreadsheet formats that are available. Thank you

Issue #1509, opened 2020-06-04
Can not set empty enclosure for CSV


matejjurancic:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question
```

### What is the expected behavior?
When enclosure is set to empty string, no enclosure should be written to CSV.

### What is the current behavior?
Empty enclosure is changed to double quotes in `PhpOffice\PhpSpreadsheet\Writer\Csv::setEnclosure()`. Code was [recently changed](https://github.com/PHPOffice/PhpSpreadsheet/commit/7517cdd008c5fa874ac18180f5835026b6368e2b) from 
```php
if ($pValue == '') {
    $pValue = null;
}
$this->enclosure = $pValue;
```
to
```php
$this->enclosure = $pValue ? $pValue : '"';
```
and now changes empty string to double quotes.

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();
$writer = new \PhpOffice\PhpSpreadsheet\Writer\Csv($spreadsheet);
$writer->setEnclosure('');
```

### Which versions of PhpSpreadsheet and PHP are affected?
1.13.0
Change has not been documented in the Changelog.

oleibman:
My apologies - it was my pull request that introduced this change. It did not break any existing tests. and I don't understand your use case. Can you please give me a testable example of how you are using this so that I can revert this behavior and add an appropriate test.

matejjurancic:
I am actually using [Laravel Excel](https://github.com/maatwebsite/Laravel-Excel), where the CSV configuration parameters are only passed to phpspreadsheet.
Basically all I want to do is generate a CSV file without enclosures, such as this:
```csv
2020-06-03;000123;06.53;14.22
2020-06-03;000234;07.12;15.44
...
```
I used to achieve this using `$writer->setEnclosure('');`, but this now sets enclosure to `"` and I get such content:
```csv
"2020-06-03";"000123";"06.53";"14.22"
"2020-06-03";"000234";"07.12";"15.44"
...
```
_CSV files that my app generates are imported in another app. That app doesn't allow enclosures so files can't be imported. I know ..._ :roll_eyes:

nickfls:
@oleibman are there any plans to merge it back any time soon?

crnkovic:
I would love an update for plans to merge for this as well.

[...]

PR #1518, opened 2020-06-10
 Fix For #1509

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

User expected no CSV enclosures after $writer->setEnclosure(''),
which had been changed to be consistent with $reader->setEnclosure('').
Writer will now omit enclosures after code above; no change to Reader.
Tests have been added for this condition.


MarkBaker:
Thank you for the quick resolution to the issue with setting empty CSV enclosures

atreitsma:
@MarkBaker could you be so kind and make this a new release so we don't need workarounds to make **CSV Enclosure** working as expected again in `Maatwebsite/Excel` ? since it uses the latest release... thanks in advance!

MarkBaker:
I really want to get the changes that I'm doing rewriting named ranges/formulae completed before doing a new release (perhaps another week or two of work), because the next steps that I'm planning after that could lead to some bc breaks, which would require a new 2.0.0 release.
Is it possible for you to work with dev-master for that time? Or does the Laravel wrapper restrict you to tagged releases?

SudoGetBeer:
> I really want to get the changes that I'm doing rewriting named ranges/formulae completed before doing a new release (perhaps another week or two of work), because the next steps that I'm planning after that could lead to some bc breaks, which would require a new 2.0.0 release.
> 
> Is it possible for you to work with dev-master for that time? Or does the Laravel wrapper restrict you to tagged releases?

You can just use  "phpoffice/phpspreadsheet": "1.12.0", in composer and everything works 😉

atreitsma:
Even though @MarkBaker just tagged a new release, the requirements changed for Guzzle 6 -> 7 and that is a no go (it's like domino, now with Google API) ... so no solution to update to latest tagged releases and that indeed leaves the option as stated by @SudoGetBeer open as final solution for now. I do understand the choice @MarkBaker to aim for a bigger release... and will keep an eye out on package updates so this will integrate nicely again.

PR #1520, opened 2020-06-12
Fix exact MATCH on ranges with empty cells

arnested:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Fixes a bug when doing exact match on ranges with empty cells.

```php
<?php

// Row: 1, null, 4, null, 8.
$sheet->getCell('A1')->setValue(1);
$sheet->getCell('A3')->setValue(4);
$sheet->getCell('A5')->setValue(8);

$sheet->getCell('B1')->setValue('=MATCH(4, A1:A5, 1)');

// Should echo 3, but echos '#N/A'.
echo $sheet->getCell('B1')->getCalculatedValue() . PHP_EOL;

// Row: 1, null, 4, null, null.
$sheet->getCell('C1')->setValue(1);
$sheet->getCell('C3')->setValue(4);

$sheet->getCell('D1')->setValue('=MATCH(5, C1:C5, 1)');

// Should echo 3, but echos '#N/A'.
echo $sheet->getCell('D1')->getCalculatedValue() . PHP_EOL;
```
Before fix:
```
#N/A
#N/A
```

After fix:
```
3
3
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 ### Fixed
 
 - Resolve evaluation of utf-8 named ranges in calculation engine [#1522](https://github.com/PHPOffice/PhpSpreadsheet/pull/1522)
+- Fix exact MATCH on ranges with empty cells [#1520](https://github.com/PHPOffice/PhpSpreadsheet/pull/1520)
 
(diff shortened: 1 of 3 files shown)


MarkBaker:
Thank you for this fix. This (and your other MATCH fix) weren't cases that we considered when we're developing the original code logic to implement MATCH()

PR #1521, opened 2020-06-12
Fix MATCH when comparing different numeric types

arnested:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Let MATCH compare numerics of different type (e.g. integers and floats).

```php
<?php

// Row: 1, 2, 3, 4, 5. MATCH for 4.6.
$sheet->getCell('A1')->setValue(1);
$sheet->getCell('A2')->setValue(2);
$sheet->getCell('A3')->setValue(3);
$sheet->getCell('A4')->setValue(4);
$sheet->getCell('A5')->setValue(5);

$sheet->getCell('B1')->setValue('=MATCH(4.6, A1:A5, 1)');

// Should echo 4, but echos '#N/A'.
echo $sheet->getCell('B1')->getCalculatedValue() . PHP_EOL;

// Row: 1, 2, 3, 3.8, 5. MATCH for 4.
$sheet->getCell('C1')->setValue(1);
$sheet->getCell('C2')->setValue(2);
$sheet->getCell('C3')->setValue(3);
$sheet->getCell('C4')->setValue(3.8);
$sheet->getCell('C5')->setValue(5);

$sheet->getCell('D1')->setValue('=MATCH(4, C1:C5, 1)');

// Should echo 4, but echos 3.
echo $sheet->getCell('D1')->getCalculatedValue() . PHP_EOL;
```
Before fix:
```
#N/A
3
```

After fix:
```
4
4
```


arnested:
I discovered this with a `.xlsx` I received where `MATCH` where used on dates. The dates in the range where all interpreted as integers whereas the date used as lookup value was interpreted as a float.

MarkBaker:
Thank you for this (and your other MATCH issue) fix.

PR #1522, opened 2020-06-13
Resolve utf-8 named ranges in calculation engine

MarkBaker:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

Modify the Calculation engine so that UTF-8 named ranges will be recognised as valid rather than generating an error


arnested:
Works on my sample of (four) real life files. 👏

Issue #1505, opened 2020-06-02
TypeError : Argument 4 passed to PhpOffice\PhpSpreadsheet\Writer\Xlsx\Worksheet::writeAttributeIf() must be of the type string


laurin1:
I just updated to 1.13.0 and now when I try to save an XLS file that I just imported, I get this error:

TypeError : Argument 4 passed to PhpOffice\PhpSpreadsheet\Writer\Xlsx\Worksheet::writeAttributeIf() must be of the type string, null given, called in C:\inetpub\Intranet_Local\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Writer\Xlsx\Worksheet.php on line 1164
C:\inetpub\Intranet_Local\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Writer\Xlsx\Worksheet.php:453
C:\inetpub\Intranet_Local\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Writer\Xlsx\Worksheet.php:1164
C:\inetpub\Intranet_Local\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Writer\Xlsx\Worksheet.php:1040
C:\inetpub\Intranet_Local\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Writer\Xlsx\Worksheet.php:67
C:\inetpub\Intranet_Local\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Writer\Xlsx.php:267
 C:\inetpub\Intranet_Local\library\Intranet\PhpOffice\XlsxWriter.php:20
 C:\inetpub\Intranet_Local\phpunit\library\Intranet\PHPExcelTest.php:52

I am not making any changes to the file, just import, then save:

```php
$sFileImportedFromEmail      = _TEST_DATA_PATH."/imported_from_email01.xls";
$sFileImportedFromEmailSaved = _TEST_DATA_PATH."/imported_from_email_saved01.xls";

$this->expectOutputString(
	"");

try{

	$oSpreadsheet =
		IOFactory::load(
			$sFileImportedFromEmail);
	$oXlsxWriter  =
		new XlsxWriter(
			$oSpreadsheet);

	$oXlsxWriter->save(
		$sFileImportedFromEmailSaved);

}
catch(Exception $oException){

	Exceptions::convertToError(
		$oException);

	$aSpreadsheet = [];

}
```

[...]

Issue #1238, opened 2019-11-13
Notice and error when reading .xls (BIFF8) having invalid xfIndex


dendemann:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] not a usage question
```

### What is the expected behavior?
Reading an Excel 97-2003 document shouldn't throw notices or errors. 

### What is the current behavior?
Reading an Excel 97-2003 documents isn't possible when the data is corrupted in a certain way.

### What are the steps to reproduce?

Legacy .xls documents created with third party libraries might contain corrupt data. In my case xfIndex read from a cell is 0 while there's only one entry within the mapCellXfIndex mapping array at index position 15. Therefore there will be an unchecked array access at pos 0.

Unfortunately I can't provide the file as it contains sensitive customer data. But I found out that, according to the shared strings table (sst), it was originally created by the library "GemBox.Spreadsheet Professional 3.3 for .NET 4.0, Version=33.3.40.1062".

Another guy ran into the same problem but I think no one was looking into it: [432120730](https://github.com/PHPOffice/PhpSpreadsheet/issues/402#issuecomment-432120730)

Reading the file with Excel or LibreOffice works without problems. 

```php
<?php

use PhpOffice\PhpSpreadsheet\Reader\Xls;

require __DIR__ . '/vendor/autoload.php';

$reader = new Xls();
$spreadsheet = $reader->load(__DIR__ . '/broken.xls');

$cell = $spreadsheet->getActiveSheet()->getCell('A1');
var_dump($cell->getFormattedValue());
```
```bash
PHP Notice:  Undefined offset: 0 in /home/user/test/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xls.php on line 3830
PHP Stack trace:
PHP   1. {main}() /home/user/test/test.php:0
PHP   2. PhpOffice\PhpSpreadsheet\Reader\Xls->load() /home/user/test/test.php:8
PHP   3. PhpOffice\PhpSpreadsheet\Reader\Xls->readLabelSst() /home/user/test/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xls.php:956
PHP Notice:  Undefined index:  in /home/user/test/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Spreadsheet.php on line 1034
PHP Stack trace:
PHP   1. {main}() /home/user/test/test.php:0
[...]

PR #1525, opened 2020-06-15
Fix for #1505

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

This problem is the same as #1238, which was resolved by #1239.
For that issue, the fix was to check in one place whether
$this->mapCellXfIndex[$xfIndex] was set before using it.
The sample spreadsheet supplied as a description for this
problem had exactly the same problem in 2 other places in the code.
In addition, there were 7 other places in the code where that
particular item was used unchecked. This fix corrects all 9 locations.
The spreadsheet supplied with the problem is used as the basis
for some new tests, which particularly test column dimensions
and styles, the problems involved in this case.


MarkBaker:
Thank you for both this and the #1238 resolution

Issue #1533, opened 2020-06-16
A reference to a cell containing a string starting with "#" leads to errors in the generated xlsx.


keylevel:
This is:

```
- [X] a bug report
- [ ] a feature request
- [X] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
That a valid spreadsheet be produced when a cell contains a reference to another cell.

### What is the current behavior?
Excel reports an error when opening a file that has been created with a reference to a cell that contains a string that starts with '#'.

### What are the steps to reproduce?
The example below sets 'A1' to "#define M" and sets 'A2' as a reference to this.

The xlsx file that is produced gets an "Alert" dialog showing "We found a problem with some content in ’test.xlsx’. Do you want us to try to recover as much as we can? If you trust the source of this workbook, click Yes." when it is opened with Excel.

Changing the content of 'A1' to "'#define M" (note the extra ' at the start) results in the generation of a valid file.

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

// add code that show the issue here...
$sheet = $spreadsheet->getActiveSheet();

$sheet->setCellValueExplicit( "A1", "#define M",  \PhpOffice\PhpSpreadsheet\Cell\DataType::TYPE_STRING );
$sheet->setCellValue( "A2", "=A1" );

$writer = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($spreadsheet, 'Xlsx');
$writer->save('./test.xlsx');


### Which versions of PhpSpreadsheet and PHP are affected?
PhpSpreadsheet 1.13.0, PHP 7.3.11

keylevel:
Thanks guys - I've copied that patch to my setup and can confirm that everything is now working as expected.

Should I close this issue now, or would it be better to wait until it is in a tagged release?

MarkBaker:
The fix is now included in the 1.14.0 release

PR #1534, opened 2020-06-17
Fix for #1533

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Code assumes that formula whose result starts with # indicates error.
Change to check entire result against error list in Functions.


MarkBaker:
Yes, a bad assumption to make. Thank you for the fix.

Issue #1235, opened 2019-11-08
Writer\Html output is much too large


werbfred:
This is:

```
- [ ] a bug report
- [x ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

When using the save method from **PhpOffice\PhpSpreadsheet\Writer\Html**, we could expect that only the usefull data is being outputed. This works fine for **PhpOffice\PhpSpreadsheet\Writer\Csv**.

### What is the current behavior?

Currently **PhpOffice\PhpSpreadsheet\Writer\Html** uses the method **_Worksheet::calculateWorksheetDimension()_** which is relying on **_Worksheet::getHighestColumn()_**.

In this case it could be better to rely on **_Worksheet::getHighestDataColumn()_**. This would avoid displaying a lot of empty cells on the right side of the html table. HTML file would also be much smaller.

### What are the steps to reproduce?

Call **_PhpOffice\PhpSpreadsheet\Writer\Html::save()_** and have the generated file rendered in your favortie web browser.

### Which versions of PhpSpreadsheet and PHP are affected?

Php version 7.3
PhpSpreadsheet version 1.9

mathias-meyer:
Looking at the source code I would suggest that in **_Html::generateSheetData()_** the line 

```
$dimension = explode(':', $sheet->calculateWorksheetDimension());
```

should be replaced by

```
$dimension = explode(':', $sheet->calculateWorksheetDataDimension());
```
That would do the job I believe.

mathias-meyer:
Any feedback if this feature makes sense ?

adjenks:
Only the stale bot is responding to my tickets as well.

mathias-meyer:
Up

PowerKiKi:
I guess it makes sense. Have any of you had a look a git blame to see if there is a reason why we don't already do that?

werbfred:
Should I make a Pull Request ?

werbfred:
Pull Request raised : **Writer\Html output is much too large #1537**

PR #1537, opened 2020-06-18
Writer\Html output is much too large

werbfred:
…stDataColumn() which seems to be more adequate than Worksheet::calculateWorksheetDimension which uses Worksheet::getHighestColumn(). This latest one includes all available columns not those with effective data.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

When using the save method from **PhpOffice\PhpSpreadsheet\Writer\Html**, we could expect that only the usefull data is being outputed. This works fine for **PhpOffice\PhpSpreadsheet\Writer\Csv*

### DIFF
--- a/src/PhpSpreadsheet/Writer/Html.php
+++ b/src/PhpSpreadsheet/Writer/Html.php
@@ -428,7 +428,7 @@ class Html extends BaseWriter
             // Write table header
             $html .= $this->generateTableHeader($sheet);
             // Get worksheet dimension
-            $dimension = explode(':', $sheet->calculateWorksheetDimension());
+            $dimension = explode(':', $sheet->calculateWorksheetDataDimension());
             $dimension[0] = Coordinate::coordinateFromString($dimension[0]);
             $dimension[0][0] = Coordinate::columnIndexFromString($dimension[0][0]);
             $dimension[1] = Coordinate::coordinateFromString($dimension[1]);


werbfred:
This Pull Request is related to issue : **Writer\Html output is much too large #1235**

MarkBaker:
There is a block of code in the HTML Writer that tries to identify images that exist in the spreadsheet below the last data row, and to ensure that they are included in the html output. I suspect that this change would break that.
In the case of a csv file, that isn't an issue, because images aren't written to csv output

PowerKiKi:
@MarkBaker, I added tests for the case you mention and fixed the bug

aagarwal8:
@PowerKiKi the same issue exists when there are charts or images after the last column. Can a similar fix be applied for it?

PowerKiKi:
if there is something after the last column, then it is correct to output a "too large" spreadsheet in order to show that thing, isn't it ? or are you talking about something else ?

aagarwal8:
Perhaps, but it is not working as expected.

Spreadsheet with 6 columns, 50 odd rows; couple of charts placed over columns G thru K. 
The only way HTML writer is showing the charts is if we populate a cell after column K (doesnt matter what we put in the cell; even a space works).

PowerKiKi:
What you describe is the "opposite" of OP. OP said "there is too many blank cells", but you're saying "there's not enough blank cell to see my chart". You should open a new PR to fix that, or at least a PR with a failing test that demonstrate the issue.

Issue #1523, opened 2020-06-13
Corrupted file message on open file in office 2007-2016 (tested) If a print area is specified on the sheet


tsukasa-mixer:
This is:

```
- [X] a bug report
- [ ] a feature request
- [X] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

use PhpOffice\PhpSpreadsheet\Spreadsheet;
use PhpOffice\PhpSpreadsheet\IOFactory;
use PhpOffice\PhpSpreadsheet\Worksheet\PageSetup;

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object

        $book = new Spreadsheet();

        $sheet = $book->getActiveSheet();
        $sheet
            ->getPageSetup()
            ->setOrientation(PageSetup::ORIENTATION_LANDSCAPE)
            ->setPaperSize(PageSetup::PAPERSIZE_A4)
            ->setFitToPage(true)
            ->setFitToWidth(1)
            ->setFitToHeight(0)
        ;

        $book->getActiveSheet()->getPageSetup()->addPrintAreaByColumnAndRow(
                1, 1,10, 10, 1
        );


        IOFactory::createWriter($book, 'Xlsx')->save('testFile.xlsx');

// add code that show the issue here...
```

### Which versions of PhpSpreadsheet and PHP are affected?
actual

diff on saved office and library generrated.
![2020-06-08__18:58:04](https://user-images.githubusercontent.com/1190435/84569947-c9299f00-ad92-11ea-83a3-5883322ad1a1.png)

incorrect code:
https://github.com/PHPOffice/PhpSpreadsheet/blob/7ab920de5bf96d75298c17434998eebd49f3e0b4/src/PhpSpreadsheet/Writer/Xlsx/DocProps.php#L56-L58

https://github.com/PHPOffice/PhpSpreadsheet/blob/7ab920de5bf96d75298c17434998eebd49f3e0b4/src/PhpSpreadsheet/Writer/Xlsx/DocProps.php#L68-L77

tsukasa-mixer:
to fix it, you need to add additional ones for each sheet with printable zones
[...]

PR #1544, opened 2020-06-24
PrintArea Causes Exception

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

I wanted to investigate #1523, but I couldn't duplicate its results
because the sample code in the issue caused an Exception to be thrown.
The exception happened because, when the Print Range Set method is
set to Insert (as oppposed to Overwrite), and the Print Range is
empty, it is created with a leading or trailing comma, and Writer/Xlsx
objects. This is, in a sense, a user error, but the software should
be more resilient. This can be accomplished by ensuring that no
leading or trailing comma is generated when Print Range is empty.

This code change fixes that problem. Since I couldn't reproduce the
original, I can't say for sure that it fixes it. However, with the
sample code provided, I can write a spreadsheet which Excel reads
without any problems, so it probably fixes the original.


MarkBaker:
Print areas are a pain point; a simple range isn't a problem, and that's the majority of cases, but there are instances where people set more complex multiple-print areas. Thank goodness that there is no reason for anybody to want to set print areas using intersections (space ` ` operator) rather than simply unions (comma `,` operator).
In theory, the same could apply to filter ranges or to print titles; but I can't think of any use-case where a simple range isn't enough for those; and we don't support them anyway.

PR #1557, opened 2020-06-27
Improving Coverage for Excel2003 XML Reader

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Reader/Xml is now 100% covered.

File templates/Excel2003XMLTest.xml, used in some tests, is *not*
readable by a current version of Excel. I have substituted a new file
excel2003.xml to be used in its place. I have not deleted the original
in case someone in future (possibly me) wants to see what it needs to
make it usable.

There are minimal code changes.
- Unused protected functions pixel2WidthUnits and widthUnits2Pixel
  are deleted.
- One regex looking to convert hex characters is changed from a-z to a-f,
  and made case insensitive.
- No calculation performed for "error" cell (previously calculation
  was attempted and threw exception).
- Empty relative row/cell is now handled correctly.
- Style applied to empty cell when appropriate.
- Support added for textRotation.
- Support added for border styles.
- Support added for diagonal borders.
- Support added for superscript and subscript.
- Support added for fill patterns.

In theory, encodings other than UTF-8 were supported.
In fact, I was unable to get SecurityScanner to pass *any* xml which is
not UTF-8. Eliminating the assumption that strings might not be UTF-8
allowed much of the code to be greatly simplified.
After that, I added some code that would permit the use of
some ASCII-compatible encodings (there is a test of ISO-8859-1).
It would be more difficult to handle other encodings (such as UTF-16).
I am not convinced that even the ISO-8859 effort is worth it,
but am willing to investigate either expanding or eliminating
non-UTF8 support.

I added a number of tests, creating an Xml directory, and moving
XmlTest to that directory.

Pull Request had problems reading old invalid sample in the code
coverage phase, not in any of the other test phases, and not in
the code coverage phase on my local machine.
As it turns out, aside from being invalid, the sample
is much larger than any of the other samples. Tests have been
adjusted accordingly.


tests/PhpSpreadsheetTests/Reader/Xml/XmlLoadTest.php
```diff
@@ -0,0 +1,175 @@
...
+        self::assertEquals(Borders::DIAGONAL_BOTH, $sheet->getCell('E18')->getStyle()->getBorders()->getDiagonalDirection());
+        self::assertEquals(Borders::DIAGONAL_DOWN, $sheet->getCell('I18')->getStyle()->getBorders()->getDiagonalDirection());
+        self::assertEquals(Borders::DIAGONAL_UP, $sheet->getCell('J18')->getStyle()->getBorders()->getDiagonalDirection());
+        self::assertEquals(Font::UNDERLINE_DOUBLE, $sheet->getCell('A24')->getStyle()->getFont()->getUnderline());
+        self::assertTrue($sheet->getCell('B23')->getStyle()->getFont()->getSubScript());
+        self::assertTrue($sheet->getCell('B24')->getStyle()->getFont()->getSuperScript());
+        //self::assertFalse($sheet->getRowDimension(30)->getVisible());
+        $hyperlink = $sheet->getCell('A21');
+        self::assertEquals('PhpSpreadsheet', $hyperlink->getValue());
+        self::assertEquals('https://github.com/PHPOffice/PhpSpreadsheet', $hyperlink->getHyperlink()->getUrl());
```
MarkBaker, line 123: This barrage of assertions should probably be broken down into a series of separate tests for each of the different style elements. It will take slightly longer to run, as the file has to be loaded for each test; but separation of tests will help with readability, and with expanding further in the future

MarkBaker:
It's good to have better testing to ensure that future changes don't break anything, and I appreciate all your efforts for helping us improve the PHPSpreadsheet library

tests/PhpSpreadsheetTests/Reader/Xml/XmlLoadTest.php
oleibman: I will do so. I probably can't get to it till the weekend.

oleibman: Tests have now been repackaged as you requested. When I pushed the changes, only Scrutinizer ran, not Travis. I do not know how to force Travis to run in this circumstance.

oleibman:
Removing stalebot tag.

PR #1566, opened 2020-07-03
Remove undefined index notice in Writer\Xlsx\Worksheet

Zombaya:
Preventing a notice for undefined index when `$pFlippedStringTable`  does not contain index `$cellValue`.

This happened when `null` was being written to a string-cell.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

This bug prevents me from saving xlsx's.


Zombaya:
I still think this MR has value and that it should be merged into the code.

Zombaya:
This is fixed in 1.6.0 using nullcoalescence.  Works equally well.

PR #1567, opened 2020-07-03
Bug Setting Superscript/Subscript to False

oleibman:
If font style Superscript is set to true, Subscript is set to false.
Likewise, setting Subscript to true sets Superscript to false.
Both of these are working as they should. However,
setting Superscript to false causes Subscript to be set to true,
and setting Subscript to false causes Superscript to be set to true.
I believe that is an error in both cases. This change fixes it.

There seem to be no existing tests for Font styles.
I added the tests necessary to validate this change.
I will put adding more on my to-do list.

I faced a conundrum changing setSuperscript and setSubscript.
The parameter is described as bool in the DocBlock.
It is not typed in the function declaration, and changing the function
declaration to type it runs a serious risk of breaking existing
applications for no good reason. So I want to cast it on entry.
But I believe Scrutinizer would object to casting to bool a parameter
declared as bool in the DocBlock. I tried !!, but Phpunit objected
to this, suggesting using a cast to bool instead. So I did this
as two separate assignments, which satisfies Phpunit, and which
I hope satisfies Scrutinizer.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Style/Font.php
+++ b/src/PhpSpreadsheet/Style/Font.php
@@ -363,15 +363,16 @@ class Font extends Supervisor
      */
     public function setSuperscript($pValue)
     {
-        if ($pValue == '') {
-            $pValue = false;
-        }
+        $pValu2 = !$pValue;
+        $pValue = !$pValu2;
         if ($this->isSupervisor) {
             $styleArray = $this->getStyleArray(['superscript' => $pValue]);
             $this->getActiveSheet()->getStyle($this->getSelectedCells())->applyFromArray($styleArray);
         } else {
             $this->superscript = $pValue;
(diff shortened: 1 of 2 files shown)


oleibman:
Removing stalebot tag.

PowerKiKi (changes requested):
Because it was already documented as boolean, and because we'd like to move toward typed code (when appropriate), I think we should add the typing and remove the confusing double affectation. Would you mind doing that please?

oleibman:
PowerKiki,
No problem at all. Just didn't want to introduce a potentially breaking change without permission. I have made, committed, and pushed the requested changes.

PR #1568, opened 2020-07-03
Rely on PSR-18 and Guzzle/psr7 only instead of full Guzzle.

### LINKED ISSUE #1562: Guzzle 6 support
This is:

```
- [ ] a bug report
- [x] a feature request
```

PHPSpreadsheet 1.14 only supports Guzzle 7. As it is only released a few days ago most of us are still on Guzzle 6. Please lower the requirements for Guzzle so we can take advantage of the bugfixes.

Pierstoval:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected
- [ ] Documentation is updated as necessary (note from PR author: this feature isn't yet documented)

Fixes #1562, check the issue for more details

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -57,7 +57,8 @@
         "markbaker/complex": "^1.4",
         "markbaker/matrix": "^1.2",
         "psr/simple-cache": "^1.0",
-        "guzzlehttp/guzzle": "^7.0"
+        "psr/http-client": "^1.0",
+        "guzzlehttp/psr7": "^1.6"
     },
     "require-dev": {
         "dompdf/dompdf": "^0.8.5",
--- a/src/PhpSpreadsheet/Calculation/Web.php
+++ b/src/PhpSpreadsheet/Calculation/Web.php
@@ -5,6 +5,7 @@ namespace PhpOffice\PhpSpreadsheet\Calculation;
 use GuzzleHttp\Psr7\Request;
 use PhpOffice\PhpSpreadsheet\Settings;
 use Psr\Http\Client\ClientExceptionInterface;
+use Psr\Http\Client\ClientInterface;
 
 class Web
 {
@@ -18,7 +19,7 @@ class Web
      *
      * @return string the output resulting from a call to the webservice
      */
-    public static function WEBSERVICE(string $url)
+    public static function WEBSERVICE(string $url, ?ClientInterface $client = null)
     {
         $url = trim($url);
         if (strlen($url) > 2048) {
@@ -30,7 +31,7 @@ class Web
         }
 
         // Get results from the the webservice
-        $client = Settings::getHttpClient();
+        $client = $client ?: Settings::getDefaultHttpClient();
         $request = new Request('GET', $url);
 
         try {
--- a/src/PhpSpreadsheet/Settings.php
+++ b/src/PhpSpreadsheet/Settings.php
@@ -8,6 +8,8 @@ use PhpOffice\PhpSpreadsheet\Chart\Renderer\IRenderer;
 use PhpOffice\PhpSpreadsheet\Collection\Memory;
 use Psr\Http\Client\ClientInterface;
 use Psr\SimpleCache\CacheInterface;
+use RuntimeException;
+use Symfony\Component\HttpClient\HttpClient;
 
 class Settings
 {
@@ -43,13 +45,6 @@ class Settings
      * @var CacheInterface
      */
     private static $cache;
-
-    /**
-     * The HTTP client implementation to be used for network request.
-     *
-     * @var ClientInterface
-     */
-    private static $client;
 
     /**
      * Set the locale code to use for formula translations and any special formatting.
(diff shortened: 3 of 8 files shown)


src/PhpSpreadsheet/Settings.php
```diff
@@ -166,23 +161,19 @@ public static function getCache()
         return self::$cache;
     }
 
-    /**
-     * Set the HTTP client implementation to be used for network request.
-     */
-    public static function setHttpClient(ClientInterface $httpClient): void
```
Pierstoval, line 172: Removed this because it must be provided at runtime by the user. Some HTTP clients implementing PSR-18 might be scoped, so it's safer to inject it at runtime instead of keeping a client in the memory

src/PhpSpreadsheet/Settings.php
```diff
@@ -166,23 +161,19 @@ public static function getCache()
...
+    public static function getDefaultHttpClient(): ClientInterface
     {
-        if (!self::$client) {
-            self::$client = new Client();
+        if (class_exists(Client::class)) {
+            // Support for Guzzle
+            return new Client();
+        } elseif (class_exists(HttpClient::class)) {
+            // Support for Symfony HttpClient
+            return HttpClient::create();
```
Pierstoval, line 174: Let's provide support for at least 2 clients, just in case 😉

composer.json
```diff
@@ -57,7 +57,8 @@
         "markbaker/complex": "^1.4",
         "markbaker/matrix": "^1.2",
         "psr/simple-cache": "^1.0",
-        "guzzlehttp/guzzle": "^7.0"
+        "psr/http-client": "^1.0",
+        "guzzlehttp/psr7": "^1.6"
```
Pierstoval, line 61: `guzzlehttp/psr7` is safe, compared to `guzzlehttp/guzzle`, because it only contains the tools to create the `Request` object, which would be cumbersome if we had to create it directly in PhpSpreadsheet.

PowerKiKi: If we do the PSR way, we hat to do it all the way. That means dropping all references to Guzzle and instead rely on https://www.php-fig.org/psr/psr-17/ to create Request object.

src/PhpSpreadsheet/Calculation/Web.php
```diff
@@ -18,7 +19,7 @@ class Web
      *
      * @return string the output resulting from a call to the webservice
      */
-    public static function WEBSERVICE(string $url)
+    public static function WEBSERVICE(string $url, ?ClientInterface $client = null)
```
PowerKiKi, line 22: The method signature must not change, because in most cases the user is not directly calling this method, instead the calculation engine is. And the calculation engine cannot provide extra parameters such as this new one.

src/PhpSpreadsheet/Settings.php
PowerKiKi: > might be scoped

What to you mean be that ? And what difference does it make for PhpSpreadsheet if an HTTP client is scoped ?

src/PhpSpreadsheet/Settings.php
PowerKiKi: This won't do. You removed dependencies on implementation in composer.json, but you re-add add them here in code. All references to specific implementations must disappear.

src/PhpSpreadsheet/Settings.php
PowerKiKi, line 172: This should probably be kept as:

```php
public static function setHttpClient(ClientInterface $httpClient, RequestFactoryInterface $requestFactory): void
```

Because both parameters must be set, in a coherent way, in order for WEBSERVICE to work.

tests/PhpSpreadsheetTests/Calculation/Functions/Web/HttpClientMock.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+namespace PhpOffice\PhpSpreadsheetTests\Calculation\Functions\Web;
+
+use Closure;
+use Psr\Http\Client\ClientInterface;
+use Psr\Http\Message\RequestInterface;
+use Psr\Http\Message\ResponseInterface;
+
+class HttpClientMock implements ClientInterface
```
PowerKiKi, line 10: This entire class should be replaced by PHPunit mock mechanic in order to also test that the method is called exactly once.

src/PhpSpreadsheet/Settings.php
Pierstoval: When using Symfony HTTP Client, it can be scoped, meaning it can have a base uri, default headers, etc., which might not be suitable for PhpSpreadsheet usage. Some HTTP clients can also contain state: history, cookies, etc., so keeping its state in the memory can have side-effects, such as leaking state, etc.

src/PhpSpreadsheet/Settings.php
Pierstoval: Well, these are **optional** dependencies, since `class_exists()` is used, so if the class doesn't exist, it will fall back to an exception, telling the user to either inject an http client, or install one of the two supported implementations.

I forgot to do it in the PR, but I was about to add a `suggest` section in `composer.json` to invite users to install either `symfony/http-client` or `guzzlehttp/guzzle` to benefit from the `WEBSERVICE` feature.

This is kinda like automatic feature-flag depending on what composer packages are installed in the project

tests/PhpSpreadsheetTests/Calculation/Functions/Web/HttpClientMock.php
Pierstoval: I don't see any value in checking the number of times the client is tested.

I could change the mock to remove the internal response when `sendRequest()` is used, this will ensure it's used only once.

Also, using phpunit mocks has a performance impact. Here, having a custom mock is costless

composer.json
Pierstoval: Even though I definitely agree on PSR-17, there are some kind of maintainability issues with it: this means that PhpSpreadsheet would have to have its own factory, plus its own Request class, plus its own Response class for testing.

Here, relying on `guzzlehttp/psr7` is safer than the entire Guzzle stack, because `guzzlehttp/psr7` depends only on `ralouphie/getallheaders` version 2 **and** 3, making it compatible with almost all packages that might already have it in their dependency tree. For instance, Drupal needs it since 8.8, Slim needs it too, but they depend on its v3. There are very few packages that need its v2, and yet guzzle's PSR-7 implementation is still compatible with it.

Seems really fair in terms of balance between maintainability and dependency tree 😉

patrickbrouwers:
@Pierstoval are you planning to improve your PR based on @PowerKiKi 's comments?

Pierstoval:
Thanks for taking the work back 👍

PR #1571, opened 2020-07-05
Restoring State After Static Changes in Tests

oleibman:
This request does not change any source code, only tests.

For a change on which I was working, a test passed when run on its own,
but failed when run as part of the full test suite. It turned out that
an existing test had changed a static value,
thousands separator in this case, and failed to restore it.
The test turned out to be AdvancedBinderTest.

The search for the offending test was more difficult than it should have
been because 26 test scripts which had nothing to do with thousands
separator nevertheless changed that value. They all changed
decimal separator, currency code, and compatibility mode as well,
again for no reason. I changed all of those to eliminate those operations.

I changed the following tests, which actually do change the static
properties identified above for a reason, to restore them as part of teardown.
- CalculationTest sets compatibilityMode and locale
- DayTest sets compatibilityMode, returnDateType, and excelCalendar
- CountTest sets compatibilityMode
- FunctionsTest sets compatibilityMode and returnDateType
- AdvancedValueBinderTest sets currencyCode, decimalSeparator, thousandsSeparator
- StringHelperTest sets currencyCode, decimalSeparator, thousandsSeparator
- NumberFormatTest sets currencyCode, decimalSeparator, thousandsSeparator
- HtmlNumberFormatTest sets currencyCode, decimalSeparator, thousandsSeparator

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Improving test suite (see above for full explanation).

### DIFF
--- a/tests/PhpSpreadsheetTests/Calculation/CalculationTest.php
+++ b/tests/PhpSpreadsheetTests/Calculation/CalculationTest.php
@@ -9,15 +9,23 @@ use PHPUnit\Framework\TestCase;
 
 class CalculationTest extends TestCase
 {
+    private $compatibilityMode;
+
+    private $locale;
+
     protected function setUp(): void
     {
+        $this->compatibilityMode = Functions::getCompatibilityMode();
+        $calculation = Calculation::getInstance();
+        $this->locale = $calculation->getLocale();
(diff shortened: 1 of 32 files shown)


MarkBaker:
Useful as we extend the test suite to cover more cases, thanks

PR #1576, opened 2020-07-09
Add FORMAT_ACCOUNTING_CNY into NumberFormat

Tabll:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

When I need to generate the following Chinese currency format

image

This takes me many times, so is it possible to add constants to help other users?


oleibman:
Removing stale label. This change should have been rejected, like PR #3113, in favor of the more robust Accounting and Currency Wizards introduced with PR #3334 in February 2023.

Issue #580, opened 2018-07-04
Get style array from cell


creatorness:
This is:

- [x ] a feature request

### What is the expected behavior?
Reverse funktion for applyFromArray to get the style array of a Cell

### What is the current behavior?
None existing.
I have to laboriously put together from the individual components of the style class.

### Which versions of PhpSpreadsheet and PHP are affected?
1.3.1

PowerKiKi:
What would be the use case for this new feature?

creatorness:
To get the style from an excel document quickly and save as configuration.
Then add it to a new document without having to worry about sorting everything out, which saves memory and time on a very large template document.

Erwane:
Reopen and use case : 
I wan't to unit test my Spreadsheet generation and ... test styles to.

Tests can be very simple with a `PhpOffice\PhpSpreadsheet\Style::toArray()` method
Thanks

PowerKiKi:
@Erwane, I'd accept a PR for this, as long as it is unit tested of course.

oleibman:
Addressed by PR #1580.

PR #1580, opened 2020-07-12
Add exportArray Method for Styles

oleibman:
Issue #580 has gone stale since I started work on this.
Nevertheless, this implements an exportArray function as an
exact counterpart of applyFromArry.
I chose the name exportArray to avoid confusion with the existing
method getStyleArray, which does something completely different.

This change also increases coverage for all the Style classes to 100%,
with the exception of Style.php itself. There were several (unchanged)
places in Style.php where I did not have sufficient understanding of
what was supposed to be happening, so could not create tests.

All properties used by applyFromArray are exported by this method.
Note that conditional styles are not covered; this is consistent
with the fact that they are not covered by applyFromArray.

The method is implemented as a final public function in Style/Supervisor,
which calls abstract protected function exportArray1, which is implemented
in each of the subclasses, and which calls final protected
function exportArray2 in Style/Supervisor.
So exportArray is usable for any of the subclasses as well.

The new method is added to the documentation.
The existing documentation for applyFromArray was alphabetized to make
it easier to follow.
One property (Style quotePrefix) was added to the documentation.
Some Borders pseudo-properties (vertical, horizontal, and outline) were
documented as usable by applyFromArray,
but aren't actually supported - they were removed.
The documentation of the properties seemed to use setProperty and
getProperty fairly randomly - it now uses setProperty exclusively.

New constants were added for the textRotation "angles" used to create a
"stacked" cell. I felt that changing the readers and writers to use
these constants was beyond the scope of this change, but it is
on my to-do list.

This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

New method requested in old issue. See above for details.


oleibman:
Removing stalebot tag.

PR #1585, opened 2020-07-18
Reconcile function discrepancies between the documented lists, and the Calculation Engine

MarkBaker:
This is:

```
- [ ] a bugfix
- [ ] a new feature
- [X] a documentation update
```

Checklist:

- [X] Code style is respected

Reconcile discrepancies between the documented lists of functions (by category and alphabetic), and the Calculation engine. Update them all to reflect the complete set of current standard Excel functions (including those introduced in "Office 365").

All existing Excel functions are now defined within the Calculation Engine code, even if tagged as "Not Yet Implemented". Where a new Excel function has replaced an older one, both are defined, and if this is simply an alias, then it is flagged as implemented if the old function had been implemented (e.g. `CHISQ.DIST.RT()` replacing `CHIDIST()`). If the replacement function has a different signature (e.g. `LOGNORMDIST()` has been replaced by `LOGNORM.DIST()`, but with a new optional argument), then the new function is treated as "Not Yet Implemented" until the change of functionality to handle the additional argument has been implemented. Completely new functions are simply flagged as "Not Yet Implemented".

In many cases, the new functions (or the behavioural changes) are easy to implement,but outside the scope of this PR

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 /analysis
 /vendor/
 /phpunit.xml
+/build
 
 ## IDE support
 *.buildpath
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 /analysis
 /vendor/
 /phpunit.xml
-/build
 
 ## IDE support
 *.buildpath
--- a/docs/references/function-list-by-category.md
+++ b/docs/references/function-list-by-category.md
@@ -4,13 +4,13 @@
 
 Excel Function      | Excel Version | PhpSpreadsheet Function
 --------------------|---------------|-------------------------------------------
-CUBEKPIMEMBER       |               | **Not yet Implemented**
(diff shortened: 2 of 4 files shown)


MarkBaker:
I'd forgotten about the function list documentation builder.

I don't particularly want to add more information to the Calculation function array than is necessary for use in the engine itself - I know it's static, but that array is still heavy on memory - and version is only of minor use for the documentation, so I'll drop it... I was also thinking about adding a link to the official MS documentation for each function, glad that I didn't now.

MarkBaker:
Removed the version column from the md files; they'll rebuild anyway, but at least the Calculation class definitions are fully up-to-date for them to build from

MarkBaker:
And I'll look at other ways that we might provide additional information such as version and perhaps links to the MS function docs in future

PR #1597, opened 2020-07-27
Drop PHP_CodeSniffer in favor of PHP-CS-Fixer

PowerKiKi:
This is:

```
- [ ] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Because PHP-CS-Fixer workflow for auto-fixing is much easier, and because
even though PHP-CS-Fixer might have more features we don't use use them, nor
need them in this particular project.

@MarkBaker I would like your opinion on this, because it kind of goes against one your recent change...


PowerKiKi:
[Mark said](https://gitter.im/PHPOffice/PhpSpreadsheet?at=5f1efe2fcd7bed0e3794aae0):

> Well I have personal reasons for preferring phpcs, which also has its own fixer (phpcbf) with its dry run mode, and modes of updating/patching files... the main developer lives about 5km from me, and i see her regularly, and know how much time and effort she puts into phpcs, especially now in the run-up to php8
And phpcs does also provide us with the very useful version check with phpcompatibility.

Issue #1616, opened 2020-08-16
`formatAsDate` does not strip language metadata


SheetJSDev:
[`formatAsDate`](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Style/NumberFormat.php#L472) attempts to strip language metadata but has a few issues.

The attached [chinesedate.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/5081214/chinesedate.xlsx) uses the format code `[$-zh-CN]dddd`.  The current regular expression does not strip it (it only matches one hyphen)

[From the XLSX spec](https://docs.microsoft.com/en-us/openspecs/office_standards/ms-oi29500/17d11129-219b-4e2c-88db-45844d21e528):

```
NFPartLocaleID =
    ASCII-LEFT-SQUARE-BRACKET
    ASCII-DOLLAR-SIGN
    1*UTF16-ANY
    [ASCII-HYPHEN-MINUS 3*8ASCII-DIGIT-HEXADECIMAL]
    ASCII-RIGHT-SQUARE-BRACKET
```

Ignoring the obvious issue (UTF16-ANY should be UTF16-ANY-WITHOUT-RIGHT-SQUARE-BRACKET), and noting that the `DBNum` specifier is of the form `\[DBNum\d\]`, the line probably should be something like:

```php
        $format = preg_replace('/^(\[DBNum\d\])*(\[\$[^\]]*\])/i', '', $format);
```

Note: This was originally left as a comment in [another issue](https://github.com/PHPOffice/PhpSpreadsheet/issues/140#issuecomment-554332866) since @cocowool 's contribution introduced the bug.

oleibman:
Fixed by PR #1618.

Issue #140, opened 2017-04-16
Change license from LGPL 2.1 to MIT ?


PowerKiKi:
From what I understand, notably from a [similar move from jQuery](http://blog.jquery.com/2012/09/10/jquery-licensing-changes/), it would technically be legal to change our license from LGPL to MIT. The reasons for such a move would to get rid of the boilerplate header in all files (which is not always consistent in our case), and make the end-user life easier by choosing a well-known, very permissive and broadly used license.

But I am not a lawyer, and I may be missing something. Is it really doable ? is there anything else to do than edit the files ? should we contact previous contributors ?

@MarkBaker, @maartenba as the two most prominent contributors, would you have any opinion on the matter ?

Also @Progi1984 what was the rationale for choosing LGPL 3 for PhpWord ? I see it was talked about in https://github.com/PHPOffice/PHPWord/issues/211, but would your mind has changed in the last 3 years ?

maartenba:
I for one am open to going for a more permissive license. Back when I started this thing ten years ago, LGPL was the license many chose, but since the world has changed a lot.

PowerKiKi:
Thanks for your input. Would you know anything about how we should do it ?

maartenba:
@MarkBaker @Progi1984 what do you think? Should we go to a different license?

lanthaler:
+1 for MIT

m8e:
+1 MIT, then I will be allowed to use PHPSpreadsheet at my daytime job :-)

PowerKiKi:
Mark confirmed via Twitter that he agreed to switch to MIT, so we can go ahead with this as soon as #147 is merged.

maartenba:
Nice!

Croydon:
jQuery's situation was different since they already did dual-licensing under GPL and MIT and they simply dropped GPL for any further versions.

PhpSpreadsheet is not dual-licensed and you want to change the license in a not compatible way. More permissive licenses are after compatible with more restrictive licenses, but not the other way around.

You either need to ask every single contributor to relicense their contribution or you need to remove their contribution (replacing it with something new) or you can't change from GPL to MIT without violating the rights of the contributors.

You could start adding a checkbox to your pull request template asking new contributors to dual-license under LGPL and MIT, then starting to contact previous contributors and/or rewriting smaller contributions.

Once done you can drop LGPL and move to MIT.

PowerKiKi:
[...]

PR #1618, opened 2020-08-18
formatAsDate strip language metadata

SheetJSDev:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

For more context:
- https://github.com/PHPOffice/PhpSpreadsheet/issues/1616
- https://github.com/PHPOffice/PhpSpreadsheet/issues/140#issuecomment-554332866

The first commit reverts the original commit by @cocowool .  This is because both the code and the tests were incorrect (and doing so removes a bit of license uncertainty around the MIT relicensing).

The linked issues explain some of the issues with the implementation

The original tests were flawed because they did not test LCID (I added a test using `[$-zh-CN]`) and they used DBNum1.  For `zh-CN`, `DBNum4` uses standard ascii digits and those generally should be preferred since they will be correct if Chinese support is improved.


MarkBaker:
Thank you for correcting this

Issue #1592, opened 2020-07-22
Uninitialized string offset in Xls.php


MatthiasKuehneEllerhold:
This is:
```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Correct parsing of an XLS file. No errors / notices thrown. No running into the execution timeout.

### What is the current behavior?
```
NOTICE (5): ErrorException: Uninitialized string offset: 2386 in /x/y/Logger.php:86
   1701 Stack trace:
   1702 #0 /x/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xls.php(7886): \Logger::Log\{closure}(8, 'Uninitialized s...', '/var/...', 7886, Array)
   1703 #1 /x/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xls.php(2981): PhpOffice\PhpSpreadsheet\Reader\Xls::getUInt2d('9\x02\x00\x009\x02\x00\x00\x12\x00\x00Plak...', 2386)
   1704 #2 /x/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xls.php(727): PhpOffice\PhpSpreadsheet\Reader\Xls->readSst()
   1705 #3 /x/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/IOFactory.php(86): PhpOffice\PhpSpreadsheet\Reader\Xls->load('/var/www/ellerh...')
   1706 #4 /Parser.php(32): PhpOffice\PhpSpreadsheet\IOFactory::load('/var/...')
```
(Same happened for offset 2387.)
The read took forever and was stopped after 30 seconds because of an execution timeout.

### What are the steps to reproduce?
This error occured when an XLS file was read. I cant make the spreadsheet public, because it contains business data.
This is the first spreadsheet that exhibit this behaviour, all other spreadsheets from this source were correctly read & parsed.

We've got a hold of the XLS file. The process was that a generated XLS file was edited by hand in Office 365 and saved. The original generated file could be parsed without any problem, the edited file lead to this error.
Reopening the edited file in Excel 2016 and saving it (without any modification) fixed it and the parsing worked flawlessly.

### Which versions of PhpSpreadsheet and PHP are affected?
[...]

PR #1623, opened 2020-08-21
Fix for bug #1592 (UPDATED)

ggiovinazzo:
This is:

```
- [X ] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

Original PR Text:
Some Xls file have bad SST length in BIFF8 record (FCh). When reading those files, if no further control is applied, the library launches a notice when trying to access out of bound string index, and in some cases when notices are disabled, this can result in an infinite loop.

In this PR there is also a test with a modified excel file that indicates a count of 15 shared strings when the real count is of 12.
Notice: This is not an error for MS Office or LibreOffice that can open the file without any problem

Update:
There was a bug when calculating SST total length causing that some files with very large SST fail to load all data. This bug, introduced with previous commit, has been solved.


MarkBaker:
Is it possible to separate the Xml namespace change into a separate PR from the Xls bugfix?

ggiovinazzo:
I Will modify my main branch ASAP. This additional commits are a side effect of automated GitHub build, test and or that I wasn't aware of.

ggiovinazzo:
PR already splitted. I submitted another PR (#1756) With xml namespace fixes.

PR #1627, opened 2020-08-25
Fix bug #1626 where values of 0 were "rounded" up/down as if they were not 0

### LINKED ISSUE #1626: ROUNDDOWN and ROUNDUP return incorrect results for values of 0
This is:

```
- [X] a bug report
- [ ] a feature request
```

Values of 0 should stay 0 when either rounded up or down.

Values of 0 are "rounded" down/up as if they were not 0.

The following code produces output `-0.01 0.01` instead of `0 0`:
```php
<?php

echo \PhpOffice\PhpSpreadsheet\Calculation\MathTrig::ROUNDDOWN(0, 2).' ';
echo \PhpOffice\PhpSpreadsheet\Calculation\MathTrig::ROUNDUP(0, 2).' ';
```

### What could be a possible solution?
A possible solution could be to explicitly handle values of zero. In `ROUNDUP`, for example, the code could be something like this:
```php
        if ((is_numeric($number)) && (is_numeric($digits))) {
            if ($number == 0.0) {
                return 0.0;
            }

            if ($number < 0.0) {
                return round($number - 0.5 * 0.1 ** $digits, $digits, PHP_ROUND_HALF_DOWN);
            }

            return round($number + 0.5 * 0.1 ** $digits, $digits, PHP_ROUND_HALF_DOWN);
        }
```

PhpSpreadsheet 1.14.1
PHP 7.2.20

Flinsch:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Fixes #1626 

Checklist:

- [X] Code style is respected
- [X] Documentation is updated as necessary (not necessary)

Values of 0 should stay 0 when either rounded up or down.


Flinsch:
#1626

Flinsch:
Sorry for two commits with identical comments.

MarkBaker:
Thank you for catching the error, and for the fix

Watch when you're updating the change log though, that you tag your change in the `## Unreleased - TBD` section rather than against a previous release

Flinsch:
> Thank you for catching the error, and for the fix
> 
> Watch when you're updating the change log though, that you tag your change in the `## Unreleased - TBD` section rather than against a previous release

Oh yes, sorry for the confusion. And thank you very much too!

PR #1629, opened 2020-08-25
Pdf Writer strtoupper() fix

adancabanas:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

PhpSpreadsheet/Writer/Pdf.php Class defines a protected static mixed array called $paperSizes, this array contains string values along with array values. 'strtoupper() expects parameter 1 to be string, array given' error happens due to array passed to $paperSize variable from that $paperSizes mixed array on the Mpdf Class where Pdf extends

Examples of cases, when a 'Letter' paper size is chosen, then no problem occurs since the index in the array is a string value, but when 'Tabloid' paper size is chosen the value in the index for that paper size is an array, that's when the strtoupper() error happens.


MarkBaker:
Thank you for identifying and fixing this issue

PR #1661, opened 2020-10-01
Calculation/DateTime Failure With PHP8

oleibman:
The following code generates an error with PHP8:
  if ($t1[1] > 29) {
    $t1[1] += 1900;

Under the "right" conditions, PHP8 evaluates the condition as true
when PHP8 is a non-numeric string and then generates an error trying
to perform += on that non-numeric string. Adding a numeric test
eliminates the problem. All unit tests involving this code now
succeed with both PHP7 and PHP8.

This is:

```
- [ ] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Calculation/DateTime.php
+++ b/src/PhpSpreadsheet/Calculation/DateTime.php
@@ -487,7 +487,7 @@ class DateTime
             if ($yearFound) {
                 array_unshift($t1, 1);
             } else {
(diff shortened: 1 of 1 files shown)


MarkBaker:
I've tried to get a  PHP8 pipeline running for testing purposes, though there's still a few problems with it; but I had noticed this particular error in the tests. Thanks for the resolution

PR #1664, opened 2020-10-01
Fix XLSX reader when having a corrupt numeric cell data type

brainfoolong:
This is:

```
- [x] a bugfix
```

Checklist:

- [x] Code style is respected

In some rare cases, when reading an XLSX file that may contain invalid cell type values, the reader throw an error. It happens when the found data type in a table cell is numeric, but the actual value is an empty string. This will throw a exception and the result is the file cannot be read.

You can open those files in any reader (LibreOffice, MS Office) without problems but PhpSpreadsheet struggles with it.

Unfortunetely i cannot add tests or a demonstration file because our files that cause this error contains sensible data. If we try to remove the sensible data, the file is automatically fixed by our Office Suite. The file we use is generated from a third party company, so we have no chance to generate files with those errors.

Hopefully this PR will be merged even if no test scenario is provided.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,7 +35,8 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 - Improve Coverage for ODS Reader [#1545](https://github.com/phpoffice/phpspreadsheet/pull/1545)
 - Named formula implementation, and improved handling of Defined Names generally [#1535](https://github.com/PHPOffice/PhpSpreadsheet/pull/1535)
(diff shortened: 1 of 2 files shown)


brainfoolong:
Ping, before it get closed for inactivity. Some maintainer here?

MarkBaker:
Thanks for highlighting this. It always surprises me how much of the Reader code is devoted to handling malformed content or corrupted/invalid files.

I've not been able to recreate a test workbook to demonstrate the problem either; but I can't see the patch causing any problems.

PR #1671, opened 2020-10-07
Update Coordinate::coordinateFromString

jszczypk:
Coordinate::coordinateFromString should return array [ string, int ] as row number is expected to be int in all other functions like Worksheet:getRowDimension(int $row)

This is:

```
- [*] a bugfix
- [ ] a new feature
```

Checklist:

- [?] Changes are covered by unit tests
- [*] Code style is respected


MarkBaker:
can you take a look at the failing tests that this creates, see what affect that it has on the reference helper?

oleibman:
`Coordinate::indexesFromString` returns suggested type.

Issue #1654, opened 2020-09-18
Make methods protected instead of private in \PhpOffice\PhpSpreadsheet\Reader\Html


johnluxor:
Hello

I need to extend the file `\PhpOffice\PhpSpreadsheet\Reader\Html` but I cannot override some methods because they are private. But those methods(listed below) use for format data, and it should be available to extend to change the formatting

Methods:
 - insertImage
 - setBorderStyle
 - applyInlineStyle

Maybe others, but I use those

```
- [*] a feature request
```

What do you think about it?

oleibman:
Removing stale label. This issue has been rejected in the past. If anyone is still following this thread, I would be interested in knowing your use cases. I can then evaluate whether to continue to do nothing about, or to integrate your needs into the project, or to change the visibility.

oleibman:
Three months, no comments. Not surprising, but I guess that means the decision is up to me. I am going to close this issue for reasons to be discussed below.

2 of the 3 routines requested are fairly large, and fairly volatile ( `setBorderStyle` is relatively small and stable). `insertImage` is 55 lines. It has had 7 changes in the last year; it would be difficult for users who have overridden it to keep up. Maybe they don't care about, say, opacity; but, in all probability, they either aren't aware of it (and are losing out because of that), or their override is because of it (and so is no longer needed). Regardless, they can't really abstain from the security patches which were needed here. `applyInlineStyle` is almost 200 lines long. It has had 5 changes in the last year (none security-related), so, again, it is difficult for overriding users to keep up.

I think any issues with these or other routines which might benefit from being overrideable are better off being dealt with as issues whose solution would benefit the entire community. I am not completely opposed to narrowly-targeted protected routines; that is certainly a possible outcome of such future issues.

PR #1682, opened 2020-10-15
Change PhpOffice\PhpSpreadsheet\Reader\Html.php methods visibility

nalatreb:
This is:

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [X] Code style is respected

I made this changes for issue [1654](https://github.com/PHPOffice/PhpSpreadsheet/issues/1654)


PowerKiKi:
Thank you for your contribution. However changing visibility from private to protected opens up the API surface en prevent us to make changes that could then break user code. This is not something we are keen to do unless there is a very strong case for it. I'm afraid this won't be merged.

PR #1687, opened 2020-10-24
Fix uncaught exception when loading files with strange range defs [#1…

### LINKED ISSUE #1686: XLSX reader - Exception : You must specify a worksheet or a scope for a Named Range
`This is a bug report`

Read a simple XLSX file (no formula inside) without exception

The XLSX reader crashes with the message

```
PhpOffice\PhpSpreadsheet\Exception : You must specify a worksheet or a scope for a Named Range
 /home/vagrant/code/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/NamedRange.php:21
 /home/vagrant/code/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/DefinedName.php:97
 /home/vagrant/code/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xlsx.php:1398
```

```php
// This works properly, file is open
$reader = new Xls();
$reader->load(storage_path('default.xls'));

// This fails to open with the above exception
$reader = new Xlsx();
$reader->load(storage_path('default.xlsx'));
```

[test.zip](https://github.com/PHPOffice/PhpSpreadsheet/files/5417931/test.zip)

The above zip file contains the XLS and XLSX files that exhibit the behaviour. The XLS file is simply the XLSX file saved as XLS (no change in the spreadsheet itself).

Package 1.15.0
PHP 7.4

pittyplatsch:
Fixes #1686 

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

To make one fellow library user happy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 - Ensure that the list of shared formulae is maintained when an xlsx file is chunked with readFilter[Issue #169](https://github.com/PHPOffice/PhpSpreadsheet/issues/1669).
 - Fix for notice during accessing "cached magnification factor" offset [#1354](https://github.com/PHPOffice/PhpSpreadsheet/pull/1354)
+- Fix uncaught exception when loading files with strange range defs [#1686]
 
 ## 1.15.0 - 2020-10-11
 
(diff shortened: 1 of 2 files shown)


JbDumaine (approved):

MarkBaker:
Do you have an example of what classifies as a 'strange range def'; it might be an edge-case that we can handle separately without just discarding

oleibman:
Closing. No update in over 2 years. The sample file in issue 1686 now loads without exception. It was very likely fixed by PR #1742 very close to when the last comment was entered on this ticket.

PR #1694, opened 2020-10-30
Final stage of preparation for PHP8

MarkBaker:
Starting work on ensuring that all methods are properly type-hinted (arguments and returns) and that all argument names for public methods are meaningful (in readiness for PHP8 named arguments)

This is:

```
- [ ] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

Final stage of preparation for PHP8


PowerKiKi:
I "finished" this PR by fixing CI and merging it, because it was already super old, and it would only get harder to merge it if we waited longer. But there are still parameters with non-optimal names that could be fixed in future PR

PR #1710, opened 2020-11-11
Fixed reading XSLS style alignments from XML

addiks:
These changes remove the redundant child-node access in the method `readAlignmentStyle` which would otherwise lead to errors.

This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

The attribute `$alignmentXml` given to the private method `readAlignmentStyle` is the alignment XML tag that contains the alignment data itself. But inside that method all data are read from another `alignment` XML tag within that given tag. This redundant child-node access resulted in the following error- / notice-message: `PHP Notice:  Trying to access array offset on value of type null in /foo/bar/baz/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xlsx/Styles.php on line 146`.


MarkBaker:
Thank you for identifying and resolving this problem

PR #1716, opened 2020-11-17
Stacked Alignment - Use Class Constant Rather than Literal

oleibman:
This is:

```
- [x] a bugfix (maintainability and coverage)
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

PR #1580 defined constants for "stacked" alignment in cells.
Using those constants outside of Style/Alignment was beyond the
scope of the original PR, but I said I would get to it.
This PR replaces all uses of literal -165, and appropriate uses of
literal 255, with the named constants, and adds tests to make sure
that the changed code is covered in the test suite.


src/PhpSpreadsheet/Writer/Xls/Xf.php
```diff
@@ -514,15 +499,11 @@ private static function mapTextRotation($textRotation)
...
-                return 1;
-            case Protection::PROTECTION_PROTECTED:
-                return 1;
             case Protection::PROTECTION_UNPROTECTED:
                 return 0;
-            default:
-                return 1;
         }
+
+        return 1;
```
MarkBaker, line 506: It may be slightly less performant checking all cases; but I prefer their explicit use for readability.

src/PhpSpreadsheet/Writer/Xls/Xf.php
```diff
@@ -535,14 +516,10 @@ private static function mapLocked($locked)
...
-                return 0;
             case Protection::PROTECTION_PROTECTED:
                 return 1;
-            case Protection::PROTECTION_UNPROTECTED:
-                return 0;
-            default:
-                return 0;
         }
+
+        return 0;
```
MarkBaker, line 523: As comment above for `mapLocked()`

src/PhpSpreadsheet/Reader/Xls.php
```diff
@@ -2286,8 +2286,8 @@ private function readXf(): void
                     $rotation = $angle;
                 } elseif ($angle <= 180) {
                     $rotation = 90 - $angle;
-                } elseif ($angle == 255) {
-                    $rotation = -165;
+                } elseif ($angle == Alignment::TEXTROTATION_STACK_EXCEL) {
+                    $rotation = Alignment::TEXTROTATION_STACK_PHPSPREADSHEET;
```
MarkBaker, line 2290: Switching to using named constants wherever appropriate is definitely a good idea going forward

MarkBaker (changes requested):

src/PhpSpreadsheet/Writer/Xls/Xf.php
oleibman: I had coded it the way I did not particularly for performance reasons, but rather for coverage reasons - there can sometimes be (and would in this case) an uncovered return statement (or even an uncovered fall out from the switch) if all possibilities are covered in the switch. I believe the way I have recoded it takes care of your readability requirement without leaving anything uncovered.

src/PhpSpreadsheet/Writer/Xls/Xf.php
oleibman: Likewise - see resolution above for mapLocked.

src/PhpSpreadsheet/Writer/Xls/Xf.php
```diff
@@ -515,11 +522,6 @@ private static function mapLocked($locked)
...
     private static function mapHidden($hidden)
     {
-        switch ($hidden) {
-            case Protection::PROTECTION_PROTECTED:
-                return 1;
-        }
-
-        return 0;
+        return array_key_exists($hidden, self::HIDDEN_ARRAY) ? HIDDEN_ARRAY[$hidden] : 0;
     }
```
MarkBaker, line 526: `self::HIDDEN_ARRAY`

But yes, I think the map makes it explicit for readability

MarkBaker (changes requested):

src/PhpSpreadsheet/Writer/Xls/Xf.php
```diff
@@ -498,14 +504,15 @@ private static function mapTextRotation($textRotation)
      */
     private static function mapLocked($locked)
     {
-        switch ($locked) {
-            case Protection::PROTECTION_UNPROTECTED:
-                return 0;
-        }
-
-        return 1;
+        return array_key_exists($locked, self::LOCK_ARRAY) ? LOCK_ARRAY[$locked] : 1;
```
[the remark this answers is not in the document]
oleibman: I'm not sure what happened here. Commit dc4f568 had the error you are asking to correct, but I replaced it immediately with commit dc908f5, which has the correction. There would have been a very short window indeed for you to see the wrong code. If you look at lines 507 and 525, they have the appropriate change. So I am marking this as resolved without making any further changes. If I am misinterpreting something, please let me know.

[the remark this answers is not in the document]
oleibman: Actually, I decided to change a comment just in case. Commit 9c279fb. If the code still looks wrong to you, I will cancel this PR and open another in its place.

src/PhpSpreadsheet/Writer/Xls/Xf.php
oleibman: See above regarding erroneous commit dc4f568 replaced by correct dc908f5 and 9c279fb.

src/PhpSpreadsheet/Writer/Xls/Xf.php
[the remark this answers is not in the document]
oleibman: Actually, I decided to add a comment just in case. Commit 9c279fb. If the code still looks wrong to you, I will cancel this PR and open another in its place.

MarkBaker:
Code is looking good now, thanks

I know in my more recent code, I'm making a lot more use of class constants, I definitely think it's a much more positive approach than literals, especially when the constants are descriptive in their names

PR #1719, opened 2020-11-18
Resolve XSS Vulnerability in the HTML Writer

MarkBaker:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

To prevent a potential XSS attack when loading a spreadsheet from an unknown source and saving as HTML


pandeydhirendra:
Hi Team,

We are using this library with version (1.10.0) in one of our projects. Request you to please release this fix in a stable version.
Thanks in advance.

shahariaazam:
Can we ship this PR to older branch like `1.8.x` and release a patch? Because 1.8.2 support PHP 5.6 and 7.0. There are lots of application that are still using PHP7.0 and that requires this fixing.

So what about we can apply this to 1.8.2 and create a new release for 1.8.3? For me, it's kind of important. I already applied this patch on 1.8.2. Check https://github.com/PHPOffice/PhpSpreadsheet/compare/1.8.2...shahariaazam:xss-fix-for-1.8.2?expand=1

j4nr6n:
Same question but for 1.12, which is the latest release that supports PHP 7.1.

deepti4666:
I have this issue for 1.15.0 

image

Issue #1712, opened 2020-11-15
When selecting a whole column, style is not applied to cells already containing a value


pdesgarets:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

Style can be set on a whole column, regardless of wether or not the column contains data.

### What is the current behavior?

Style can be set on a whole column, only for the cells not having data.

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

use PhpOffice\PhpSpreadsheet\Spreadsheet;
use PhpOffice\PhpSpreadsheet\Style\Fill;
use PhpOffice\PhpSpreadsheet\Writer\Xlsx;

$rgb = '4467b8';

// 1. Set values before style

$spreadsheet = new Spreadsheet();

$spreadsheet->getActiveSheet()->fromArray(['OK', 'KO']);


// "CELL" style is always OK
$spreadsheet->getActiveSheet()->getStyle('A1')->getFill()
    ->setFillType(Fill::FILL_SOLID)
    ->getStartColor()->setRGB($rgb);;

// "COLUMN" style is KO because value is already set
$spreadsheet->getActiveSheet()->getStyle('B')->getFill()
    ->setFillType(Fill::FILL_SOLID)
    ->getStartColor()->setRGB($rgb);

var_dump($spreadsheet->getActiveSheet()->getCell('A1')->getStyle()->getFill()->getStartColor()->getRGB());
// string(6) "4467b8"
var_dump($spreadsheet->getActiveSheet()->getCell('B1')->getStyle()->getFill()->getStartColor()->getRGB());
// string(6) "FFFFFF"

$xlsx = new Xlsx($spreadsheet);
$xlsx->save('/tmp/ko.xlsx');

// 2. Set values after style
$spreadsheet = new Spreadsheet();

// "CELL" style is always OK
$spreadsheet->getActiveSheet()->getStyle('A1')->getFill()
    ->setFillType(Fill::FILL_SOLID)
    ->getStartColor()->setRGB($rgb);;

// "COLUMN" style is OK because value is not set yet
[...]

PR #1721, opened 2020-11-20
Apply Column and Row Styles to Existing Cells

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

This is a fix for issue #1712.
When a style is applied to an entire row or column, it is currently
only effective for cells which don't already contain a value.
The code needs to iterate through existing cells in the row/column
in order to apply the style to them.
This could be considered a breaking change, however, I believe that
the change makes things operate as users would expect, and that the
existing implementation is incomplete.

The change also removes protected element conditionalStyles from
the Style class. That element is an unused remnant, and can no longer be
set or retrieved - methods getConditionalStyles and setConditionalStyles
actually act on an element in the Worksheet class.

Finally, additional tests are added so that Style, and in fact the
entire Style directory, now has 100% test coverage.


oleibman:
There seems to be a problem with PHP8 tests on Web server - hangs after several thousand tests.  I don't have a problem with PHP8 on my own machine. So, converting to draft for now. Will open a new ticket to see if I can pinpoint the problem.

oleibman:
Problem appears to be with PHP8 environment for PR. Opened issue #1725. Restoring status of this ticket from draft to "ready to review".

MarkBaker:
The switch from travis CI to github actions may have a few teething issues, but was certainly necessary after travis dropped their OSS support earlier this week.

Thanks for all the work that you're contributing to the project; you're making an incredibly difference. I know you've done a lot with the calculation engine and formula as well, are you looking forward to implementing the new LAMBDA() and LET() functions in Excel as much as I am (not)?

oleibman:
Glad to be of help. The LET and LAMBDA functions look fascinating. Alas, they seem too complicated for me to figure out how to implement them. I'll look forward to seeing how you do it.

PR #1727, opened 2020-11-23
worksheet: fix if cellValue does not exist

jasverix:
The condition is FALSE if the cell does not exist in the flipped table, but anyway, it is sent in to a method requiring 'string' type, causing it to fail.

This is:

```
- [ x ] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

I have a big export through Excel, and it fails somewhere because if this error. I see that the condition allows it to be not set, but then it is sending 'null' and null is not allowed in that method.


jasverix:
Only a minor fix. Please look at this, and see if it is a correct fix, because it is blocking me right now. Looks like the failing unit-tests is just error in pipeline (error in downloading from github). Probably method should rather accept `?string` instead of running `?? ''`?

MarkBaker:
Thanks for the contribution

PR #1728, opened 2020-11-23
make voku/anti-xss an optional package

jasverix:
This package installs voku/portable-utf8 which "sanitizes" REQUEST_URI in globals, which in turn crashed our entire system. 

Library packages should never 'do' anything, just be libraries.

This is:

```
- [ x ] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Our system did break down some time ago after installing voku/portable-utf8 because it sanitizes global variable REQUEST_URI. After that, the package is completely banned from our code base. I don't know if you will accept making it optional (only used one place..), but I will give it a try.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,8 +58,7 @@
         "markbaker/matrix": "^1.2||^2.0",
         "psr/simple-cache": "^1.0",
         "psr/http-client": "^1.0",
-        "psr/http-factory": "^1.0",
-        "voku/anti-xss": "^4.1"
+        "psr/http-factory": "^1.0"
     },
     "require-dev": {
         "dompdf/dompdf": "^0.8.5",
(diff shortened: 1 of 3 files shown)


jasverix:
Yes, I understand, and I just wanted to open discussion, because I did not think you would just remove it. But voku/portable-utf8 has a bootstrap.php file which does a little more than we can accept - libraries should in my world just add classes and not do things on each request.

I would not have any problems using htmlpurifier, that library is not adding any bootstrap file at least (and has much less dependencies). I can fix the PR to do that instead, and then you can consider. I am not using the HTML Writer myself.

jasverix:
Tried to implement this now, but you have unit-tests requiring that this string is not allowed:
```<div class="comment">javascript:alert(1)</div>```

That's not dangerous HTML, it's just strange plain text. As an attribute, though, it is more dangerous. But should we still not allow this text? Looks like anti-xss removed it completely, but HTMLPurifier leaves it (but the first test `<script>alert('test')</script>` is passing)

MarkBaker:
I've updated the tests, and modified the code to use HTMLPurifier instead of anti-xss in a [separate PR](https://github.com/PHPOffice/PhpSpreadsheet/pull/1751)

jasverix:
Great! Thank you!

PR #1733, opened 2020-11-26
Fix pixelsToPoints conversion ratio (for HTML/CSS & PDF col width)

guilliamxavier:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

When converting an XLSX to PDF, the resulting columns widths are smaller.

"0.67777777" is wrong, the correct pt/px ratio is 72/96 = 3/4 = 0.75

Refs https://github.com/PHPOffice/PHPWord/pull/234
Replaces https://github.com/PHPOffice/PHPExcel/pull/497

Additionally, I updated `* 1.333333333` to `/ 0.75` in `pointsToPixels` (rather than updating `* 0.75` to `/ 1.333333333` in `pixelsToPoints`) for symmetry/consistency (and because 3/4 has an exact representation in both decimal and binary, while 4/3 hasn't)

### DIFF
--- a/src/PhpSpreadsheet/Shared/Drawing.php
+++ b/src/PhpSpreadsheet/Shared/Drawing.php
@@ -98,7 +98,7 @@ class Drawing
      */
     public static function pixelsToPoints($pValue)
     {
-        return $pValue * 0.67777777;
+        return $pValue * 0.75;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


MarkBaker:
Thank you for correcting this error

Issue #1730, opened 2020-11-23
Error/Exception using reader since upgrading to 1.15 - "Tou must specify a Formula value for a Named Formula"


mtindall89:
This is:

```
- [x ] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Loading a spreadsheet via reader

### What is the current behavior?
PHP Exception

### What are the steps to reproduce?
Since upgrading to v 1.15 from v1.5, the following exception happens when trying to load files:
"Tou must specify a Formula value for a Named Formula"

This is traced to NamedFormula.php lines 20-22.

Very similar to Issue #1723 

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$reader = new \PhpOffice\PhpSpreadsheet\Reader\Xlsx();

$reader->setReadDataOnly(true);
$spreadsheet = $reader->load($filePath);
```

### Which versions of PhpSpreadsheet and PHP are affected?
PHPSpreadsheet - 1.15
PHP v 7.4.12

mtindall89:
Can anyone address this? I can provide an excel sheet example if needed. I think it may be any excel sheet that has formulas in it.

syslord2k2:
I got the same error after upgrading.

sussh:
Here are my logs of a similar error.
`` array(2) {
  'error' =>
  string(60) "Alsma:: You must specify a Formula value for a Named Formula"
  'traces' =>
  array(9) {
    [0] =>
    array(6) {
      'file' =>
      string(84) "/var/www/synergy-domains/pp55/api/application/classes/PhpSpreadsheet/DefinedName.php"
      'line' =>
      int(94)
      'function' =>
      string(11) "__construct"
      'class' =>
      string(37) "PhpOffice\PhpSpreadsheet\NamedFormula"
      'type' =>
      string(2) "->"
      'args' =>
      array(5) {
        [0] =>
        string(7) "HTML_OS"
        [1] =>
        NULL
        [2] =>
        string(1) "0"
        [3] =>
        bool(false)
        [4] =>
        NULL
      }
    }
[...]

Issue #1723, opened 2020-11-21
Exception: You must specify a worksheet or a scope for a Named Range


DammWare:
This is a bug report

### What is the expected behavior?
Read an xlsx file of approximately 47000 rows.
[sat.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/5576831/sat.xlsx)

### What is the current behavior?
error 1
![image](https://user-images.githubusercontent.com/30880547/99864588-d2047500-2b69-11eb-8d7d-31300a5042fd.png)
error 2
![image](https://user-images.githubusercontent.com/30880547/99864687-7a1a3e00-2b6a-11eb-9967-b7a91fc4bc45.png)

### What are the steps to reproduce?
```php
<?php
# error 1
require_once("vendor/autoload.php");
use PhpOffice\PhpSpreadsheet\Reader\Xlsx;
$reader = new Xlsx();
$reader->load("sat.xlsx");
?>
```

```php
<?php
# error 2
require_once("vendor/autoload.php");
$file_type = null;
$reader = null;
$spreadsheet = null;
$data = null;
$rutaArchivo = "sat.xlsx";
$libro = 0;
$file_type = PhpOffice\PhpSpreadsheet\IOFactory::identify($rutaArchivo);
$reader = PhpOffice\PhpSpreadsheet\IOFactory::createReader($file_type);
$isReadDataOnly = $reader -> getreadDataOnly();
$spreadsheet = $reader->load($rutaArchivo);
$data = $spreadsheet->getActiveSheet($libro)->toArray();
var_dump( count($data) );
die();
?>
```

### Which versions of PhpSpreadsheet and PHP are affected?
PhpSpreadsheet  1.15.0.0
PHP 7.4.11

PR #1742, opened 2020-12-06
Fix for 3 Issues Involving ReadXlsx and NamedRange

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Issues #1686 and #1723, which provide sample spreadsheets, are probably
solved by this ticket. Issue #1730 is also probably solved, but I have
no way to verify.

There are two problems with how PhpSpreadsheet is handling things now.
Although the first problem is much less severe, and isn't really a factor
in the issues named above, it is helpful to get it out of the way first.
If you define a named range in Excel, and then delete the sheet where
the range exists, Excel saves the range as #REF!. If there is a cell which
references the range, it will similarly have the value #REF! when you open
the Excel file.
Currently, PhpSpreadsheet discards the #REF! definition, so a cell which
references the range will appear as #NAME? rather than #REF!.
This PR changes the behavior so that PhpSpreadsheet retains the #REF!
definition, and cells which reference it will appear as #REF!.

The second problem is the more severe, and is, I believe, responsible
for the 3 issues identified above.
If you define a named range and the sheet on which the range is defined
does not exist at the time, Excel will save the range as something like:

'[1]Unknown Sheet'!$A$1

If a cell references such a range, Excel will again display #REF!.
PhpSpreadsheet currently throws an Exception when it encounters
such a definition while reading the file. This PR changes
the behavior so that PhpSpreadsheet saves the definition as #REF!,
and cells which reference it will behave similarly.

For the record, I will note that Excel does not magically recalculate when a
missing sheet is subsequently added, despite the fact that the reference
might now become resolvable. PhpSpreadsheet behaves likewise.


MarkBaker:
Good catch

PR #1746, opened 2020-12-07
WIP: Skip adding a defined name if the sheet isn't found

vhr:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

in case when in the xlsx file has unused range name and the library will throw a exception with the following message as below, every time when we try to read the file
> You must specify a worksheet or a scope for a Named Range


oleibman:
I believe that PR #1742 accomplishes the same goal, in a manner consistent with how Excel operates.

PR #1698, opened 2020-11-02
Xmltest3

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Changes are covered by unit tests
- [x] Code style is respected
- [x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ ] CHANGELOG.md contains a short summary of the change
- [ ] Documentation is updated as necessary

### Why this change is needed?
I omitted one test from my PR to improve test coverage for XML Read,
because Travis reported an error that I could not duplicate on my system.
The likeliest explanation seemed to be that there were commits to master
which were not present in my fork. Whether or not that is true,
I synced my fork with the master, and duplicated and fixed the problem.

PR #1756, opened 2020-12-14
Xml fix namespaces

ggiovinazzo:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

When reading an Xml/Excel file with namespace defined in header tag and never used (ie: xmlns:x)
The reader fails to load xml file; It tries to access 'x' namespace index but the namespace was never read.


oleibman:
When I run your testNoNameSpaceError script against the current PhpSpreadsheet code, it runs to completion and all the cells have their expected value. Presumably you see an error or a bad value when you try this. Based on the date of your PR (2020-12-14), I think your code base might not have the changes from PR #1698 (merged 2020-12-17), which might (I'm not positive) also have addressed this problem in a different way. Do you still see a problem when you run against the current code, and, if so, what is the problem that you see?

ggiovinazzo:
You're right. The bug has beeen fixed with your commits. I'll close this PR. Thanks.

PR #1760, opened 2020-12-20
Fix/chart axis titles

hazybluedot:
…axisLabel or YaxisLabel

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

For some Xlsx files when only a horizontal axis title was given it would map to a YAxisTitle. The reader was treating varAx as a title for only the YAxisTitle, but at least for scatter charts in Excel 2019, varAx is used for both the horizontal and vertical axis labels. This fix uses the axPos element to determine axis position, bottom and top are mapped to XAxisLabel, while left and right are mapped to YAxisLabel.


MarkBaker:
Thank you for this resolution to problems with single-axis charts

PR #1761, opened 2020-12-20
Fix/sheets xlsx chart

hazybluedot:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

When exporting Google Sheets as Xlsx, charts are positioned using oneCellAnchor which previously had not been implemented. In additional exported Xlsx files do not contain the *Cache element. Since *Chache elements had been assumed to be children of numRef and strRef elements, "Node has been deleted" errors were thrown when attempting to read an Xlsx exported from Google Sheets.


MarkBaker:
Thank you for providing this resolution to one-cell-anchor charts, and the issue with cache elements

PR #1762, opened 2020-12-21
Fix compatibility with ext-gd on php 8

derrabus:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

On PHP 8, the GD extension migrated away from the `resource` type and uses objects instead. This PR makes the necessary adjustments.


MarkBaker:
Thank you for that fix: we missed it when we were upgrading the code to support PHP8. Pity that it's an annoying break for scrutinizer

PR #1769, opened 2020-12-29
Handle absolute path in worksheets Target

tiagof:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected
- [n.a.] Documentation is updated as necessary

Some(*) xlsx files have the target relationship value as an absolute path which breaks the code. This aims at validating that scenario and adjusting the worksheet path calculation accordingly.

(*) To be more clear, a client is providing an xlsx file where these paths absolute. I have no idea how they generate these nor do I know how to create one myself. Nevertheless, this is a real life case.

Additionally I'd like to ask for help from the community in order to add the relevant test for his scenario since, as I said, I don't even know how to create a such a file.

Let me know any comments!


oleibman:
I have created a file which uses absolute paths everywhere that did not cause a problem for Excel when it tried to open it. I will upload it to this ticket.

As for your change, I am unable to see a problem with the existing code that you are trying to change in "load". The getFromZipArchive method checks for the condition you are trying to address, so your change appears to be superfluous. It certainly is for my file - PhpSpreadsheet loads it without a problem. If you have a file where you think your change to "load" is needed, I would have to see the file in question; or, if you cannot load this file successfully, I would need to compare your code to mine.

I agree that your change to listWorksheetInfo corrects a problem. Please use the uploaded file to create a test case which can be included with your PR.
[pr1769e.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/6330322/pr1769e.xlsx)

tiagof:
Hi. You're right about the change in `load`: it is unnecessary.
I've reverted it and added the test.

oleibman:
Can you revert your change to CHANGELOG.md in order to move your change forward? We can apply it afterwards.

tiagof:
Done (added it to Fixes)

PR #1775, opened 2021-01-04
Fix Four Tests for 2021

oleibman:
Four date tests omit year and therefore default to current year,
which means that the test results must be updated yearly.

This is:

```
- [x] a bugfix (testing only)
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Required annual change to confirm that PhpSpreadsheet uses current year when year is omitted in certain functions.


oleibman:
Not sure what Scrutinizer problem is. Doesn't really matter - #1774 got here first, so closing this request in favor of that one.

PR #1778, opened 2021-01-05
Do not swallow previous exception when loading content as DOM Document fails

JaZo:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

I didn't add tests to cover this change as there isn't something like `expectPreviousException` in PHPUnit. However, I'm open for suggestions to cover this with tests!

The previous exception will be included when loading the content as DOM Document fails. This makes debugging the reason behind the failure much easier.

N.B. I'm using this library with Laravel which converts all warnings/errors to an ErrorException so the error isn't reported but thrown.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Html.php
+++ b/src/PhpSpreadsheet/Reader/Html.php
@@ -650,7 +650,7 @@ class Html extends BaseReader
             $loaded = false;
         }
         if ($loaded === false) {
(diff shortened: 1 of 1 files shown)


MarkBaker:
Thanks you for this useful addition to diagnostics coverage

PR #1785, opened 2021-01-10
Optimize applyFromArray by caching existing styles

### LINKED ISSUE #1784: Possible to optimize Style::applyFromArray()?
This is:

```
- [x] a feature request
```

Using `applyFromArray` on `$sheet->getStyle("A")` could be faster.

It takes some time.

```php
<?php

require __DIR__ . "/vendor/autoload.php";

$data = array_fill(0, 15000, array_fill(0, 20, 'cell'));

$spreadsheet->getActiveSheet()->fromArray($data);

// Fill up some unique styles in the document
$spreadsheet->getActiveSheet()->getStyle('B1:B1')->applyFromArray(['numberFormat' => ['formatCode' => '0.0000']]);
$spreadsheet->getActiveSheet()->getStyle('B2:B2')->applyFromArray(['numberFormat' => ['formatCode' => '1.0000']]);
$spreadsheet->getActiveSheet()->getStyle('B3:B3')->applyFromArray(['numberFormat' => ['formatCode' => '2.0000']]);
$spreadsheet->getActiveSheet()->getStyle('B4:B4')->applyFromArray(['numberFormat' => ['formatCode' => '3.0000']]);


// Apply style on entire column
$spreadsheet->getActiveSheet()->getStyle('A')->applyFromArray(['numberFormat' => ['formatCode' => '0.0000']]);
$spreadsheet->getActiveSheet()->getStyle('B')->applyFromArray(['numberFormat' => ['formatCode' => '1.0000']]);
```

image
#### applyFromArray()
image

```
phpoffice/phpspreadsheet  1.16.0 
PHP 7.4
```

eigan:
This is:

```
- [x] a new feature
```

Checklist:

- [x] Code style is respected
- :x: Documentation is updated as necessary 
   - _(Please let me know if this is needed)_.

Prevent calling clone and getHashCode when not needed
because these calls are very expensive.

When applying styles to a range of cells can we cache the
styles we encounter along the way so we don't need to look
them up with getHashCode later.

With these changes is `clone` and `getHashCode` only called on unique styles in the set of cells being updated.

#### Memory usage
Since we cache the hashcode for each unique style in the range of cells we update, will sheets with many styles use a bit more memory than before. Though I don't think it should be _that_ much. 

#### Before

```
~ 0.90s
```

#### After

```
~ 0.28s
```

Fixes #1784

### DIFF
--- a/src/PhpSpreadsheet/Style/Style.php
+++ b/src/PhpSpreadsheet/Style/Style.php
@@ -64,6 +64,27 @@ class Style extends Supervisor
     protected $quotePrefix = false;
 
     /**
+     * Internal cache for styles
+     * Used when applying style on range of cells (column or row) and cleared when
+     * all cells in range is styled.
+     *
+     * PhpSpreadsheet will always minimize the amount of styles used. So cells with
+     * same styles will reference the same Style instance. To check if two styles
+     * are similar is Style::getHashCode() used. This call is expensive. To minimize
+     * the need to call this method can we cache the internal PHP object id of the
+     * Style in the range. Style::getHashCode() will then only be called when we
+     * encounter a unique style.
+     *
+     * @see Style::applyFromArray()
+     * @see Style::getHashCode()
+     *
+     * @phpstan-var null|array{styleByHash: array<string, Style>, hashByObjId: array<int, string>}
+     *
+     * @var array<string, array>
+     */
+    private static $cachedStyles;
+
+    /**
      * Create a new Style.
      *
      * @param bool $isSupervisor Flag indicating if this is a supervisor or not
@@ -341,8 +362,14 @@ class Style extends Supervisor
             // Selection type, inspect
             if (preg_match('/^[A-Z]+1:[A-Z]+1048576$/', $pRange)) {
                 $selectionType = 'COLUMN';
+
+                // Enable caching of styles
+                self::$cachedStyles = ['hashByObjId' => [], 'styleByHash' => []];
             } elseif (preg_match('/^A\d+:XFD\d+$/', $pRange)) {
                 $selectionType = 'ROW';
+
+                // Enable caching of styles
+                self::$cachedStyles = ['hashByObjId' => [], 'styleByHash' => []];
             } else {
                 $selectionType = 'CELL';
             }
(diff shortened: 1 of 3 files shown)


eigan:
Any comments @MarkBaker?

eigan:
I am rebasing and fixing the checks and conflicts whenever we are upgrading this dependency in our project. Let me know if you are going to do a review.

eigan:
@MarkBaker @oleibman  anything wrong with this pull request? Don't get why you guys never respond but still the repo is fully active. Seems to be over 200 other pull requests merged since this PR was opened.

oleibman:
My apologies, I do not have sufficient expertise in the area of PHP performance to properly evaluate your change. I will try to read up on it, but it will take a while.

eigan:
@oleibman Thanks. I will rebase and take a second look at the logic myself. Been a while.

eigan:
The work done in a189d933f2e1ff33b582cab976636ec47b6d71c8 messed up my PR. Need to refactor a bit.

eigan:
PR updated. You can test the effect by using the snippet from the linked issue. 

Note: PHPStan bug encountered. $newStyle is actually always defined. Not sure what you guys would prefer to do. Either ignore phpstan message or add if `$newStyle === null {`. Please let me know.

oleibman:
You have a php-cs-fixer problem as well. I believe the phpstan-var statement at line 81 in Style.php needs to be followed by a blank line (well, a line with just an asterisk in column 6) to eliminate the problem.

As for the phpstan problem, I don't yet understand what you're doing well enough to make a recommendation. In the code before you changed it, newStyle was always assigned a value (as a clone of style). With your change, it is only assigned in the first part of an if ... else ... statement. So, Phpstan's analysis is correct - newStyle may be unassigned. You could eliminate the phpstan problem by moving the assignment to newStyle before the if ... but this might affect your performance numbers if you have to make a lot of unused clones. The other logical alternative is to assign null to newStyle before that `if`, and then test if newStyle isn't null before using it later. That should not affect your performance much, but I'm not completely convinced that doing nothing in this case is the correct action. It *probably* is, especially if you can convince me that newStyle will always be non-null here.

eigan:
> I don't yet understand what you're doing well enough to make a recommendation

Not sure how I can help you here. The comments is not enough?

> if you can convince me that newStyle will always be non-null here.

image

Remember, `$newStyle` is only used if `$existingStyle` is falsy.

1) This is OK. `$newStyle` is always set, regardless of `$existingStyle`. 
2) In the else, `$newStyle` is set if `$existingStyle` is `null`. 

Now, there is no way you could end up with `$existingStyle` beeing falsy AND `$newStyle` not defined. https://github.com/phpstan/phpstan/issues/5805

oleibman:
Okay, you've convinced me, you still need to convince Phpstan. This can be done by changing the addCellXf at the bottom of your code:
```php
$workbook->addCellXf($newStyle ?? new Style());
```

This should add minimal overhead and Phpstan should be satisfied.

src/PhpSpreadsheet/Style/Style.php
```diff
@@ -63,6 +63,27 @@ class Style extends Supervisor
...
+     * Internal cache for styles
+     * Used when applying style on range of cells (column or row) and cleared when
+     * all cells in range is styled.
+     *
+     * PhpSpreadsheet will always minimize the amount of styles used. So cells with
+     * same styles will reference the same Style instance. To check if two styles
+     * are similar is Style::getHashCode() used. This call is expensive. To minimize
+     * the need to call this method can we cache the internal PHP object id of the
+     * Style in the range. Style::getHashCode() will then only be called when we
+     * encounter a unique style.
```
PowerKiKi, line 76:
```suggestion
     * are similar Style::getHashCode() is used. This call is expensive. To minimize
     * the need to call this method we can cache the internal PHP object id of the
     * Style in the range. Style::getHashCode() will then only be called when we
     * encounter a unique style.
```

PowerKiKi (approved):

PowerKiKi:
Thank you for your patience. As you probably have guessed by now, there are only three guys on our team, and we all do this in our spare time. So it can take a bit of time until we can process PRs. But quality PR like yours do end up being merged, sooner or later.

On a side note, for your next PR, you might want to consider to break (existing) things in smaller methods. We already have some extremely huge methods, and we'd rather try to avoid that in the future. In this specific case I would probably have written a small class dedicated to cache, to better clarify concerns of each parts, and keep Style.php smaller.

madhurbhaiya:
Sorry, but I think that this PR and eventually v1.19 has broken a few things.

Please check the reported issue here: https://github.com/PHPOffice/PhpSpreadsheet/issues/2366

PR #1805, opened 2021-01-22
Fix Xlsx reader overriding manually set number format with builtin number format

rvanvelzen:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Relevant spreadsheet: https://circabc.europa.eu/ui/group/0e5f18c2-4b2f-42e9-aed4-dfe50ae1263b/library/3bbc2ffd-d4c1-4cbd-b8e5-71aa44f6ed3c/details

This spreadsheet overrides codes 50 and 51, but the reader ignores that and uses the built-in formats instead.

Verified by running:

```
var_dump($worksheet->getCell('E2')->getFormattedValue());
```

... which should output `2`, not some garbled mess

(I was unable to quickly produce a reduced payload - the original file is ~1.5MB which isn't suitable to include here)


MarkBaker:
Thank you for that submission

Issue #1789, opened 2021-01-13
Exception when rounding #VALUE! Cell with PHP 8


davidlienhard:
Hi
I stumbled uppon this issue with **PHP 8** when rounding a cell. I'm not sure if PhpSpreadsheet already supports PHP 8, but according to `composer.json` it does.

This is:
```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
An array with the contents of the Excel-Sheet when using `$worksheet->toArray()`

### What is the current behavior?
An exception is thrown:
```
PHP Fatal error:  Uncaught TypeError: round(): Argument #1 ($num) must be of type int|float, string given in /srv/app/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Calculation/Calculation.php:4714
Stack trace:
#0 [internal function]: round()
#1 /srv/app/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Calculation/Calculation.php(4714): call_user_func_array()
#2 /srv/app/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Calculation/Calculation.php(3442): PhpOffice\PhpSpreadsheet\Calculation\Calculation->processTokenStack()
#3 /srv/app/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Calculation/Calculation.php(3233): PhpOffice\PhpSpreadsheet\Calculation\Calculation->_calculateFormulaValue()
#4 /srv/app/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/Cell.php(257): PhpOffice\PhpSpreadsheet\Calculation\Calculation->calculateCellValue()
#5 /srv/app/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Worksheet/Worksheet.php(2525): PhpOffice\PhpSpreadsheet\Cell\Cell->getCalculatedValue()
#6 /srv/app/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Worksheet/Worksheet.php(2599): PhpOffice\PhpSpreadsheet\Worksheet\Worksheet->rangeToArray()
[...]

PR #1837, opened 2021-02-11
ROUND Accepts null, false, and true as First Parameter

oleibman:
Issue #1789 was addressed by PR #1799. In a follow-up discussion,
it came to light that ROUND was not handling the unexpected case where the
first parameter is an empty cell in the same manner that Excel does.
Subsequent investigation showed that a boolean first parameter is permitted.
I broadened my investigation to include the following related functions.
- ROUNDUP
- ROUNDDOWN
- MROUND
- TRUNC
- INT
- FLOOR
- FLOOR.MATH
- FLOOR.PRECISE
- CEILING
- CEILING.MATH
- CEILING.PRECISE

All of these allow a NULL first parameter, and all except MROUND allow boolean.
For completeness, I will note that all treat null string as invalid.
I suspect there are other functions which permit
similarly unexpected parameters, but I consider them out of scope for this PR.

CEILING.MATH and CEILING.PRECISE were unimplemented, and are now supported
as part of this PR.

The tests for each of these functions have been re-coded, though all the original
test data is still included in the test cases, plus several new cases for each.
The new tests now take place as a user would invoke the functions,
through a spreadsheet cell rather than a
direct call to the appropriate function within Calculation/MathTrig.
Aside from being more realistic, the new tests are also more complete.
For example, FLOOR.MATH can take from 1-3 arguments, and the existing tests
confirmed that the function in Calculation could handle a single argument.
However, the function list in Calculation.php erroneously set the number of
arguments for FLOOR.MATH to exactly 3, so, if a user tried to get the calculated
result of a cell containing FLOOR.MATH(1.2), the result would be an Exception.

Aside from the parameter support, there are a few minor code changes.
Ods, as well as Gnumeric, allows the omission of the second parameter for
FLOAT and CEILING; Excel does not. A potential divide-by-zero error is
avoided in CEILING, FLOOR, and FLOORMATH.

I will note that it would probably be beneficial in terms of maintainability
to break MathTrig up into many individual modules. The same would hold for the
other Calculation modules. I would be willing to look into this if you agree
that it would be worthwhile.

This is:

```
- [x] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected


MarkBaker:
> I will note that it would probably be beneficial in terms of maintainability
to break MathTrig up into many individual modules. The same would hold for the
other Calculation modules. I would be willing to look into this if you agree
that it would be worthwhile.

I very much agree.
Not just MathTrig, but all those calculation function classes. I'd already started considering this when I updated the `CONVERT()` function in `Engineering`, and moved it into its own separate class. My thinking was that I would gradually move other functions or groups of functions into their own separate classes (the `BESSEL` and `Complex` functions would be other prime candidates in Engineering). As you can see with `ConvertUOM()`, the method still exists inside `Engineering.php`, but only as a wrapper to call the `Engineering\CONVERTUOM` class methods, and flagged as Deprecated, with a view to removing it completely for a version 2 release at the end of this year.

I should also have modified the entry in `Calculation.php` to reference `'functionCall' => [Engineering\ConvertUOM::class, 'CONVERTUOM']` instead of simply leaving it as `'functionCall' => [Engineering::class, 'CONVERTUOM']`

It will take a bit of effort, especially in Financial where so many functions are dependent on calculations within other functions; but at least it can be done gradually, group by group over a period of time.

---

I'd also considered one other major change to the way the Calculation engine handles Excel errors for the version 2 release; and that is to use an ExcelException class to represent Excel errors rather than the '#NUM!' type string as at present. This should allow better error trapping/handling in the code, and make it a lot easier to bubble errors through more complex formulae. I'd even created a [branch](https://github.com/PHPOffice/PhpSpreadsheet/pull/1605/files) to test the principle of this approach, but got rather bogged down in the implementation. It's  large chunk of work, but I think it would also help make the function code easier when it came to Excel error handling. What do you think?

oleibman:
I haven't thought about it, so I will just give you my immediate gut reaction. I think issuing Exception instead of the Excel error strings could be disruptive. It might make it difficult to read an existing spreadsheet that contained this type of error. And, even when creating a new one, I believe that you don't automatically perform an evaluation when you set a cell to a formula, so the Exception may occur long after the damage. In particular, it may not show up until you attempt to save the spreadsheet to a file, an unexpected and difficult-to-debug situation. If you do this, it might be best to have a global setting so that, depending on the setting, Function::NAN() etc. either returns the error string or issues an Exception.

oleibman:
Based upon your agreement about breaking up MathTrig, I will start that effort by continuing to work on this change to move the affected functions to their own classes. Expect a new version by the end of the weekend.

oleibman:
All 12 functions are now moved to their own classes. No further changes planned for this PR.

src/PhpSpreadsheet/Calculation/MathTrig.php
```diff
@@ -212,47 +212,6 @@ public static function BASE($number, $radix, $minLength = null)
         return Functions::VALUE();
     }
 
-    /**
```
MarkBaker, line 215: Can you leave the docblock and stub code that calls these new class/methods in the MathTrig class with a @deprecated notice

[You can see my examples in the Engineering class in master branch](https://github.com/PHPOffice/PhpSpreadsheet/blob/42ecc270ecfb85ea7c419481cd3622be2e6d1e7e/src/PhpSpreadsheet/Calculation/Engineering.php#L46).

It is possible to call these MathTrig methods directly without using the calculation engine (I believe that one of our examples in sample code actually does this, or used to do so); and this allows us to flag the approach as deprecated now, so that existing code will still work; but prepares for removal in a 2.0.0 release.

src/PhpSpreadsheet/Calculation/MathTrig/CeilingMath.php
```diff
@@ -0,0 +1,61 @@
...
+    /**
+     * Let CEILINGMATH complexity pass Scrutinizer.
+     *
+     * @param mixed $number Number to round
+     * @param mixed $significance Significance
+     * @param mixed $mode direction to round negative numbers
+     *
+     * @return bool
+     */
+    private static function ceilingMathTest($significance, $number, $mode)
```
MarkBaker, line 57: With private internal methods like this, we can actually start to use typehinting and return typehinting in their signatures

src/PhpSpreadsheet/Calculation/MathTrig.php
oleibman: Pushed a new set of changes. All of the routines which had existed are now restored as deprecated comment block plus stubs in MathTrig. CEILINGMATH and CEILINGPRECISE are new, and I did not create deprecated stubs for them. I added a cursory test member to confirm that each of the deprecated calls continues to work as expected. That member should be deleted when the deprecated routines are removed.

src/PhpSpreadsheet/Calculation/MathTrig/CeilingMath.php
oleibman: Done, in the new push.

PR #1840, opened 2021-02-11
Fix PPMT, IMPT, CUMPRINC, CUMIPMT, AMORLINC & AMORDEGRC

ElPopcorn:
While using PHPSpreadsheet, I have encountered some inconsistent behaviour compared to Excel, i.e:
* PPMT & IPMT would not return the same error values that Excel returns
* CUMPRINC & CUMIPMT calls would crash if one of the call to PPTM or IPMT returned an error (in the form of a string)
* AMORLINC & AMORDEGRC calls would crash if a call to YEARFRAC returned an error (i.e an Excel error in the form of a string)

So I :
* Updated PPMT & IPMT implementations to better stick with Excel error values
* Updated CUMPRINC & CUMIPMT implementation to prevent a crash, and return the error value.
* Updated AMORLINC & AMORDEGRC implementation to prevent a crash, and return the error value returned by YEARFRAC
* Added test sets related to aforementioned issues

This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

### Why is this change is needed?

This increases PHPSpreadsheet consistency with Excel.


frantzmiccoli:
I have reviewed and tested this PR with test files (I made a few similar fixes a couple years ago), if that helps to ease the review.

MarkBaker (approved):
Thank you for this improvement to error handling in the Excel Financial functions that are implemented for PhpSpreadsheet; it's good to see improvements in the "unhappy path"

PR #1842, opened 2021-02-11
Number format mask helpers

MarkBaker:
This is:

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [X] Code style is respected

Provide wizard helpers for the creation of more complex Number, Currency and Accounting number format masks, providing easy support for
- leading/trailing sign (+/-)
- optionally displaying a positive sign
- with/without thousands separators
- configurable number of decimal places
- automatic identification of currency symbol for a given currency code
- leading/trailing currency symbols
- wrapping negative values in braces for accounting masks

If the Intl extension is enabled, then locale rules will be applied based on standards for the specified locale; although these can be overridden for additional flexibility


PowerKiKi:
Converting to draft according to Mark in https://gitter.im/PHPOffice/PhpSpreadsheet?at=60605e9a33346c1627806006

PR #1844, opened 2021-02-12
Bug fix #1685. PHP7.4 and WMF file error

### LINKED ISSUE #1685: PHP 7.4 excel with macrosTrying to access array offset on value of type bool
hello,

In a _symfony4_ project, I generate an **excel** file with **macros** with spreadsheet.
with **php7.2** it 's work fine. but when I try with a **7.4** php version I get this: 
**_Notice: Trying to access array offset on value of type_** bool

my pfile is :  "zip:///home/..../test.xlsm#xl/media/image62.wmf" & xl/media/image62.wmf is added automatically 
`getimagesize($pfile); `==> return false so `return image_type_to_mime_type($image[2]);` return this Notice
any help plz?

hirossyi73:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

- If use PHP7.4 and contains .wmf file in xlsx file, and save, throw Exception.

```
Notice: Trying to access array offset on value of type bool
```

- I explain why this error throws.

```
// When $pFile is .wmf, return "false". (When $pFile is .png, return fileinfo array.)
$image = getimagesize($pFile);

// When $image[2] throws exception on PHP7.4 when $image[2] is false.
return image_type_to_mime_type($image[2]);
```

- Sample files are here. Contains wmf file. (I cannot simple .wmf file.... So only xlsx file contains wmf.)
[WmfFile.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/5970940/WmfFile.xlsx)

- Sample code is here.

``` php
<?php

$reader = IOFactory::createReader('Xlsx');
$spreadsheet = $reader->load('C:\\work\\WmfFile.xlsx');

$writer->save('C:\\work\\hello world.xlsx');
```


hirossyi73:
@MarkBaker 
Hi, I have to solve this.
Scrutinizer is time out (I think this is system error). What I should do?

oleibman (changes requested):
The function image_type_to_mime_type is documented to take an int argument, but you might pass null. Will your change work if you specify 0 instead of null in your ternary expression? Also, can you write a formal unit test to show that your change works as desired?

Yehorko (approved):
I think this construction with returning null is right idea.
Because when we using this functionality in php 7.3, 
in this line `return image_type_to_mime_type($image[2]);` when $image[2] not set, it returns null by it's self but in 7.4 we need manually return null to avoid errors.
In addition i can say that when i switched my project to php version 7.4, all such errors i resolving in this style: $image[2] ?? null; Because in  previous php version(7.3) it already returns null;

Yehorko:
> The function image_type_to_mime_type is documented to take an int argument, but you might pass null. Will your change work if you specify 0 instead of null in your ternary expression? Also, can you write a formal unit test to show that your change works as desired?

In php 7.3 in this case we passing null to image_type_to_mime_type(null) and this function returns us a string "application/octet-stream". And after proposed fix it will return as the same result "application/octet-stream". As result all depending functionality has to keep working  as it was before. Because nothing will changed for them.

oleibman:
PHP 8.1 has been aggressively adding enforcement of argument and return types to the builtin functions. Supplying null for an argument documented as int will result in failure sooner or later. My testing indicates that using 0 instead of null also results in application/octet-stream for all releases of Php on which PhpSpreadsheet can run. Do you see different results when you try it?

Yehorko:
> PHP 8.1 has been aggressively adding enforcement of argument and return types to the builtin functions. Supplying null for an argument documented as int will result in failure sooner or later. My testing indicates that using 0 instead of null also results in application/octet-stream for all releases of Php on which PhpSpreadsheet can run. Do you see different results when you try it?

Agree with you, 0 will necessary for php 8.

oleibman:
Converting to draft as PR #2339 supersedes it.

PowerKiKi:
Superseded by #2339

PR #1854, opened 2021-02-14
Update RowCellIterator.php

ShahrozButt:
There is no specific way to get a cell through the RowCellIterator. I have seen many posts where developers are getting the rowIndex through row and then putting in the column index that they have to get to a specific cell. Where I think as the Row (RowCellIterator knows current $rowIndex ) knows what it's index is, we should only have to provide the Column (in String) to get to the cell. Similar function should be added for ColumnCellIterator where Column knows its index a row index is to be provided for the cell.

This is:

```
- [ ] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/RowCellIterator.php
+++ b/src/PhpSpreadsheet/Worksheet/RowCellIterator.php
@@ -173,6 +173,21 @@ class RowCellIterator extends CellIterator
     {
         return $this->currentColumnIndex;
     }
+    
+     /**
+     * Return the current Cell by Column position.
+     * Row Iterator already knows the current $rowIndex
+     * @param string $column The column address for the cell
+     * Edited Shahroz BUTT
+     * Email : shahroz.butt@yahoo.com
+     *
+     * @return Cell
+     */
+    public function getCellByColumn($column)
+    {
+        $column = Coordinate::columnIndexFromString($column);
+        return $this->worksheet->getCellByColumnAndRow($column, $this->rowIndex);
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Worksheet/RowCellIterator.php
```diff
@@ -173,6 +173,21 @@ public function getCurrentColumnIndex()
...
+     * Row Iterator already knows the current $rowIndex
+     * @param string $column The column address for the cell
+     * Edited Shahroz BUTT
+     * Email : shahroz.butt@yahoo.com
+     *
+     * @return Cell
+     */
+    public function getCellByColumn($column)
+    {
+        $column = Coordinate::columnIndexFromString($column);
```
MarkBaker, line 188: This functionality can be achieved using `$myRowCellIterator->seek('B')->current();` which has the added advantage of validating that the input column is within the specified range; though it also has the side-effect of updating the iterator index which may or may not be the required behaviour.

If the user wanted a column that wasn't within the rowCellIterator range of columns, then the functionality should really be in the rowIterator, rather than in the rowCellIterator. That might well be a better option anyway, because it gives direct access to cells at a higher-level, and leaves the cellIterator as a pure iterator.

MarkBaker (commented):
Please note the use of PSR12 coding standards.in the project

And you're correct; if this functionality was to be provided, it should be provided when iterating both b columns and by rows.

Issue #1872, opened 2021-02-22
Error: Call to a member function getHashCode() on null


ivanstan:
Got error 
```
Error: Call to a member function getHashCode() on null
```
In following code there is no check if `$definedName->getWorksheet()` returns `null`

https://github.com/PHPOffice/PhpSpreadsheet/blob/1318b90330773d852ee15c8b432a7e7175c4bd55/src/PhpSpreadsheet/ReferenceHelper.php#L611

while in definition of the method `getWorksheet`, `null` is listed as a possible outcome. 

https://github.com/PHPOffice/PhpSpreadsheet/blob/1318b90330773d852ee15c8b432a7e7175c4bd55/src/PhpSpreadsheet/DefinedName.php#L162

MarkBaker:
Fixed

PR #1873, opened 2021-02-22
Update ReferenceHelper.php

ivanstan:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

Fixes issue https://github.com/PHPOffice/PhpSpreadsheet/issues/1872

### DIFF
--- a/src/PhpSpreadsheet/ReferenceHelper.php
+++ b/src/PhpSpreadsheet/ReferenceHelper.php
@@ -608,7 +608,12 @@ class ReferenceHelper
         // Update workbook: define names
         if (count($pSheet->getParent()->getDefinedNames()) > 0) {
             foreach ($pSheet->getParent()->getDefinedNames() as $definedName) {
-                if ($definedName->getWorksheet()->getHashCode() === $pSheet->getHashCode()) {
+                $hash = null;
+                if ($definedName->getWorksheet()) {
+                    $hash = $definedName->getWorksheet()->getHashCode();
+                }
+
+                if ($hash === $pSheet->getHashCode()) {
                     $definedName->setValue($this->updateCellReference($definedName->getValue(), $pBefore, $pNumCols, $pNumRows));
                 }
             }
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/ReferenceHelper.php
```diff
@@ -608,7 +608,12 @@ public function insertNewBefore($pBefore, $pNumCols, $pNumRows, Worksheet $pShee
         // Update workbook: define names
         if (count($pSheet->getParent()->getDefinedNames()) > 0) {
             foreach ($pSheet->getParent()->getDefinedNames() as $definedName) {
-                if ($definedName->getWorksheet()->getHashCode() === $pSheet->getHashCode()) {
+                $hash = null;
+                if ($definedName->getWorksheet()) {
```
MarkBaker, line 612: Just for the sake of readability, can you do an explicit comparison for not null)..

This is something that I've been doing in all the code that I've been working with recently, and it does make the if comparison a lot more explicit rather than relying on the loose typed truthiness/falsiness of PHP

MarkBaker (commented):
Thank you for identifying this problem, and for the resolution

src/PhpSpreadsheet/ReferenceHelper.php
```diff
@@ -608,7 +608,7 @@ public function insertNewBefore($pBefore, $pNumCols, $pNumRows, Worksheet $pShee
         // Update workbook: define names
         if (count($pSheet->getParent()->getDefinedNames()) > 0) {
             foreach ($pSheet->getParent()->getDefinedNames() as $definedName) {
-                if ($definedName->getWorksheet()->getHashCode() === $pSheet->getHashCode()) {
+                if ($definedName->getWorksheet() !== null && $definedName->getWorksheet()->getHashCode() === $pSheet->getHashCode()) {
```
ivanstan, line 611: @MarkBaker does this make sense? 
I don't think updateCellReference should be called on a null worksheet

MarkBaker: Makes perfect sense. thanks for the quick update

MarkBaker (approved):

PR #1880, opened 2021-02-23
Initial experiments using the new Database query logic with Conditional Statistical Functions

MarkBaker:
Initial experiments using the new Database query logic with Conditional Statistical functions (e.g. COUNTIF(), AVERAGEIF(), COUNTIFS(), etc)

This is:

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [X] Code style is respected

Using the new Database function query logic to support all relevant forms of queries in the Conditional Statistical functions, including support for boolean datatypes, wildcard text, etc

Also using this to refactor the Conditional functions outside of the main Statistical class, and into a dedicated class as part of the big refactoring of all Excel function implementations


MarkBaker:
Down to just 2 failing tests; and those are related to how MS Excel handles counting values, and the differences between literals and values from cells... MS! Why do you make things so inconsistent?

Issue #1887, opened 2021-03-01
Setting selected cells sometimes has no effect with XLSX writer


francescom:
This is:

- [ X ] a bug report

### What is the expected behavior?
At the end of the creation of a multiworksheet Spreadsheet I use setSelectedCells to leave the way the client expected: with one specific cell selected. The expected behaviour is that opening the generated file with Excel the user finds the cell selected

### What is the current behavior?
The current behaviour is that on the first worksheet the right cell is selectred while on subsequent (different) worksheets the selected cell is the bottom right cell (the last written). When the file is saved, the functions take care of the saving, call getStyle() that calls setSelectedCells() altering the selection of the worksheets.

I managed to track the changing of the selection to an unwanted value inserting some debug code in Worksheet.php and this is a stack trace of the time the value is altered:

        Array
        (
            [0] => Array
                (
                    [file] => C:\Users\fmm\Dropbox\sharedWorks\Web\www_htdocs\report\report\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Worksheet\Worksheet.php
                    [line] => 1427
                    [function] => setSelectedCells
                    [class] => PhpOffice\PhpSpreadsheet\Worksheet\Worksheet
                    [type] => ->
                )

            [1] => Array
                (
                    [file] => C:\Users\fmm\Dropbox\sharedWorks\Web\www_htdocs\report\report\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Cell\Cell.php
                    [line] => 532
                    [function] => getStyle
                    [class] => PhpOffice\PhpSpreadsheet\Worksheet\Worksheet
                    [type] => ->
                )

            [2] => Array
                (
                    [file] => C:\Users\fmm\Dropbox\sharedWorks\Web\www_htdocs\report\report\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Calculation\Calculation.php
                    [line] => 3391
                    [function] => getStyle
                    [class] => PhpOffice\PhpSpreadsheet\Cell\Cell
                    [type] => ->
                )

            [3] => Array
                (
                    [file] => C:\Users\fmm\Dropbox\sharedWorks\Web\www_htdocs\report\report\vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Calculation\Calculation.php
                    [line] => 3233
                    [function] => _calculateFormulaValue
[...]

PR #1908, opened 2021-03-07
Fix for Issue #1887 - Lose Track of Selected Cells After Save

oleibman:
Issue #1887 reports that selected cells are lost after saving Xlsx. Testing indicates that this applies to the object in memory, though not to the saved spreadsheet.

Xlsx writer tries to save calculated values for cells which contain formulas. Calculation::_calculateFormulaValue issues a getStyle call merely to retrieve the quotePrefix property, which, if set, indicates that the cell does not contain a formula even though it looks like one. A side-effect of calls to getStyle is that selectedCell is updated. That is clearly accidental, and highly undesirable, in this case. Code is changed to save selectedCell before getStyle call and restore it afterwards.

The problem was reported only for Xlsx save. To be on the safe side, test is made for output formats of Xlsx, Xls, Ods, Html (which basically includes Pdf), and Csv. For all of those, the object in memory is tested after the save. For Xlsx and Xls, the saved file is also tested. It does not make sense to test the saved file for Csv and Html. It does make sense to test it for Ods, but the necessary support is not yet present in either the Ods Reader or Ods Writer - a project for another day.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Bug fix - see details above.

### DIFF
--- a/src/PhpSpreadsheet/Calculation/Calculation.php
+++ b/src/PhpSpreadsheet/Calculation/Calculation.php
@@ -3389,9 +3389,17 @@ class Calculation
     {
         $cellValue = null;
 
+        $pCellParent = ($pCell !== null) ? $pCell->getWorksheet() : null;
         //  Quote-Prefixed cell values cannot be formulae, but are treated as strings
-        if ($pCell !== null && $pCell->getStyle()->getQuotePrefix() === true) {
-            return self::wrapResult((string) $formula);
+        if ($pCell !== null) {
+            $selected = isset($pCellParent) ? $pCellParent->getSelectedCells() : '';
+            $quoted = $pCell->getStyle()->getQuotePrefix();
+            if ($pCellParent !== null) {
+                $pCellParent->setSelectedCells($selected);
+            }
+            if ($quoted === true) {
+                return self::wrapResult((string) $formula);
+            }
         }
 
         if (preg_match('/^=\s*cmd\s*\|/miu', $formula) !== 0) {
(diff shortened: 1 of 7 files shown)


oleibman:
Actually, I do not understand why getStyle should affect the selected cells at all. I chose this approach so as to have the minimum effect on users. If they frequently use getStyle for this purpose, then perhaps its best to leave it alone. On the other hand, why should they use in this way? There's a really simple API setSelectedCells available for precisely this purpose. So the application change required by anyone adversely affected would be really trivial. I would certainly be willing to eliminate the selectedCellls side effect of getStyle (which would mean we don't have to save and restore) if you feel that is the way we should proceed.

oleibman:
No, that seems like a non-starter. Everything I try to do in Worksheet winds up with a ton of both errors and failures in the test suite. "Everything" includes:
- commenting out setSelected
- adding a parameter which, when not default, restores selected
- some other desperation moves

I especially don't understand why the second option doesn't work; it seems like it should be somewhat equivalent to what's already been done.

oleibman:
Aha! I can't move this to Worksheet, but I can move it to Cell. And there is justification for this - Cell::getCalculatedValue saves off ActiveSheetIndex before doing the calculation and restores it after. Saving and restoring SelectedCells here would be done for exactly the same reason. I can have a new version ready tomorrow.

oleibman:
I have also figured out support for ODS ActiveSheet and SelectedCells for Reader and Writer. I will add it to tomorrow's change as well, unless you'd prefer me to schedule that for later.

MarkBaker (approved):
Nice work, and good to see selected cells added to ODS as well

I'm slightly concerned about calculations that call functions which in turn call the calculation engine again (e.g. the Database functions, or functions that assess conditions like COUNTIF(), MAXIFS(), etc... but I think this will be enough for now

PR #1914, opened 2021-03-10
Add null typehint to Worksheet::getColumnDimension() since it returns null

simPod:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Should this be somehow more documented? It has always returned null.

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -1337,7 +1337,7 @@ class Worksheet implements IComparable
      * @param int $pRow Numeric index of the row
      * @param bool $create
      *
-     * @return RowDimension
+     * @return null|RowDimension
      */
     public function getRowDimension($pRow, $create = true)
     {
@@ -1390,7 +1390,7 @@ class Worksheet implements IComparable
      *
      * @param int $columnIndex Numeric column coordinate of the cell
      *
-     * @return ColumnDimension
+     * @return null|ColumnDimension
      */
     public function getColumnDimensionByColumn($columnIndex)
     {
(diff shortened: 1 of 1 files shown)


simPod:
CI says there's some violation but does not say what is the violation.

oleibman:
To get rid of the CI violation, try `@return null|ColumnDimension` rather than `@return ColumnDimension|null` in your Doc block.

simPod:
oh it's the other way it usually is. thx

MarkBaker:
In most cases, when a phpdoc annotation indicates that a method accepts or returns multiple different datatypes, they should be listed alphabetically (e.g. `float|int` rather than `int|float`); but `null` is a special case, in the same way that we typehint `?MyObjectClass`, with the `?` in front of the object class name

MarkBaker (changes requested):
The same issue applies to `getRowDimension()` and to `getColumnDimensionByColumn()` as well: could you include fixing those docblocks in the same PR?

Thanks

simPod:
Yes. Done in the next commit.

Btw what will the special case be with PHP8 union types for eg. `fcn(): MyObjectClass|null`?

MarkBaker:
> Btw what will the special case be with PHP8 union types for eg. fcn(): MyObjectClass|null?

I'm not sure that any standards for using the keyword `null` in a union return-type have been defined yet; my inclination would be to use `?MyObjectClass` as at present; and apply union types only when you have something like `fcn(): false|MyObjectClass`; and with scalars before objects.

We try to adhere to PSR-12 for PHPSpreadsheet, so really it's a case of waiting for the FIG to identify appropriate standards; but until we have PHP8 as a minimum version, it's not an option for us (and that's a few years away yet)

simPod:
just curious, thanks for input!

MarkBaker (approved):
Thanks for the contribution, and for the quick response for my request to fix the other methods as well.

PR #1917, opened 2021-03-11
Fix SpreadsheetML (xml) detection

christof-b:
Replace the unrequired product signature by the
required namespace definition for XML Spreadsheet.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Fixes the issue #1917  and allows processing of exported SpreadsheetML files from other sources than MS Excel itself.


MarkBaker (approved):
You're dragging my mind back now to building the XML Reader without any documentation.

Thanks for the correction

PR #1920, opened 2021-03-12
Coverage for Helper/Samples

oleibman:
I was perplexed by the fact that Helper/Samples seemed to be entirely uncovered when running the test suite, since I know all the samples are run as part of the test. I think that what must be happening is that the Helper code is invoked mostly as part of a Data Provider (and therefore not counted), not as part of the test proper (which would count). So, this change adds a small number of tests which result in Samples being 100% covered.

Covering one statement was tricky - simulating the inability to create a test directory. Mocking, a technique I have not used before, solves this problem admirably.

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] code coverage
```

Checklist:

- [x] Code style is respected

Code coverage.

### DIFF
--- a/src/PhpSpreadsheet/Helper/Sample.php
+++ b/src/PhpSpreadsheet/Helper/Sample.php
@@ -71,7 +71,7 @@ class Sample
     /**
      * Returns an array of all known samples.
      *
-     * @return string[] [$name => $path]
+     * @return string[][] [$name => $path]
      */
     public function getSamples()
     {
--- a/src/PhpSpreadsheet/Helper/Sample.php
+++ b/src/PhpSpreadsheet/Helper/Sample.php
@@ -132,6 +132,11 @@ class Sample
         $this->logEndingNotes();
     }
 
+    protected function isDirOrMkdir(string $folder): bool
+    {
+        return \is_dir($folder) || \mkdir($folder);
+    }
+
     /**
      * Returns the temporary directory and make sure it exists.
      *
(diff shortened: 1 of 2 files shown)


MarkBaker:
I try to avoid mocking as much as I can while writing unit tests, but sometimes it's impossible to avoid. (I'm looking at the code for INDIRECT() and OFFSET() at the moment, and can't see how it will be possible to unit test those functions without mocking)

There is a useful little library that can be used with PHPUnit to simulate the filesystem, that I have considered using for some of the Readers/Writers.... https://github.com/bovigo/vfsStream - I don't know if that would help

MarkBaker:
I know that a number of the mocking functions in PHPUnit are different with the more recent releases (e.g. some assertions and expectations are dropped in favour of others; and the MockBuilder `setMethods()` is deprecated; but for the range of versions we cover, we still need to use it).

- I'm looking at using a set of polyfills so that we can write tests that use the latest PHPUnit version assertions and expectations, and the tests will still run with the older versions of PHPUnit, which will allow us to reduce the range of PHPUnit versions that we need to work with to support the range of PHP versions that we cover.
- I'm also looking at starting to use the PHPUnit `@covers` annotations to make the tests more specific to the units that they are actually testing, which will drop our coverage stats initially, but make it easier to see which parts of the code are really covered by unit tests, and which parts are only covered peripherally, and need proper testing themselves.
- And again, I'm looking at switching some of the existing assertions to be more specific (e.g. `assertSame()` rather than `assertEquals()` testing both type and value).

So many things to consider; and not enough time.... but for the moment, this exercise in splitting up the Excel Function classes is showing real improvements... improving the code as we go along, improving the testing for edge cases, identifying and fixing bugs and discrepancies (although I'm dreading the Financial functions, because they're so interdependent)... and I'm really grateful for all your help.

oleibman:
I added parameter validation in the most recent push.
I also changed from assertEquals to assertSame in tests, and will try to do so consistently in future when appropriate.
The test in this change was quite amenable to use of `@covers`, so I added it; however, unlike assertSame, I think it needs to be carefully evaluated on a case by case basis.

I think vfsStream would be overkill for this test. I would not be comfortable adding it as a dependency.

I would be willing to add breaking up Financials to my to-do list. It has some dependencies on DateTime, which I am currently working on breaking up, and I would like to complete that work first.

PR #1955, opened 2021-03-26
clean drawing from active worksheet

holacliente:
This is:

```
- [ ] a bugfix
- [ x] a new feature
```

Checklist:

- [x] Code style is respected

Sometimes we need show the excel without images


holacliente:
i so hope it will be added, because i was looking for this method

MarkBaker (changes requested):
Thank you for this addition

Can you set gitignore for the sublime settings

holacliente:
> Thank you for this addition
> 
> Can you set gitignore for the sublime settings

Sorry for it, i will fix it

PowerKiKi:
Thank you for the contribution. However I don't think this need to be added because you can get the same result without any change in PhpSpreadsheet:


```php
<?php

$sheet->getDrawingCollection()->exchangeArray([]);
```

PR #1976, opened 2021-04-01
Semi-Automatic Update of Doc Members with Pointers to Functions

oleibman:
A lot of functions have moved around lately. We're not done yet, but updating the documentation now is probably a worthwhile effort. It would be a tedious effort to do so manually. I've written a way to automate it. It runs as a test which creates the doc members in the same directory as the output of the samples. I didn't want to make it one of the samples, because it doesn't really execute any PhpSpreadsheet code. I also didn't want it to update the doc member automatically because that seems danagerous. Having the developer manually copy the new files to replace the existing docs after running the tests sounds like about the right combination of risk and accuracy to me.

There is only a single source line changed, and it is not even executable - the array of functions in Calculation is made protected rather than private to allow the test to inherit it.

There are some minor differences between the old version and the new. Where the old one says, for example, "CATEGORY_DATE_AND_TIME", the new just says "Date and Time". It is only a bit difficult to make it work like the old; I like this better, so I need to be nudged if someone prefers the old. I did, however, retain the complete class name, e.g. `PhpOffice\PhpSpreadsheet\Calculation\Financial\Securities\AccruedInterest::periodic`. I think the tables might look a bit better by eliminating at least `PhpOffice\PhpSpreadsheet\`, and possibly even `Calculation\`, all of which are repeated on almost every line (except for unimplemented functions), and am willing to proceed in that vein if deemed worthwhile.

This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/docs/references/function-list-by-category.md
+++ b/docs/references/function-list-by-category.md
@@ -1,544 +1,545 @@
 # Function list by category
 
-## CATEGORY_CUBE
-
-Excel Function      | PhpSpreadsheet Function
---------------------|-------------------------------------------
-CUBEKPIMEMBER       | **Not yet Implemented**
-CUBEMEMBER          | **Not yet Implemented**
-CUBEMEMBERPROPERTY  | **Not yet Implemented**
-CUBERANKEDMEMBER    | **Not yet Implemented**
-CUBESET             | **Not yet Implemented**
-CUBESETCOUNT        | **Not yet Implemented**
-CUBEVALUE           | **Not yet Implemented**
-
-## CATEGORY_DATABASE
-
-Excel Function      | PhpSpreadsheet Function                                      
---------------------|-------------------------------------------
-DAVERAGE            | \PhpOffice\PhpSpreadsheet\Calculation\Database::DAVERAGE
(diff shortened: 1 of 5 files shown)


oleibman:
Even though it says the change passes muster, Scrutinizer is behaving strangely. Where it lists "changed elements" there are a bunch of things listed that this PR does not touch. It looks like it's reporting simultaneously on my PR and one from Mark Baker. I'm not sure if this has anything to do with the Github "This branch is out-of-date with the base branch" message that shows on my PR.

It seems to be there now for every recent PR. It seems like a good way for Github to break its own model - won't every merge cause all other branches, and therefore PRs based on the other branches, to go out-of-date with base? It's not blocking things now, but I'm concerned.

Thank you for the invitation to join the project. I'm not sure what I might do with my new powers. I will certainly proceed slowly, but the invitation had an expiry date, so I thought it would be better to accept now and learn later.

MarkBaker:
There is actually a script that updates these tables automatically I believe: `generate-document` in `/bin`

I've already re-invented this wheel myself once.

oleibman:
Great! It uses Reflection, which I'm not really familiar with, so I get to learn something new and valuable even though my change isn't used. Thanks!

PR #1982, opened 2021-04-03
PHPStan Level 2

PowerKiKi:


PowerKiKi:
@MarkBaker, @oleibman, I'd like to bring to your attention the new [Coordinate::indexesFromString()](https://github.com/PHPOffice/PhpSpreadsheet/blob/d02352845c998f5b340d50f8692ad34e417ff3fc/src/PhpSpreadsheet/Cell/Coordinate.php#L43-L58) methods, that take a coordinates `"A3"` and returns indexes `[1, 3]`. That allows [simplifications](https://github.com/PHPOffice/PhpSpreadsheet/commit/d02352845c998f5b340d50f8692ad34e417ff3fc#diff-17b9222ccc342fb3f39c26a50ef3a844eb9b484db84bab2d5b1145a9d6dace7fR378
) while being more type safe.

But most importantly it properly [handle absolute coordinates](https://github.com/PHPOffice/PhpSpreadsheet/blob/d02352845c998f5b340d50f8692ad34e417ff3fc/tests/data/Cell/IndexesFromString.php#L46-L73), whereas this was not usually not handled at all in "manual" code and would probably lead to bugs.

PR #1985, opened 2021-04-04
Continue MathTrig Breakup - Completion!

oleibman:
Continuing the process of breaking MathTrip.php up into smaller classes. This round takes care of everything that was left:
- ABS
- DEGREES
- EXP
- RADIANS
- SQRT
- SQRTPI
- SUMSQ, SUMX2MY2, SUMX2PY2, SUMXMY2

The only notable logic change was that the 3 SUMX* functions had accepted arrays of unlike length; in that condition, they now return N/A, as Excel does. There had been no tests for this condition.

All the functions in MathTrig.php are now deprecated. Except for COMBIN, the test suite executes them only from MathTrig MovedFunctionsTest. COMBIN is still directly called by some Statistics Binomial functions which have not yet had the opportunity to be re-coded for the new location.

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] code refactoring
```

Checklist:

- [x] Code style is respected

Maintainability


oleibman:
Just to give you some feedback, PHPStan had a success with this PR, but wasn't entirely successful. My fork was created before PHPStan had been added, so I didn't see any local results. I will recreate my fork after this to see how it works locally. In the meantime, for my initial push, PHPStan detected a capitalization problem that Scrutinizer probably would also have detected, but PHPStan detected it much earlier, so I could work on the fix faster. Win for PHPStan! On the other hand, the doc blocks for the 4 functions moved to SumSqares indicated returns float, as the original doc blocks in MathTrig.php did, but, like most of the Excel functions, should have indicated returns float|string. Scrutinizer did complain about those, but PHPStan let them pass.

oleibman:
If this is merged, I am considering a follow-up PR to change the funcWhatever method names to evaluate, in order to give the code a more consistent look. Do you think this is worthwhile? If so, should I deprecate the use of funcWhatever, or is it okay to just get rid of it?

MarkBaker:
> All the functions in MathTrig.php are now deprecated. Except for COMBIN, the test suite executes them only from MathTrig MovedFunctionsTest. COMBIN is still directly called by some Statistics Binomial functions which have not yet had the opportunity to be re-coded for the new location.

I have my Deprecation Analyser that checks the codebase for any calls/references to deprecated functions/methods/constants; and while I've updated a few such references, I'm really waiting till everything is completed before I do a final update, because everything will still work for the moment if there's calls to the deprecated methods.

I do want to go through the old Functions classes doing a sanity check on all the docblocks, and on the functions array in Calculation, before release.

> Just to give you some feedback, PHPStan had a success with this PR, but wasn't entirely successful. My fork was created before PHPStan had been added, so I didn't see any local results. I will recreate my fork after this to see how it works locally. In the meantime, for my initial push, PHPStan detected a capitalization problem that Scrutinizer probably would also have detected, but PHPStan detected it much earlier, so I could work on the fix faster. Win for PHPStan! On the other hand, the doc blocks for the 4 functions moved to SumSqares indicated returns float, as the original doc blocks in MathTrig.php did, but, like most of the Excel functions, should have indicated returns float|string. Scrutinizer did complain about those, but PHPStan let them pass.

I spent most of yesterday afternoon/evening going through the original 480+ phpstan issues, got it reduced to less than 50; and only one or two of those were also issues in Scrutinizer. Conversely, fixing a couple of Scrutinizer issues had no affect on PHPStan.... and i've introduced one new issue this morning according to Scrutinizer, where it complains that a value I'm comparing to 0 in an `if` test is always 0... and I know that isn't the case... so another Scrutinizer false positive.
The biggest benefits with PHPStan is that we can run it locally, and can gradually change the level setting.

> If this is merged, I am considering a follow-up PR to change the funcWhatever method names to evaluate, in order to give the code a more consistent look. Do you think this is worthwhile? If so, should I deprecate the use of funcWhatever, or is it okay to just get rid of it?

I do think that it's worthwhile. And don't deprecate the `funcWhatever` methods, just rename them... the only value in deprecating them would have been if we'd done a formal release with those names.

src/PhpSpreadsheet/Calculation/Calculation.php
```diff
@@ -228,7 +228,7 @@ class Calculation
     private static $phpSpreadsheetFunctions = [
         'ABS' => [
             'category' => Category::CATEGORY_MATH_AND_TRIG,
-            'functionCall' => [MathTrig::class, 'builtinABS'],
+            'functionCall' => [MathTrig\Abs::class, 'evaluate'],
```
MarkBaker, line 231: `absolute` might be a better method name in this case, no risk of confusion with the core PHP equivalent name

src/PhpSpreadsheet/Calculation/Calculation.php
```diff
@@ -2185,7 +2185,7 @@ class Calculation
         ],
         'SERIESSUM' => [
             'category' => Category::CATEGORY_MATH_AND_TRIG,
-            'functionCall' => [MathTrig::class, 'SERIESSUM'],
+            'functionCall' => [MathTrig\SeriesSum::class, 'funcSeriesSum'],
```
MarkBaker, line 2188: Not `evaluate()`?

src/PhpSpreadsheet/Calculation/Calculation.php
```diff
@@ -2205,7 +2205,7 @@ class Calculation
         ],
         'SIN' => [
             'category' => Category::CATEGORY_MATH_AND_TRIG,
-            'functionCall' => [MathTrig::class, 'builtinSIN'],
+            'functionCall' => [MathTrig\Sin::class, 'funcSin'],
```
MarkBaker, line 2208: not `evaluate()`?

tests/data/Calculation/MathTrig/SQRT.php
```diff
@@ -1,10 +1,16 @@
 <?php
 
 return [
-    ['#VALUE!'], // exception not enough args
-    ['#VALUE!', '"ABC"'],
+    ['exception'], // exception not enough args
```
MarkBaker, line 4: Just as a future reference; a useful trick with unit test dataproviders is to use an associative key for each test case, e.g.
```
'exception not enough args' => ['exception'],
'Non-numeric first argument' => ['#VALUE!', '"ABC"'],
```
It makes things a lot easier to see which test cases have failed when running tests, because PHPUnit doesn't simply report it as failing test case <#n>, but as failing the `exception not enough args` test

src/PhpSpreadsheet/Calculation/Calculation.php
oleibman: SeriesSum was not part of this PR. I just noticed that I had neglected to update the function table for SeriesSum when I did implement its replacement. Since I am planning a follow-up PR to change funcWhatever, I would prefer to defer this suggestion until then.

src/PhpSpreadsheet/Calculation/Calculation.php
oleibman: Sin was not part of this PR. I just noticed that I had neglected to update the function table for Sin when I did implement its replacement. Since I am planning a follow-up PR to change funcWhatever, I would prefer to defer this suggestion until then.

src/PhpSpreadsheet/Calculation/Calculation.php
oleibman: Done.

MarkBaker (approved):
The additional unit tests covering invalid arguments and the exception cases are always good to have.

Once I've finished the Statistical and Financial classes, I'll do a sanity check over the whole exercise and see just what's left; and probably start rationalising all the different validation classes that I've created into a single validations class (or at least rationalise the basic validation methods... with throwing exceptions, they can easily be built on and expanded, so `validatePositiveInteger()` can call `validateInteger()`, etc. to build a more detailed validation set

PR #1990, opened 2021-04-06
Financial functions more rationalization

MarkBaker:
This is:

```
- [ ] a bugfix
- [ ] a new feature
- [X] refactoring
```

Checklist:

- [X] Code style is respected


MarkBaker:
I wish I could figure out what it is that phpstan is complaining about

oleibman:
I have a wild guess about phpstan. If nothing else, it shouldn't take long to test. Your code includes calls to getDateValue, which phpstan doesn't object to, and calls to, among others, validateBasis, which it doesn't like. Here's the docBlock for getDateValue:
```php
    /**
     * getDateValue.
     *
     * @param mixed $dateValue
     *
     * @return float Excel date/time serial value
     */
```

And here's the docBlock for validateBasis:
```php
    /**
     * @param mixed $basis
     */
```

Is it possible that adding the equivalent of the first 2 lines in the getDateValue docBlock to the validateBasis docBlock will make the problem go away? It shouldn't matter, but it's easy to test.

PowerKiKi:
PHPStan is complaining that your code is perfect :sunglasses: and that he cannot find the error that used to be there.

I rebased your branch on master (without any modifications), then I double checked that all PHPStan errors started with "Ignored error pattern", and finally I ran `vendor/bin/phpstan analyse --generate-baseline` to automatically update `phpstan-baseline.neon`. That last step was made in  https://github.com/PHPOffice/PhpSpreadsheet/pull/1990/commits/dbe1d2a3f154b16c0c1ea75db69ab6303b082577.

That file can be edited either manually or automatically, like I did. In both cases we should be careful to only remove things, and never add anything, in order to (one day) reach zero error at PHPStan max level and delete the file entirely.

You can read more about baseline over there: https://phpstan.org/user-guide/baseline

PowerKiKi:
As a side note, I bought a license for [PHPStan Pro](https://phpstan.org/blog/introducing-phpstan-pro), it is absolutely not required, and it's not even a "game changer", but it's slightly more friendly to fix huge batch of issues. You might want to consider it, or not. For me the _main_ reason was to support the developer, and I appreciate the slightly improved feedback loop (though it can be reproduced with CLI tools).

PowerKiKi:
And one last thing, I plan to refactor `Worksheet::getCell()`, in order to remove **a lot** of PHPStan issues, like so:

```diff
-    public function getCell($pCoordinate, $createIfNotExists = true): ?Cell
+    public function getCell($pCoordinate): Cell
```

PR #1992, opened 2021-04-07
Two Problems with Sample19

oleibman:
19_NamedRange.php was not changed to use absolute addressing when that was introduced to Named Ranges. Consequently, the output from this sample has been wrong ever since, for both Xls and Xlsx.

There was an additional problem with Xls. It appears that the Xls Writer Parser does not parse multiple concatenations using the ampersand operator correctly. So, `=B1+" "+B2` was parsed as `=B1+" "`. I believe that this is due to ampersand being treated as a condition rather than an operator. I attempted to resolve this, but my attempt is only partially successful. I tried to parse it as other operators by treating the left side as a 'term' rather than an 'expression'. In the resulting spreadsheet, the cell contains the correct calculated value, but it does not contain the actual formula. This is better than it had been, so I have left that change intact while I continue to research.

There are already more than ample tests for Named Ranges, so I did not add a new one for that purpose. However, I did add a new test for the Xls parser problem.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/samples/Basic/19_Namedrange.php
+++ b/samples/Basic/19_Namedrange.php
@@ -31,8 +31,8 @@ $spreadsheet->getActiveSheet()->setCellValue('A1', 'Firstname:')
 
 // Define named ranges
 $helper->log('Define named ranges');
-$spreadsheet->addNamedRange(new NamedRange('PersonName', $spreadsheet->getActiveSheet(), 'B1'));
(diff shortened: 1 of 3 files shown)


oleibman:
I tried to push a complete fix for the Xlsx Writer concatenation operator problem, but it didn't automatically merge. I will close this PR and submit another in its place tonight.

oleibman:
Closing, superseded by #2080.

PR #1998, opened 2021-04-13
Fix problem with UNC Path on Windows

### LINKED ISSUE #1144: Use of Built-In PHP Function is_readable() Problematic With UNC Path
Use of PHP's built-in function `is_readable()` (https://www.php.net/manual/en/function.is-readable.php) can be problematic with UNC (network) file paths.

It appears `is_readable()` is used going back to PHPOffice/PhpSpreadsheet 1.6, at least.  Here's a link to one such instance in 1.8:

https://github.com/PHPOffice/PhpSpreadsheet/blob/48ccdc271674070528f7252bfbb9c32a58df2813/src/PhpSpreadsheet/Shared/File.php#L140

There are a number of related, open PHP bug reports pertaining to `is_readable()`:

- https://bugs.php.net/bug.php?id=69834
- https://bugs.php.net/bug.php?id=68926
- https://bugs.php.net/bug.php?id=73543

For the moment, our workaround involves copying the network file to a local directory and then providing the local file path (i.e. instead of the network/UNC path) to PHPOffice/PhpSpreadsheet.

Pe46dro:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

Fix #1144 using the suggested workaround in [php bug 68926](https://bugs.php.net/bug.php?id=68926)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 - Fixed issue with Xlsx@listWorksheetInfo not returning any data
 - Fixed invalid arguments triggering mb_substr() error in LEFT(), MID() and RIGHT() text functions. [Issue #640](https://github.com/PHPOffice/PhpSpreadsheet/issues/640)
 - Fix for [Issue #1916](https://github.com/PHPOffice/PhpSpreadsheet/issues/1916) - Invalid signature check for XML files
+- Fix for [Issue #1144](https://github.com/PHPOffice/PhpSpreadsheet/issues/1144) - Could not open file for reading in UNC path
 
 ## 1.17.1 - 2021-03-01
 
--- a/src/PhpSpreadsheet/Shared/File.php
+++ b/src/PhpSpreadsheet/Shared/File.php
@@ -146,8 +146,15 @@ class File
             throw new InvalidArgumentException('File "' . $filename . '" does not exist.');
         }
 
-        if (!is_readable($filename)) {
-            throw new InvalidArgumentException('Could not open "' . $filename . '" for reading.');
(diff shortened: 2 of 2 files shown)


Pe46dro:
Maybe php-cs-fixer failed because `e` in fopen is not supported?

oleibman:
I am on Windows, I connect to some data using Samba, and I cannot duplicate your problem. This means only that there is something different between your environment and mine. However, it is a very tough sell to try make a change when (a) the problem can't be easily reproduced, and (b) the solution is difficult to comprehend (I don't really understand what mode "e" means and I'd be concerned if all my file opens had to use it). I think this problem needs to be resolved by PHP, not PhpSpreadsheet. They've had a lot of time to look at, and have yet to fix it, so I imagine that the ultimate solution is more complicated than yours.

All that being said, you can implement your solution on your own *without* having to make any changes to PhpSpreadsheet. What you need to do is to add a namespaced version of is_readable. It's pretty easy to implement. I'm uploading a file readable.zip with this message which contains a demonstration of this technique in action. Will that work for you? If not, what might it need to make it acceptable?
[readable.zip](https://github.com/PHPOffice/PhpSpreadsheet/files/6366663/readable.zip)

Pe46dro:
Hi, thanks for your response, I already made the override of `is_readable` in my code to bypass the problem.
I was able to reproduce the problem on my PC and on a server, I'm using a share with AD authentication maybe a share with Everyone don't have the problem (I didn't investigated further).
Since the problem should not workaround-ed PhpSpreadsheet but fixed by PHP Team, maybe the override solution should be written in documentation to prevent further PR regarding the problem.

By the way I used the `e` for POSIX runtime like `cygwin` or `MinGW`.
From php.net
> 'e' | Set close-on-exec flag on the opened file descriptor. Only available in PHP compiled on POSIX.1-2008 conform systems.

PR #1995, opened 2021-04-08
Improved Support for INDIRECT, ROW, and COLUMN Functions

oleibman:
This should address issues #1913 and #1993. INDIRECT had heretofore not supported an optional parameter intended to support addresses in R1C1 format which was introduced with Excel 2010. It also had not supported the use of defined names as an argument.

The unit tests for INDIRECT had used mocking, and were sorely lacking (tested only error conditions). They have been replaced with normal, and hopefully adequate, tests. This includes testing globally defined names, as well as locally defined names, both in and out of scope.

The test case in 1913 was too complicated for me to add as a unit test. The main impediments to it are now removed, and its complex situation, will I hope, be corrected with this fix.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Changes are covered by unit tests
- [x] Code style is respected
- [x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ ] CHANGELOG.md contains a short summary of the change
- [ ] Documentation is updated as necessary

Issue #1993, opened 2021-04-08
INDIRECT function not working with Defined names in excel


Lzolcsi:
This is:

```
- [X] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

I have an excel file that uses defined names and I use the INDIRECT function to reference them, but this throws an error when trying to access the calculated value of the cell.

### What is the expected behavior?
-return the correct value (300 in this case)

### What is the current behavior?
PhpSpreadsheet throws the following Exceptions:

PHP Fatal error:  Uncaught PhpOffice\PhpSpreadsheet\Calculation\Exception: Worksheet!B3 -> Invalid cell coordinate =B1 in /Users/zolcsi/www/szamlazo/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/Cell.php:272
Stack trace:
#0 /Users/zolcsi/www/szamlazo/bin/inc/modules/Project.php(5014): PhpOffice\PhpSpreadsheet\Cell\Cell->getCalculatedValue()
#1 /Users/zolcsi/www/szamlazo/bin/inc/modules/Project.php(155): Project->XlsGyartoSzallitoMegfeleltetes()
#2 /Users/zolcsi/www/szamlazo/bin/inc/common.php(48): Project->Execute()
#3 /Users/zolcsi/www/szamlazo/html/index.php(19): require_once('/Users/zolcsi/w...')
#4 {main}
  thrown in /Users/zolcsi/www/szamlazo/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/Cell.php on line 272

Fatal error: Uncaught PhpOffice\PhpSpreadsheet\Calculation\Exception: Worksheet!B3 -> Invalid cell coordinate =B1 in /Users/zolcsi/www/szamlazo/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/Cell.php:272
Stack trace:
#0 /Users/zolcsi/www/szamlazo/bin/inc/modules/Project.php(5014): PhpOffice\PhpSpreadsheet\Cell\Cell->getCalculatedValue()
#1 /Users/zolcsi/www/szamlazo/bin/inc/modules/Project.php(155): Project->XlsGyartoSzallitoMegfeleltetes()
[...]

Issue #1913, opened 2021-03-10
No Support For R1C1 Style Address in INDIRECT Function


hubobbb:
This is:

```
- [X ] a bug report


### What is the expected behavior?
fixed

### What is the current behavior?

save file error 
[...]

PR #2004, opened 2021-04-16
Improved Support for INDIRECT, ROW, and COLUMN Functions

oleibman:
This should address issues #1913 and #1993. INDIRECT had heretofore not supported an optional parameter intended to support addresses in R1C1 format which was introduced with Excel 2010. It also had not supported the use of defined names as an argument. This PR is a replacement for #1995, which is currently in draft status and which I will close in a day or two.

The ROW and COLUMN functions also should support defined names. I have added that, and test cases, with the latest push. ROWS and COLUMNS already supported it correctly, but there had been no test cases. Because ROW and COLUMN can return arrays, and PhpSpreadsheet does not support dynamic arrays, I left the existing direct-call tests unchanged to demonstrate those capabilities.

The unit tests for INDIRECT had used mocking, and were sorely lacking (tested only error conditions). They have been replaced with normal, and hopefully adequate, tests. This includes testing globally defined names, as well as locally defined names, both in and out of scope.

The test case in 1913 was too complicated for me to add as a unit test. The main impediments to it are now removed, and its complex situation will, I hope, be corrected with this fix.

INDIRECT can also support a reference of the form Sheetname!localName when localName on its own would be out of scope. That functionality is added. It is also added, in theory, for ROW and COLUMN, however such a construction is rejected by the Calculation engine before passing control to ROW or COLUMN. It might be possible to change the engine to allow this, and I may want to look into that later, but it seems much too risky, and not nearly useful enough, to attempt to address that as part of this change.

Several unusual test cases (leading equals sign, not-quite-as-expected name definition in file, complex indirection involving concatenation and a dropdown list) were suggested by @MarkBaker and are included in this request.

Three grandfathered phpstan errors were eliminated and removed from baseline.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Calculation/LookupRef/Helpers.php
```diff
@@ -0,0 +1,81 @@
...
+use PhpOffice\PhpSpreadsheet\Cell\AddressHelper;
+use PhpOffice\PhpSpreadsheet\Cell\Cell;
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Worksheet\Worksheet;
+
+class Helpers
+{
+    private static function convertR1C1(string &$cellAddress1, ?string &$cellAddress2, bool $a1): string
+    {
+        if (!$a1) {
```
MarkBaker, line 14: Can we use explicit typed comparisons rather than lazy comparisons, the addition of class constants like 
```php
public const CELLADDRESS_USE_A1 = true;
public const CELLADDRESS_USE_R1C1 = true;
```
and similarly within the `INDIRECT()` implementation

This make it very obvious to anybody reading the code

These constants can then be used in the `INDIRECT()` function implementation as well

src/PhpSpreadsheet/Calculation/LookupRef/Indirect.php
```diff
@@ -2,45 +2,75 @@
...
 use PhpOffice\PhpSpreadsheet\Calculation\Functions;
 use PhpOffice\PhpSpreadsheet\Cell\Cell;
 use PhpOffice\PhpSpreadsheet\Worksheet\Worksheet;
 
 class Indirect
 {
+    /**
+     * Determine whether cell address is in A1 (true) or R1C1 (false) format.
+     *
+     * @param null|bool|float|int|string $a1fmt
```
MarkBaker, line 16: If you're listing all these potential parameter values, it's best to use `mixed` (and in various other method docblocks as well)

src/PhpSpreadsheet/Calculation/LookupRef/Helpers.php
```diff
@@ -0,0 +1,81 @@
...
+    {
+        return strcasecmp($cellAddress1, $definedName) === 0 && ($scope === null || $scope === $sheet || $sheetTitle === $sheetName);
+    }
+
+    public static function extractCellAddresses(string $cellAddress, bool $a1, Spreadsheet $spreadsheet, Worksheet $sheet, ?string $sheetName = null): array
+    {
+        $cellAddress1 = $cellAddress;
+        $cellAddress2 = null;
+        $namedRanges = $spreadsheet->getNamedRanges();
+        foreach ($namedRanges as $namedRange) {
```
MarkBaker, line 44: The `resolveName()` method in `DefinedName` should handle a lot of this logic for you without needing a loop, including handling the scope checking... it returns either a null, or a `DefinedName` object

MarkBaker (changes requested):

src/PhpSpreadsheet/Calculation/LookupRef/Helpers.php
oleibman: Done.

src/PhpSpreadsheet/Calculation/LookupRef/Indirect.php
oleibman: Done.

src/PhpSpreadsheet/Calculation/LookupRef/Helpers.php
oleibman: This isn't possible without a change to resolveName, which I have now added to the PR. The specific case that resolveName had not been able to handle was a reference of the form sheetName!localName, where unqualified localName is out of scope for the referring cell. Excel permits this usage. Adding an optional sheetName parameter to resolveName does indeed simplify the code in Helpers, and doesn't make resolveName too complex.

Lzolcsi:
I'm not very familiar with Github PR-s, but am I right that this should be fixed and now we're just waiting for MarkBaker to review the changes that he requested? (no pressure, I'm just curious to know... :) )

oleibman:
Correct.

MarkBaker (approved):

Lzolcsi:
Thank you very much for you work!

PR #2006, opened 2021-04-17
MathTrig - Change Names of funcWhatever to evaluate

oleibman:
Per discussions while MathTrig was being broken up, this would help standardize the code. That idea was adopted partway through the breakup. This PR applies that standardization to the earlier efforts. A similar effort is required for DateTime; that will come later.

The only 2 remaining funcWhatevers in MathTrig are both in SUM, which required two different methods depending on whether or not string parameters were to be ignored. It seems appropriate to leave those method names non-standardized in order to require a decision about which is to be used if they are invoked internally.

I also changed the MatrixFunctions method names as part of this PR. This could create a conflict with PR #2005, even though I think the names I chose in this PR and the names in that one are identical. I can remove that part of this change if desired.

3 Phpstan grandfathered errors were eliminated as part of this change, and its baseline has changed accordingly.

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1746,21 +1746,6 @@ parameters:
 			path: src/PhpSpreadsheet/Calculation/TextData/Extract.php
 
 		-
-			message: "#^Parameter \\#1 \\$number of static method PhpOffice\\\\PhpSpreadsheet\\\\Calculation\\\\MathTrig\\\\Mround\\:\\:funcMround\\(\\) expects float, float\\|int\\|string given\\.$#"
-			count: 1
-			path: src/PhpSpreadsheet/Calculation/TextData/Format.php
-
-		-
-			message: "#^Parameter \\#1 \\$number of function floor expects float, float\\|int\\|string given\\.$#"
-			count: 1
-			path: src/PhpSpreadsheet/Calculation/TextData/Format.php
-
-		-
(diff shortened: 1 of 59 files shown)


MarkBaker:
Watch out for case-sensitivity in class/method names... we always try to use lower-camel-case for method names

oleibman:
There are too many conflicts with #2005. I am converting to draft while I figure out the best way to handle this. I usually handle  conflicts by hand, but that seems too unwieldy here. I could use a suggestion on what you feel the best approach would be.

I was not sure what your camel-case comment referred to. I thought all my name changes used 'evaluate'.  Possibly you were referring to the 3 matrix functions, which I changed to all lower case as part of this PR. At any rate, please let me know if there are other names that still need attention.

oleibman:
Aha - found uppercase Multiply, Inverse, Identity in my test members. I will resolve those now.

oleibman:
I seem to have done something wrong - Github is "waiting for status to be reported". I will start over.

MarkBaker:
PHPStorm's merge tool is generally pretty good; but if it doesn't help you, I can probably do them a lot more quickly because I know exactly what I'm looking at

oleibman:
Thank you for the tip. I'll look into it so that I'm better prepared next time this happens.

oleibman:
Closing, replaced by #2008.

PR #2008, opened 2021-04-18
MathTrig - Change Names of funcWhatever to evaluate

oleibman:
Per discussions while MathTrig was being broken up, this would help standardize the code. That idea was adopted partway through the breakup. This PR applies that standardization to the earlier efforts. A similar effort is required for DateTime; that will come later. This PR replaces #2006.

The only 2 remaining funcWhatevers in MathTrig are both in SUM, which required two different methods depending on whether or not string parameters were to be ignored. It seems appropriate to leave those method names non-standardized in order to require a decision about which is to be used if they are invoked internally.

3 Phpstan grandfathered errors were eliminated as part of this change, and its baseline has changed accordingly.

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
```

Checklist:

- [x] Code style is respected


oleibman:
@MarkBaker and @PowerKiKi - this change seems a little snakebit. Scrutinizer fails because it received a segfault in phpcs? Even though Github phpcs runs fine? Seems inconceivable that my change has anything to do with the Scrutinizer problem, but, whether or not that's true, I have no idea how to move forward.

MarkBaker (approved):

PR #2010, opened 2021-04-19
correct type annotation

crishoj:
This is:
- [x] a tiny fix of a type annotation which makes phpstan complain

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -1033,7 +1033,7 @@ class Worksheet implements IComparable
     /**
      * Get highest worksheet column.
      *
-     * @param string $row Return the data highest column for the specified row,
+     * @param int $row Return the data highest column for the specified row,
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -1033,7 +1033,7 @@ public function setProtection(Protection $pValue)
     /**
      * Get highest worksheet column.
      *
-     * @param string $row Return the data highest column for the specified row,
+     * @param int $row Return the data highest column for the specified row,
```
MarkBaker, line 1036: Be careful when making changes like this: the argument passed in could be a string containing  numeric value... but it is also nullable

MarkBaker (changes requested):

src/PhpSpreadsheet/Worksheet/Worksheet.php
crishoj: True. How about `string|int|null`?

PR #2020, opened 2021-04-23
MathTrig - Fix Phpstan Accomodations

oleibman:
This should be the last of my mass changes to MathTrig. All he Phpstan violations found in baseline which  are part of MathTrig are now fixed and removed from baseline. There were about 20 of these.

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
```

Checklist:

- [x] Code style is respected


oleibman:
Yipe. Phpstan says (correctly) that the  str_split function in Arabic can only return an array, so complains if you test for false. Scrutinizer incorrectly thinks that it can return true. Trying to figure out how to mollify both.

Issue #860, opened 2019-01-24
Read XLSX with namespaces in their XML


JulienChavee:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Xlsx that contains namespaces in their XML are correctly loaded

### What is the current behavior?
The array of data after reading the xlsx is empty

### What are the steps to reproduce?

I'm using the following code to read this file: [example.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/2791984/example.xlsx)

```php
<?php
		$reader = IOFactory::createReaderForFile($filePath);
		$spreadsheet=$reader->load($filePath);
		$sheet=$spreadsheet->getSheet(0);
                print_r($sheet->toArray());
?>
```

### Which versions of PhpSpreadsheet and PHP are affected?

* PhpSpreadsheet 1.6
* PHP >= 5.6 

### Solution

As requested in #206, I tried to find a better solution instead of just `str_replace`.
As I'm not an expert in XLSX and XML, I don't know if it works correctly for all cases.

All tests I've done are OK with this solution, so can you take a look this and give me your review? If it's seems good, I can submit a pull request with this.

``` php
private function removeXMLNamespaces($xml)
{
	$nameSpaceToRemove = array('x','d');

		if(!empty($xml))
		{
			$simpleXMLElement=new \SimpleXMLElement($xml);
			$domSimpleXMLElement=dom_import_simplexml($simpleXMLElement);

			if(!is_null($domSimpleXMLElement))
			{
				$dom=new \DOMDocument('1.0');
				$domSimpleXMLElement=$dom->importNode($domSimpleXMLElement,true);
				$domSimpleXMLElement=$dom->appendChild($domSimpleXMLElement);

				$element=$dom->childNodes->item(0);

				foreach($simpleXMLElement->getDocNamespaces() as $name=>$uri)
				{
					if(in_array($name, $nameSpaceToRemove))
						$element->removeAttributeNS($uri,$name);
				}

				return $dom->saveXML();
			}
			else
				return $xml;
		}
		else
			return $xml;
}
```

JulienChavee:
Just up this issue else it'll be closed.
If you need any help or something, let me know!

LouiseGs:
[...]

PR #2022, opened 2021-04-25
Gnumeric Better Namespace Handling

oleibman:
There have been a number of issues concerning the handling of legitimate but unexpected namespace prefixes in Xlsx spreadsheets created by software other than Excel and PhpSpreadsheet/PhpExcel.I have studied them, but, till now, have not had a good idea on how to act on them. A recent comment https://github.com/PHPOffice/PhpSpreadsheet/issues/860#issuecomment-824926224 in issue #860 by @IMSoP has triggered an idea about how to proceed.

Although the issues exclusively concern Xlsx format, I am starting out by dealing with Gnumeric. It is simpler and smaller than Xlsx, and, more important, already has a test for an unexpected prefix, since, at some point, it changed its generic prefix from gmr to gnm. I added support and a test for that some time ago, but almost certainly not in the best possible manner. The code as changed for this PR seems simpler and less kludgey, both for that exceptional case as well as for normal handling.

My hope is that this change can be a template for similar Reader changes for Xml, Ods, and, especially, Xlsx.

All grandfathered Phpstan issues with Gnumeric are fixed and eliminated from baseline as part of this change.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Reader/Gnumeric.php
```diff
@@ -77,16 +82,18 @@ public function canRead($pFilename)
...
+            }
         }
 
         return $data == chr(0x1F) . chr(0x8B);
     }
 
     private static function matchXml(string $name, string $field): bool
     {
-        return 1 === preg_match("/^(gnm|gmr):$field$/", $name);
+        return 1 === preg_match("/:$field$/", $name);
```
IMSoP, line 96: I think the correct test here is against `namespaceUri` (to be sure it's not something with the same name from a different namespace) and `localName` (which is the name of the element without any prefix) instead of against `name` (which might or might not have a prefix, depending on the `xmlns` attributes in scope), e.g.

```php
private static function matchXml(XMLWriter $node, string $expectedLocalName) {
     return
          $xml->namespaceUri == self::NAMESPACE_GNM
          && $xml->localName == $expectedLocalName
          && $xml->nodeType == XMLReader::ELEMENT;
}
```

At a glance, the extra check against nodeType seems to be present everywhere this is used, so could be moved here, e.g. line 145 would become just:

```php
if (self::matchXml($xml, 'Sheet')) {
```

IMSoP (commented):
I'm glad my comment was helpful (and hope it didn't sound too critical; XML namespaces _are_ confusing). :)

I did a quick experiment with XlsxReader the other day, and found that a number of places use `$foo['bar']` to access attributes, and fail if you select the namespace explicitly. They'll need to be tracked down and either converted to `$foo->attributes()->bar`, or have `$foo = $foo->attributes()` added above. 

(The reason has to do with an oddity in the XML Namespaces standard, where an _element_ with no prefix is in the default namespace, but an _attribute_ with no prefix has no defined meaning.)

src/PhpSpreadsheet/Reader/Gnumeric.php
oleibman: Thank you both for the original suggestion, and the new suggestion for how to deal with XMLReader. I have adapted your suggested code and included it in the latest push, and will try to be alert for the attribute problem when I get around to looking at Xlsx.

src/PhpSpreadsheet/Reader/Gnumeric.php
```diff
@@ -77,16 +82,20 @@ public function canRead($pFilename)
...
         }
 
         return $data == chr(0x1F) . chr(0x8B);
     }
 
-    private static function matchXml(string $name, string $field): bool
+    private static function matchXml(XMLReader $xml, string $expectedLocalName): bool
     {
-        return 1 === preg_match("/^(gnm|gmr):$field$/", $name);
+        return $xml->namespaceURI == self::NAMESPACE_GNM
```
MarkBaker, line 96: Should probably be using strict comparisons here, even if it means casting the xml node values explicitly to strings

src/PhpSpreadsheet/Reader/Gnumeric.php
```diff
@@ -536,35 +566,54 @@ private function processColumnWidths(SimpleXMLElement $sheet, int $maxCol): void
...
             }
             while ($c <= $maxCol) {
-                $this->spreadsheet->getActiveSheet()->getColumnDimension(Coordinate::stringFromColumnIndex($c + 1))->setWidth($defaultWidth);
+                $this->setColumnWidth($c, $defaultWidth);
                 ++$c;
             }
         }
     }
 
+    private function setRowHeight(int $r, float $defaultHeight): void
```
MarkBaker, line 585: Can we use a more meaningful argument name than `$r`, e.g. `$rowNumber`? And expand abbreviated variable names like `$rowDim` to `$rowDimension`?

We should probably do the same for `$r` and `$c` in the main loop, but I know you inherited that, and it isn't really in scope for a PR that's intended to test the new approach to XML namespacing, so that might be a later PR just to make variable names more explicit

src/PhpSpreadsheet/Reader/Gnumeric.php
```diff
@@ -639,27 +691,29 @@ private static function addStyle2(array &$styleArray, string $key1, string $key,
...
 
-        self::addStyle($styleArray, 'borderStyle', $borderAttributes['Style']);
+            self::addStyle($styleArray, 'borderStyle', $borderAttributes['Style']);
+        }
 
         return $styleArray;
     }
 
-    private function parseRichText($is)
+    private function parseRichText(string $is): RichText
```
MarkBaker, line 708: I do love the fact that we can use this exercise as an excuse to apply typehinting as well... always try to leave the codebase in a better state than you found it

src/PhpSpreadsheet/Reader/Gnumeric/PageSetup.php
```diff
@@ -68,7 +63,7 @@ public function sheetMargins(SimpleXMLElement $sheet): self
 
     private function buildMarginSet(SimpleXMLElement $sheet, array $marginSet): array
     {
-        foreach ($sheet->PrintInformation->Margins->children($this->gnm, true) as $key => $margin) {
+        foreach ($sheet->PrintInformation->Margins->children(Gnumeric::NAMESPACE_GNM) as $key => $margin) {
```
MarkBaker, line 66: Do we have any examples of the older `gmr` namespaced files that would allow us to test with them as well?

MarkBaker (commented):
With all tests passing, it looks like the new approach to namespacing i working remarkably well

src/PhpSpreadsheet/Reader/Gnumeric/PageSetup.php
oleibman: Yes, I added a very small old.gnumeric with gmr namespacing to samples/templates a while back. It has been used ever since in tests in IOFactoryTest and GnumericLoadTest.
I will push the changes you requested tonight.
FYI, Xlsx is proving to be far more difficult than Gnumeric. I don't expect to have anything ready for weeks.

src/PhpSpreadsheet/Reader/Gnumeric.php
oleibman: Done. I usually reflexively do this, but this particular section of code was copy-pasted without enough thought.

src/PhpSpreadsheet/Reader/Gnumeric.php
oleibman: I have renamed 'c' and 'r' throughout to 'whichColumn' and 'whichRow', in order to avoid conflicting with existing 'row' and 'column' variables. I suppose rowNumber and columnLetter might have been better choices. Also changed rowDim->rowDimension and colDum->columnDimension.

src/PhpSpreadsheet/Reader/Gnumeric.php
oleibman: I took the opportunity to change a few other comparisons to strict, when I was able to convince myself that was correct.

src/PhpSpreadsheet/Reader/Gnumeric.php
oleibman: I think that was one of the grandfathered Phpstan exceptions. That baseline file is pretty useful.

MarkBaker:
@IMSoP Critical, not nearly as critical as the people that demand that we solve the namespace problems without ever making any suggestions about how to resolve it... thanks for the guidance on how it should be done

PR #2023, opened 2021-04-25
BREAKING `Worksheet::getCellByColumnAndRow()` cannot return null anymore

PowerKiKi:
`Worksheet::getCellByColumnAndRow()` used to optionally return null if passed a
second argument. This second argument was removed entirely and the
method always returns a Cell (possibly creating it if needed).

This make the API more predictable and easier to do static analysis
with tools such as PHPStan.

If you relied on that second parameter, you should instead use the
`Worksheet::cellExistsByColumnAndRow()` before calling `getCellByColumnAndRow()`.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -4811,11 +4811,6 @@ parameters:
 			path: src/PhpSpreadsheet/ReferenceHelper.php
 
 		-
-			message: "#^Cannot call method setXfIndex\\(\\) on PhpOffice\\\\PhpSpreadsheet\\\\Cell\\\\Cell\\|null\\.$#"
-			count: 1
-			path: src/PhpSpreadsheet/ReferenceHelper.php
-
-		-
 			message: "#^Parameter \\#1 \\$columnIndex of static method PhpOffice\\\\PhpSpreadsheet\\\\Cell\\\\Coordinate\\:\\:stringFromColumnIndex\\(\\) expects int, float\\|int given\\.$#"
 			count: 1
 			path: src/PhpSpreadsheet/ReferenceHelper.php
@@ -6251,16 +6246,6 @@ parameters:
 			path: src/PhpSpreadsheet/Style/Style.php
 
 		-
-			message: "#^Cannot call method getXfIndex\\(\\) on PhpOffice\\\\PhpSpreadsheet\\\\Cell\\\\Cell\\|null\\.$#"
-			count: 2
-			path: src/PhpSpreadsheet/Style/Style.php
-
-		-
-			message: "#^Cannot call method setXfIndex\\(\\) on PhpOffice\\\\PhpSpreadsheet\\\\Cell\\\\Cell\\|null\\.$#"
-			count: 1
-			path: src/PhpSpreadsheet/Style/Style.php
-
-		-
 			message: "#^Result of && is always true\\.$#"
 			count: 1
 			path: src/PhpSpreadsheet/Worksheet/AutoFilter.php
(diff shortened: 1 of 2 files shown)


PowerKiKi:
Merged as 2e39ece3739fad95a0fda0dbf9dfc6c757a585bb

MarkBaker:
This, and the previous similar change to `getCell()` are having a adverse side-affect on the calculation engine, in that formulae which reference a cell that doesn't exist will now create that cell.... it doesn't affect the result of any calculation, because a null value is now derived from the new cell itself rather than from the null return, but it does affect execution speed and memory usage for formulae that use row or column references (e.g. `=SUM(A:A)` or `=SUM(1:2`) because now an entire column or row of new cells will be created.... and that will also affect any file that is subsequently saved.

I'll need to make some calculation engine changes to modify that behaviour, and test for the existence of the cell before accessing it. Fortunately, I'm looking at that part of the calculation code at the moment (which is how I noticed the issue).

PowerKiKi:
That is unfortunate and totally unexpected since, while my changes are quite a significant logical change, they are actually very localized.

I my first commit the calculation has only very minimal change, which happen to be a simplification: https://github.com/PHPOffice/PhpSpreadsheet/commit/d85eaacfa339341932d904e62b53ba5e607784e2#diff-34836315744a536ab78241f537e5dbdedc2919bbd29d85893dc7f7aa4bbeae1e

My second commit is even smaller as the second argument was actually never used anywhere in our code: https://github.com/PHPOffice/PhpSpreadsheet/commit/2e39ece3739fad95a0fda0dbf9dfc6c757a585bb

And the final commit is a bit bigger but does not touch calculation at all: https://github.com/PHPOffice/PhpSpreadsheet/commit/4e2259c13582ef705f02f279b3c4c506a63255a4

So basically I didn't touch calculation. Maybe there is a side-effect on the few type casting I added ?

Since `getCellOrNull()` was never meant to become public, maybe we can fix this another way. Were you able to pinpoint what method call produce the increase of memory usage ?

PowerKiKi:
In fact, I just tested each commit. Before that commit and the commit itself, and PHPUnit always report exactly the same memory used for before and after. I would expect this would show at least a small difference since we have quite a few tests.

Do you have a code sample to reproduce the memory increase ?

# d85eaacfa339341932d904e62b53ba5e607784e2

Before: 

```
Time: 1.74 minutes, Memory: 206.50 MB

OK, but incomplete, skipped, or risky tests!
Tests: 9800, Assertions: 13821, Skipped: 3, Incomplete: 2.
```

After:

```
Time: 1.04 minutes, Memory: 206.50 MB

OK, but incomplete, skipped, or risky tests!
Tests: 9800, Assertions: 13821, Skipped: 3, Incomplete: 2.
```



# 2e39ece3739fad95a0fda0dbf9dfc6c757a585bb

Before: 

```
Time: 1.11 minutes, Memory: 210.50 MB

OK, but incomplete, skipped, or risky tests!
Tests: 9903, Assertions: 13936, Skipped: 3, Incomplete: 2.
```

After:

```
Time: 49.04 seconds, Memory: 210.50 MB

OK, but incomplete, skipped, or risky tests!
Tests: 9903, Assertions: 13936, Skipped: 3, Incomplete: 2.
```

# 4e2259c13582ef705f02f279b3c4c506a63255a4

Before: 

```
Time: 48.86 seconds, Memory: 210.50 MB

OK, but incomplete, skipped, or risky tests!
Tests: 9903, Assertions: 13936, Skipped: 3, Incomplete: 2.

```

After: 

```
Time: 58.73 seconds, Memory: 210.50 MB

OK, but incomplete, skipped, or risky tests!
Tests: 9903, Assertions: 13938, Skipped: 3, Incomplete: 2.
```

MarkBaker:
It was an indirect effect, triggered when a formula references a cell that doesn't exist. All tests still passed, because the calculation still evaluated correctly, but it was creating new cells whenever they were referenced in a formula.
This was particularly highlighted when I started looking at row/column range formulae, where a formula like `=SUM(A:A)` was referencing ever row in column A, and the resulting slowdown in execution time. Row/column ranges were never officially supported by the Calculation Engine, but that's never stopped people from using them. It's also why we didn't have tests for them.

I was looking at providing that support, so wrote some test cases that do use row/column ranges; and that's when I noticed the affect.

It's no longer a problem for the changes I've made to the way the calculation engine works with row/column ranges, providing proper support, and limiting the ranges to the actual active cell range; and ensured that proper checks are made for the existence of a cell before accessing it. These changes are covered by those unit tests, and I can see that they don't use excessive memory any more.

So all's good!

Issue #2029, opened 2021-04-27
Error "Invalid cell coordinate A-1" reading HTML file


danmodini:
This is:

```
- [*] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Open flawessly the html file

### What is the current behavior?
Gives an error  "Invalid cell coordinate A-1"

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

$objPHPExcel = PhpSpreadsheet\IOFactory::load('Nomina_30539141909_202102_000.xls'); // it is a html file

### Which versions of PhpSpreadsheet and PHP are affected?
1.17.1
[Nomina_30539141909_202102_000.xls](https://github.com/PHPOffice/PhpSpreadsheet/files/6386224/Nomina_30539141909_202102_000.xls)

I find a personal solution limiting the value of $row to 1 in line 473:

           if ($this->tableLevel > 1 && $row>1) {
                --$row;
            }
this is because the hyphen in the error is a minus sign. I dont know if this workaround breakes something else...
Thank you at advance

Daniel

PR #2032, opened 2021-04-28
Fix for Issue 2029 (Invalid Cell Coordinate A-1)

oleibman:
Fix for #2029. When Html Reader encounters an embedded table, it tries to shift it up a row. It obviously should not attempt to shift it above row 1. @danmodini reported the problem, and suggests the correct solution. This PR implements that and adds a test case.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Reader/Html.php
```diff
@@ -878,14 +878,14 @@ private function applyInlineStyle(&$sheet, $row, $column, $attributeArray): void
 
                 case 'width':
                     $sheet->getColumnDimension($column)->setWidth(
-                        (float) str_replace('px', '', $styleValue)
+                        (float) str_replace(['px', 'pt'], '', $styleValue)
```
MarkBaker, line 881: I have half of a helper class that covers a wide range of the UoM that can be applied in html styles, with conversion factors for import to Spreadsheet styles... but this will work well enough for the moment... it's the relative units that are a real annoyance, because we can't adjudge `50%` accurately, as we don't know `50%` of what? I'll do that as a separate PR at some point though

MarkBaker (approved):

src/PhpSpreadsheet/Reader/Html.php
oleibman: That would be really helpful, especially because Excel widths and heights aren't well documented. They aren't pixels, or points, or twips, or ... I know the default column width in Excel units and in pixels, but I haven't had time to establish whether there is a linear relationship between them. The Html writer and reader probably make things too wide or too narrow because of this.

oleibman: Didn't really mean non-linear above - it is linear, but it depends on the font. Which may be too sophisticated to try to emulate (as with your 50% comment). But I think it translates better to expressing the unit as ems rather than points or pixels.

PR #2036, opened 2021-04-29
Mpdf Caching

oleibman:
Mpdf caches font data in a temporary directory. When running the test suite, it uses different directories for samples vs. normal tests. One is sufficient. This will slightly reduce disk usage and overhead when running test suite.

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] test suite performance
```

Checklist:

- [x] Code style is respected


oleibman:
I haven't yet merged any of my own changes, but two weeks with no comments, combined with a really simple change, says this should be the one.

PR #2048, opened 2021-04-30
Fix to export csv file with numbers in an element

joern19:
When trying to export an Spreadsheet with an field only containing numbers, the following error occurs:
PHP Fatal error:  Uncaught TypeError: str_replace(): Argument #3 ($subject) must be of type array|string, int given
Fixed by wrapping the element with strval

This is:

```
- [ ] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Writer/Csv.php
+++ b/src/PhpSpreadsheet/Writer/Csv.php
@@ -367,7 +367,7 @@ class Csv extends BaseWriter
                 if (!$this->enclosureRequired && strpbrk($element, "$delimiter$enclosure\n") === false) {
                     $enclosure = '';
                 } else {
-                    $element = str_replace($enclosure, $enclosure . $enclosure, $element);
+                    $element = str_replace($enclosure, $enclosure . $enclosure, strval($element));
                 }
             }
             // Add enclosed string


oleibman:
I am not able to duplicate your error with the following code:
```php
<?php
declare(strict_types=1);
require __DIR__ . '/PhpSpreadsheet' . '/vendor/autoload.php';
$obj = new PhpOffice\PhpSpreadsheet\Spreadsheet();
$sheet0 = $obj->getActiveSheet();
$sheet0->setCellValue('A1', 2);
$sheet0->setCellValue('A2', 3);
$sheet0->setCellValue('B1', 4);
$sheet0->setCellValue('B2', 5);
$out = 'strict4.csv';
$writer = new \PhpOffice\PhpSpreadsheet\Writer\Csv($obj);
$writer->setEnclosureRequired(true);
$writer->save($out);
echo "$out has been saved\n";
```
I stuck a var_dump at the code where you saw a failure, and verified that `$element` was an int all 4 times. I tried this with several versions of PHP, including 8.0.0. Can you share your source code and your PHP version to trigger this problem? Also, which version of PhpSpreadsheet, and is your strval change the only change you made to Writer/Csv?

joern19:
Now I feel stupid. I have added a small feature to the csvwriter and also added `declare(strict_types=1);` because we add that to all PHP classes. Sorry for the inconvenience caused

Issue #984, opened 2019-05-14
Negative matching of text in conditional formatting


naitsirch:
This is:

```
- [ ] a bug report
- [x] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

I want some styling to be applied if a cell does NOT contain a specified string.

### What is the current behavior?

Currently it is not supported.

### What are the steps to reproduce?

```php
<?php
use PhpOffice\PhpSpreadsheet\Style\Conditional;
use PhpOffice\PhpSpreadsheet\Style\Color;

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

$conditional = new Conditional();
$conditional->setConditionType(Conditional::CONDITION_CONTAINSTEXT); // == 'containsText'
$conditional->setOperatorType(Conditional::OPERATOR_NOTCONTAINS); // == 'notContains'
$conditional->setText('published');
$conditional->getStyle()->getFont()->getColor()->setARGB(Color::COLOR_RED);
$conditional->getStyle()->getFont()->setBold(true);

$sheet = $spreadsheet->getActiveSheet();
$cell = $sheet->getCellByColumnAndRow(1, 1);
$cell->setValue('open');
$cell->getStyle()->setConditionalStyles([$this->getConditionalColorForStatus()]);
```

### Which versions of PhpSpreadsheet and PHP are affected?

All.

### Explanation

In the above code example I want to style the text in red and bold if the cell does NOT contain the text "published". But this is not possible. At least if you generate an XSLX file.

The OOXML requires the conditional formatting for this case to be defined like this:

```xml
	<conditionalFormatting sqref="A1:A1">
		<cfRule type="notContainsText" dxfId="0" priority="3" operator="notContains" text="published">
			<formula>ISERROR(SEARCH("published",A1))</formula>
		</cfRule>
	</conditionalFormatting>
```

But there is no condition type available to get `notContainsText`. So this has to be added.

oleibman:
Fixed by PR #2049 in May 2021.

PR #2049, opened 2021-04-30
#984 Add support notContainsText for conditional styles in xlsx

xandros15:
This is:

- [ ] a bugfix
- [x] a new feature

Checklist:

- [x] Code style is respected

Xlsx support notContainsText type of condition that wasn't implemented in PhpSpreadsheet. We could workaround with expression and `<>` operator, but that would required cords in condition. Now can works without cords.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
@@ -634,15 +634,21 @@ class Worksheet extends WriterPart
 
                     self::writeAttributeif(
                         $objWriter,
-                        ($conditional->getConditionType() == Conditional::CONDITION_CELLIS || $conditional->getConditionType() == Conditional::CONDITION_CONTAINSTEXT)
-                        && $conditional->getOperatorType() != Conditional::OPERATOR_NONE,
+                        in_array($conditional->getConditionType(), [
+                            Conditional::CONDITION_CELLIS,
+                            Conditional::CONDITION_CONTAINSTEXT,
+                            Conditional::CONDITION_NOTCONTAINSTEXT,
+                        ]) && $conditional->getOperatorType() != Conditional::OPERATOR_NONE,
                         'operator',
                         $conditional->getOperatorType()
                     );
 
                     self::writeAttributeIf($objWriter, $conditional->getStopIfTrue(), 'stopIfTrue', '1');
 
-                    if ($conditional->getConditionType() == Conditional::CONDITION_CONTAINSTEXT) {
+                    if (in_array($conditional->getConditionType(), [
+                        Conditional::CONDITION_CONTAINSTEXT,
+                        Conditional::CONDITION_NOTCONTAINSTEXT,
+                    ])) {
                         self::writeTextCondElements($objWriter, $conditional, $cellCoordinate);
                     } else {
                         self::writeOtherCondElements($objWriter, $conditional, $cellCoordinate);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 - Improved support for Row and Column ranges in formulae [Issue #1755](https://github.com/PHPOffice/PhpSpreadsheet/issues/1755) [PR #2028](https://github.com/PHPOffice/PhpSpreadsheet/pull/2028)
 - Implemented the CHITEST(), CHISQ.DIST() and CHISQ.INV() and equivalent Statistical functions, for both left- and right-tailed distributions.
 - Support for ActiveSheet and SelectedCells in the ODS Reader and Writer. [PR #1908](https://github.com/PHPOffice/PhpSpreadsheet/pull/1908)
+- Support for notContainsText Conditional Style in xlsx [Issue #984](https://github.com/PHPOffice/PhpSpreadsheet/issues/984)
 
 ### Changed
 
--- a/src/PhpSpreadsheet/Style/Conditional.php
+++ b/src/PhpSpreadsheet/Style/Conditional.php
@@ -15,6 +15,7 @@ class Conditional implements IComparable
     const CONDITION_CONTAINSBLANKS = 'containsBlanks';
     const CONDITION_NOTCONTAINSBLANKS = 'notContainsBlanks';
     const CONDITION_DATABAR = 'dataBar';
+    const CONDITION_NOTCONTAINSTEXT = 'notContainsText';
 
     // Operator types
     const OPERATOR_NONE = '';
--- a/tests/PhpSpreadsheetTests/Writer/Xlsx/ConditionalTest.php
+++ b/tests/PhpSpreadsheetTests/Writer/Xlsx/ConditionalTest.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace PhpOffice\PhpSpreadsheetTests\Writer\Xlsx;
+
+use PhpOffice\PhpSpreadsheet\Settings;
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Style\Conditional;
+use PhpOffice\PhpSpreadsheet\Style\Fill;
+use PhpOffice\PhpSpreadsheet\Writer\Xlsx;
+use PhpOffice\PhpSpreadsheetTests\Functional\AbstractFunctional;
+
+class ConditionalTest extends AbstractFunctional
+{
+    /**
+     * @var int
+     */
+    private $prevValue;
+
+    protected function setUp(): void
+    {
+        $this->prevValue = Settings::getLibXmlLoaderOptions();
+
+        // Disable validating XML with the DTD
+        Settings::setLibXmlLoaderOptions($this->prevValue & ~LIBXML_DTDVALID & ~LIBXML_DTDATTR & ~LIBXML_DTDLOAD);
+    }
+
+    protected function tearDown(): void
+    {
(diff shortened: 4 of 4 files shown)


oleibman:
Can you make your embedded arrays into class constants? This should reduce overhead, and may make it easier to avoid phpcs problems.

MarkBaker:
Also, if you're providing support for this in the Xlsx Writer, it's probably a good idea to provide support for it in the Reader as well

And check that having support for that new condition in the Xlsx Writer doesn't create errors for Xls and Ods Writers if the condition is used in a file that's being saved to one of those formats

xandros15:
> Can you make your embedded arrays into class constants? This should reduce overhead, and may make it easier to avoid phpcs problems.

Added.

Imho is less readable and make a more effort to add new condition.

> Also, if you're providing support for this in the Xlsx Writer, it's probably a good idea to provide support for it in the Reader as well

Added.

> And check that having support for that new condition in the Xlsx Writer doesn't create errors for Xls and Ods Writers if the condition is used in a file that's being saved to one of those formats

I don't know anything about xls and ods formats. Any tips how I can do test for this?

MarkBaker:
> And check that having support for that new condition in the Xlsx Writer doesn't create errors for Xls and Ods Writers if the condition is used in a file that's being saved to one of those formats

> I don't know anything about xls and ods formats. Any tips how I can do test for this?

A simple manual test - create a sample spreadsheet using `notContainsText`, (either in code, or loading an xlsx file using that CF option) and writing that spreadsheet to all appropriate formats (I use xlsx, xls, ods and html) looking for any errors in CLI output; then trying to open those files in Excel, OpenOffice and a browser... that should be enough to identify if there are any problems, and fixing them if they do occur. If there are no errors; then just confirm in the comments here.

xandros15:
Code:
```php
<?php

require_once __DIR__ . '/vendor/autoload.php';

use PhpOffice\PhpSpreadsheet\Style\Conditional;
use PhpOffice\PhpSpreadsheet\Style\Fill;

$worksheet->fromArray([
    ['A', 'B', 'C', 'D'],
]);
$condition = new Conditional();
$condition->setConditionType(Conditional::CONDITION_NOTCONTAINSTEXT);
$condition->setOperatorType(Conditional::OPERATOR_NOTCONTAINS);
$condition->setText("C");
$condition->getStyle()->applyFromArray([
    'fill' => [
        'color' => ['argb' => 'FFFFC000'],
        'fillType' => Fill::FILL_SOLID,
    ],
]);
$worksheet->setConditionalStyles('A1:A5', [$condition]);

(new \PhpOffice\PhpSpreadsheet\Writer\Xls($spreadsheet))->save('test.xls');
(new \PhpOffice\PhpSpreadsheet\Writer\Ods($spreadsheet))->save('test.ods');
(new \PhpOffice\PhpSpreadsheet\Writer\Html($spreadsheet))->save('test.html');
```

No errors in cli.

Checked xls and html. No errors.

I will check later ods.

[test.ods](https://github.com/PHPOffice/PhpSpreadsheet/files/6412362/test.ods)
[test.xls](https://github.com/PHPOffice/PhpSpreadsheet/files/6412363/test.xls)
//github doesn't support html as link

src/PhpSpreadsheet/Reader/Xlsx/ConditionalStyles.php
```diff
@@ -45,6 +45,7 @@ private function readConditionalStyles($xmlSheet)
                         || (string) $cfRule['type'] == Conditional::CONDITION_CONTAINSTEXT
                         || (string) $cfRule['type'] == Conditional::CONDITION_CONTAINSBLANKS
                         || (string) $cfRule['type'] == Conditional::CONDITION_NOTCONTAINSBLANKS
+                        || (string) $cfRule['type'] == Conditional::CONDITION_NOTCONTAINSTEXT
```
oleibman, line 48: Scrutinizer thinks this change makes the code too complex, when it didn't think it was too complex before. That will prevent your change from being merged. It's one of the reasons why I recommended the use of a constant class array earlier - you can reduce 6 separate tests in this statement to 1, and Scrutinizer will be happy with that.

xandros15:
Don't know what constant class array you mean. I could add some methods to Conditional model to validate it:

```php

    public function hasOperator(): bool
    {
        return $this->operatorType != Conditional::OPERATOR_NONE && in_array($this->conditionType, [
            Conditional::CONDITION_CELLIS,
            Conditional::CONDITION_CONTAINSTEXT,
            Conditional::CONDITION_NOTCONTAINSTEXT,
        ]);
    }


    public function isTextCondition(): bool
    {
        return in_array($this->conditionType, [
            Conditional::CONDITION_CONTAINSTEXT,
            Conditional::CONDITION_NOTCONTAINSTEXT,
        ]);
    }

    /**
     * Verify if param is valid condition type
     *
     * @param string $type
     *
     * @return bool
     */
    public static function isValidConditionType(string $type): bool
    {
        return in_array($type, [
            self::CONDITION_CELLIS,
            self::CONDITION_CONTAINSBLANKS,
            self::CONDITION_CONTAINSTEXT,
            self::CONDITION_DATABAR,
            self::CONDITION_EXPRESSION,
            self::CONDITION_NONE,
            self::CONDITION_NOTCONTAINSBLANKS,
            self::CONDITION_NOTCONTAINSTEXT,
        ]);
    }
```

oleibman:
Here's what your code would look like with a constant class array (my formatting may or may not be a little off):
```php
    private const HAS_OPERATOR = [
        Conditional::CONDITION_CELLIS,
        Conditional::CONDITION_CONTAINSTEXT,
        Conditional::CONDITION_NOTCONTAINSTEXT,
    ];

    /**
     * Verify if param is valid condition type
     */
    public static function isValidConditionType(string $type): bool
    {
        return in_array($type, self::HAS_OPERATOR);
    }
```
This way you don't have to allocate the array every time you invoke this method. I also removed the typehints from the DocBlock since you are explicitly type-hinting. You might be able to just invoke in_array inline without creating a separate method; I think Scrutinizer would be neutral about that, so do whatever makes sense.

xandros15:
There are 3 different conditions:

1. in readConditionalStyles, you need yo pass all valid condition types
2. in writeAttributeif (operator attribute), that should be only true when is cellIs, notContainsText or containsText, and also need to have a operator
3. to decide if sould use writeTextCondElements, when should pass only if notContainsText or containsText

xandros15:
> I also removed the typehints from the DocBlock since you are explicitly type-hinting.

image

oleibman:
Sorry about the bad advice, then. I believe that none of the tools we are using require a docBlock hint when an explicit typhint is present, and some of them can be configured to reject the situation where both are present, so it's useful to know that some IDE's insist on both.

src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
```diff
@@ -634,15 +634,21 @@ private function writeConditionalFormatting(XMLWriter $objWriter, Phpspreadsheet
 
                     self::writeAttributeif(
                         $objWriter,
-                        ($conditional->getConditionType() == Conditional::CONDITION_CELLIS || $conditional->getConditionType() == Conditional::CONDITION_CONTAINSTEXT)
-                        && $conditional->getOperatorType() != Conditional::OPERATOR_NONE,
+                        (
+                            $conditional->getConditionType() == Conditional::CONDITION_CELLIS
+                            || $conditional->getConditionType() == Conditional::CONDITION_CONTAINSTEXT
+                            || $conditional->getConditionType() == Conditional::CONDITION_NOTCONTAINSTEXT
+                        ) && $conditional->getOperatorType() != Conditional::OPERATOR_NONE,
```
oleibman, line 641: Please use `===` and `!==` where possible.

src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
```diff
@@ -634,15 +634,21 @@ private function writeConditionalFormatting(XMLWriter $objWriter, Phpspreadsheet
...
                         $conditional->getOperatorType()
                     );
 
                     self::writeAttributeIf($objWriter, $conditional->getStopIfTrue(), 'stopIfTrue', '1');
 
-                    if ($conditional->getConditionType() == Conditional::CONDITION_CONTAINSTEXT) {
+                    if (
+                        $conditional->getConditionType() == Conditional::CONDITION_CONTAINSTEXT
+                        || $conditional->getConditionType() == Conditional::CONDITION_NOTCONTAINSTEXT
+                    ) {
```
oleibman, line 651: Please use `===` and `!==` where possible.

oleibman (changes requested):
Please use `===` and `!==` where possible.

xandros15:
Tbh I don't that both condition at all. Also used double = instead of triple, because that was used previous.

oleibman (approved):

oleibman:
You can run phpcs, php-cs-fixer, and phpstan on your local system before pushing to try to avoid these errors. All except one - php-cs-fixer is objecting to the typehints in both DocBlock and method declaration after all. If you truly can't get your IDE to accept this, I would be okay with eliminating the typehint from the method declaration in this one case, not as a general rule. Let me know if you need assistance running the tests locally, or dealing with the other errors.

xandros15:
>If you truly can't get your IDE to accept this, I would be okay with eliminating the typehint from the method declaration in this one case

Nah, I can skip it, but imho that DocBlock should be there.

>Let me know if you need assistance running the tests locally

I run .\vendor\bin\php-cs-fixer fix --format=checkstyle and it make some changes. Should I commit it?

(Some typo changes, object to static call [didn't know the phpunit asserts should be call statically], phpDoc)

Issue #2035, opened 2021-04-29
Corruption when writing an xlsx doc with conditional formatting


ndench:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

Xlsx documents with conditional formatting should be able to be edited using PhpSpreadsheet and not cause corrupt file errors when later opened in Excel.

### What is the current behavior?

After creating a spreadsheet in Excel that contains conditional formatting, then reading and writing it with PhpSpreadsheet, an error will display next time it's opened in Excel. Excel prompts to recover the file because it found a problem in the document. After choosing to repair the document we see details that the `xl/styles.xml` was the culprit and it was repaired by removing formatting. Another prompt also occurs saying that the file is being edited by another user and we have to open it in read only mode.

![Screenshot 2021-04-29 164228](https://user-images.githubusercontent.com/2062388/116511672-2591a080-a90a-11eb-9aae-83986921ca62.png)
![Screenshot 2021-04-29 164320](https://user-images.githubusercontent.com/2062388/116511670-24f90a00-a90a-11eb-9661-a325b1530fba.png)
![Screenshot 2021-04-29 164335](https://user-images.githubusercontent.com/2062388/116511663-23c7dd00-a90a-11eb-8e85-71e71cb95d6b.png)

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet     = PhpOffice\PhpSpreadsheet\IOFactory::load('repro.xlsx');

$writer = PhpOffice\PhpSpreadsheet\IOFactory::createWriter($spreadsheet, 'Xlsx');
$writer->save('repro-output.xlsx');
```

Here is the [repro.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/6396749/repro.xlsx) file I created by opening a blank spreadsheet in Excel, and adding conditional formatting to a cell.

Here is the [repro-output.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/6396754/repro-output.xlsx) file I created by running the above script.

[...]

PR #2065, opened 2021-05-05
Expecting this new test to fail on PHP8 only

MarkBaker:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

New test to see if PHP8 is failing to correctly set the Default Font Size, but using no size at all (i.e. 0), if none is provided by the style definition in the XML

This is an experiment triggered by a discrepancy when reading/writing the same file against PHP7.4 and PHP8 as identified by [Issue #2035](https://github.com/PHPOffice/PhpSpreadsheet/issues/2035)

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -7282,7 +7282,7 @@ parameters:
 
 		-
 			message: "#^Cannot call method getSubscript\\(\\) on PhpOffice\\\\PhpSpreadsheet\\\\Style\\\\Font\\|null\\.$#"
-			count: 2
+			count: 1
 			path: src/PhpSpreadsheet/Writer/Xlsx/StringTable.php
 
 		-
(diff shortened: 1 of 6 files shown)


MarkBaker:
And sure enough, all tests succeed against PHP 7.4, and we have a failure for the new test against PHP 8

```
1) PhpOffice\PhpSpreadsheetTests\Reader\DefaultFillTest::testDefaultConditionalFontSize
Failed asserting that 0.0 is identical to 10.
```

oleibman:
From the title, I'm not sure if this is something you're planning to install. If you are, I think Conditional Styles depend on the initial font values being null. For example, if I have a condition where I want to set the cell to italic, I want properties like font name and size to remain unchanged (you can test for this because the fields are currently nullable), not set back to the default.

MarkBaker:
Not something to install! I should have marked it as draft.

It was a basic experiment to verify whether there was a difference between PHP7.4 behaviour and PHP8 behaviour, and then to try and determine why the conditional font size was being set to zero with PHP8; and I made the mistake of following down the rabbit hole.

MarkBaker:
Close, as this was just a test

PR #2073, opened 2021-05-07
Resolve issue with conditional font size set to zero in PHP8

MarkBaker:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

PHP 8 is setting conditional font size to 0 if no font size defined in the dxfs, which results in a corrupted output file when saving again


MarkBaker:
This should now fix the PHP8 issue with conditional font size, where the value was being set to zero rather than left as a null if the size wasn't defined in the dxfs, highlighted by [Issue #2035](https://github.com/PHPOffice/PhpSpreadsheet/issues/2035).
So now the Writer won't write size at all unless a size was explicitly defined, and the Reader won't explicitly define a size of zero if no size is defined in the original file

Issue #2076, opened 2021-05-08
Adding tests for AddressHelper::convertFormulaToA1


ndench:
This is:

```
- [ ] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

While working on #2060 I noticed that the [AddressHelper::convertFormulatToA1](https://github.com/PHPOffice/PhpSpreadsheet/blob/d2e6db71fa3605b56dd7e6162a6f3a101d9b71b1/src/PhpSpreadsheet/Cell/AddressHelper.php#L53) doesn't have any tests.

I'd like to write these tests because I think this is a very valuable function that contains a lot of logic. However there are some parts of the function that I don't understand.

The main things I'm not sure of are:
1. What sort of functions start with `of:`
2. What sort of functions contain `"` and why do we only convert the non-quoted parts?

Here's the function, with my questions in comments blocks.

```php
/**
 * Converts a formula that uses R1C1 format cell address to an A1 format cell address.
 */
public static function convertFormulaToA1(
    string $formula,
    int $currentRowNumber = 1,
    int $currentColumnNumber = 1
): string {
    if (substr($formula, 0, 3) == 'of:') {
        /*
         * I don't understand this entire if block.
         * What sort of formula starts with 'of:' and contains double quotes?
         * Why are we simply removing square brackets and full stops?
         * If this formula is in R1C1 format, shouldn't we be converting something to A1 somwhere?
         */

        $formula = substr($formula, 3);

        $temp = explode('"', $formula);
        $key = false;
        foreach ($temp as &$value) {
            //    Only replace in alternate array entries (i.e. non-quoted blocks)
            if ($key = !$key) {
                $value = str_replace(['[.', '.', ']'], '', $value);
            }
        }
    } else {
        /*
         * Again, why does the formula contain quotes?
         */

        //    Convert R1C1 style references to A1 style references (but only when not quoted)
        $temp = explode('"', $formula);
        $key = false;
        foreach ($temp as &$value) {
            //    Only replace in alternate array entries (i.e. non-quoted blocks)
            if ($key = !$key) {
                /*
                 * Inside this if block, I understand. It simply locates all R1C1 references and converts them A1.
                 */

[...]

PR #2086, opened 2021-05-11
Test AddressHelper::convertFormulaToA1

ndench:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected
- [x] Documentation is updated as necessary (not necessary)

As discussed in #2076, the `AddressHelper::convertFormulaToA1 function was not tested. This is quite a logically intensive function and as a result I actually identified a bug while writing the tests. The bug was identified with this test:

```
1) PhpOffice\PhpSpreadsheetTests\Cell\AddressHelperTest::testConvertFormulaToA1FromR1C1Relative with data set #5 ('=E3-E7-G5-C5-E5-5', '=R[-2]C-R[2]C-RC[2]-RC[-2]-RC-5', 5, 5)
PhpOffice\PhpSpreadsheet\Exception: Invalid R1C1-format Cell Reference, Value out of range
```

As you can see from [this regex101 example](https://regex101.com/r/GeHd87/1) the old regex fails for relative R1C1 conversion when the column is the current column and the reference is followed by a minus sign, eg `R[-2]C-R[2]C`.

image

I updated the regex to use non-capturing groups to enforce that the `R` or `C` is followed by either:
- A positive number
- A number inside square brackets with an optional negation sign

Which you can see from [this regex101 example](https://regex101.com/r/HHxUVk/1) that it extracts the references correctly.

image

I decided against splitting this function into two different functions (as it handles both R1C1 and SpreadsheetXML formats) in this PR as I thought it best to get some functioning tests first. Then we can split it up later.


tests/data/Cell/ConvertFormulaToA1FromR1C1Absolute.php
```diff
@@ -0,0 +1,16 @@
+<?php
+
+return [
+    // Basic arithmetic
+    ['=D3+F7+G4+C6+5', '=R3C4+R7C6+R4C7+R6C3+5'],
+    ['=D3-F7-G4-C6-5', '=R3C4-R7C6-R4C7-R6C3-5'],
+    ['=D3*F7*G4*C6*5', '=R3C4*R7C6*R4C7*R6C3*5'],
+    ['=D3/F7/G4/C6/5', '=R3C4/R7C6/R4C7/R6C3/5'],
+    // Formulas
+    ['=SUM(E1:E5)', '=SUM(R1C5:R5C5)'],
```
MarkBaker, line 10: Just as a useful little technique for your data provider array.... using associative indexes for each test can make it easier to recognise which tests are failing, and also has benefits if you report tests using additional phpunit tools
e.g.
```
return [
    'Sum with relative range' => ['=SUM(E1:E5)', '=SUM(R1C5:R5C5)'],
    'Sum with relative range and additional value' => ['=SUM(E1:E5, D5)', '=SUM(R1C5:R5C5, R5C4)'],
];
```
When a test does fail, it's reported by the name that you've given it, rather than simply it's offset number, which can make it a lot easier to identify in large data providers.

And not every test case needs to be named, so the array can be a mix of enumerated and associative

src/PhpSpreadsheet/Cell/AddressHelper.php
```diff
@@ -72,7 +72,7 @@ public static function convertFormulaToA1(
             foreach ($temp as &$value) {
                 //    Only replace in alternate array entries (i.e. non-quoted blocks)
                 if ($key = !$key) {
-                    preg_match_all('/(R(\[?-?\d*\]?))(C(\[?-?\d*\]?))/', $value, $cellReferences, PREG_SET_ORDER + PREG_OFFSET_CAPTURE);
+                    preg_match_all('/(R((?:\[-?\d*\])|(?:\d*))?)(C((?:\[-?\d*\])|(?:\d*))?)/i', $value, $cellReferences, PREG_SET_ORDER + PREG_OFFSET_CAPTURE);
```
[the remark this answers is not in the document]
ndench: Yep, good point. I mainly kept it here so you could see the change I made easily. But it's certainly better as a constant. Fixed up :+1:

tests/data/Cell/ConvertFormulaToA1FromR1C1Absolute.php
ndench: Ah yes, I completely forgot about this. I've added in proper names for each data set.

MarkBaker (commented):
Have you ever had one of those moments, when reviewing a PR, when you notice a bug in the code... not a new bug, not one that you've introduced, but something that has been in the code for years, and nobody has ever noticed.... one that I introduced when I created the original SpreadsheetML Reader.

Consider the formula
```
of:=1.23+2.34
```
or
```
of:=[.D3]+[.F7]+[.G4]+[.C6]+5.67
```
in a SpreadsheetML file... it'll process cleanly and with no error through the convertor, but it won't be the same formula.

Don't worry about it; I'll implement a fix as a separate PR, with some additional unit tests.... I'm just very, very annoyed at myself

ndench:
Ah yes, I can say I've definitely had that before :joy:. Should we merge this PR then? So that we can add that as a unit test case? I'm happy to do it if you'd like a hand?

PR #2100, opened 2021-05-15
fix: Set font size to 10 when given 0

drola:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

This change restored behavior from PHP7 in PHP8. In PHP7 calling
setSize(0) resulted in font size being set to 10. The fix addresses
change to equal comparisons in PHP8. Extra comparison is added to keep
 result from PHP7 in PHP8 for the setSize(0) case.

### DIFF
--- a/src/PhpSpreadsheet/Style/Font.php
+++ b/src/PhpSpreadsheet/Style/Font.php
@@ -249,15 +249,22 @@ class Font extends Supervisor
     /**
      * Set Size.
      *
-     * @param float $pValue
+     * @param mixed $pValue A float representing the value of a positive measurement in points (1/72 of an inch)
      *
      * @return $this
      */
     public function setSize($pValue)
     {
-        if ($pValue == '') {
-            $pValue = 10;
+        if (is_string($pValue) || is_int($pValue)) {
+            $pValue = (float) $pValue; // $pValue = 0 if given string is not numeric
         }
+
+        // Size must be a positive floating point number
+        // ECMA-376-1:2016, part 1, chapter 18.4.11 sz (Font Size), p. 1536
+        if (!is_float($pValue) || !($pValue > 0)) {
+            $pValue = 10.0;
+        }
+
         if ($this->isSupervisor) {
             $styleArray = $this->getStyleArray(['size' => $pValue]);
             $this->getActiveSheet()->getStyle($this->getSelectedCells())->applyFromArray($styleArray);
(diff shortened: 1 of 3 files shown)


src/PhpSpreadsheet/Style/Font.php
```diff
@@ -255,7 +255,7 @@ public function getSize()
      */
     public function setSize($pValue)
     {
-        if ($pValue == '') {
+        if ($pValue == '' || $pValue == 0) {
```
MarkBaker, line 258: I think the problems with invalid values lie a little bit deeper than simply trying to set a font size to 0; try setting it to -10 and see what the resulting value is... if you're going to fix the method when given an invalid value, you should probably fix it for any invalid value, not just 0

tests/PhpSpreadsheetTests/Style/FontTest.php
```diff
@@ -39,4 +39,25 @@ public function testSuperSubScript(): void
...
+        $font->setSize(12);
+        self::assertEquals(12, $font->getSize(), 'Accepted new font size');
+
+        $font->setSize('');
+        self::assertEquals(10, $font->getSize(), 'Set to 10 after trying to set an invalid value.');
+        
+        $font->setSize(12);
+        $font->setSize(0);
+        self::assertEquals(10, $font->getSize(), 'Set to 10 after trying to set an invalid value.');
+    }
```
MarkBaker, line 62: Tests should really include setting for negative numeric values, and for null or non-empty, non-numeric string

MarkBaker (changes requested):

MarkBaker:
Thank you for the contribution; although any fix to the values passed to the method should include other invalid values as well

drola:
Thank you for the review.

I expanded the test and added the handling of all other invalid values.

MarkBaker (approved):
Thanks for making those changes to full validate the value, and for ensuring that it's covered by unit tests... 

When we first wrote those basic setters, we never really expected any developer to pass through invalid values, so the validation checks were minimal.... over the years, we've been proven wrong time and again, but never really got round to improving the validation checks, as there were always more important demands for more serious bugs or new features

src/PhpSpreadsheet/Style/Font.php
```diff
@@ -255,9 +255,16 @@ public function getSize()
      */
     public function setSize($pValue)
```
MarkBaker, line 256: The docblock needs modifying to `mixed` to keep phpstan happy, and it looks like phpcs is complaining about trailing white space in the comment lines.
If you can just make those changes, then the PR is good to merge

MarkBaker (changes requested):
The docblock needs modifying to mixed to keep phpstan happy, and it looks like phpcs is complaining about trailing white space in the comment lines.
If you can just make those changes, then the PR is good to merge

drola:
Thank you. Updated the docblock, ran `composer fix`, `composer check` and rebased on top of master.

PR #2110, opened 2021-05-19
Update reading-and-writing-to-file.md

athena-metis:
Added a note to the documentation about formulas still being calculated where column autosizing is turned on, even if pre-calculation is set to false. This is true at least for the Xlsx writer but probably others to if they use calculateColumnWidths from Worksheet/Worksheet.php

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [X] a documentation addition 
```

Checklist:

- [ ] Code style is respected

Explains somewhat unexpected behaviour (you can see why it happens but may not consider it initially)


oleibman:
Very interesting observation. Might it be possible for you to write a formal test which demonstrates the behavior that you are documenting?

MarkBaker:
The reason for this is that automatic column sizing should be based on the display value of the cells, which includes formatting (font size, format masking, etc) and when a cell contains a formula, the style settings need to be applied to the calculated value for that formula rather than the formula itself

oleibman:
It appears that there are no tests for pre-calculation false. So, I will write some, bearing this change in mind, and move this change forward.

oleibman:
@athena-metis , thank you for your contribution.

Issue #2116, opened 2021-05-25
Financial::COUPNCD and Google Tables different results


ElShapovalov:
cpnNextAt = 30.09.2021
maturityAt = 31.03.2022
cpnFreq = 2
dayCntDes = 0

Financial::COUPNCD = **30.09.2021**
Google Tables = **31.03.2022**

oleibman:
Problem is caused because code subtracts 6 months from a PHP DateTime, say, 2022-03-31, which results in 2021-10-01 rather than 2021-09-30. Fixing that is a little messy, but doable, but then there are all sorts of regression possibilities. I am looking into it.

PR #2119, opened 2021-05-26
Error in COUPNCD

oleibman:
See issue #2116. Code for handling end of month (method couponFirstPeriodDate) needed a fix. Fixed it, confirmed it covered the reported issue with no regression problems. Then added some extra similar tests to all the callers of couponFirstPeriodDate, and ...

One new test, in COUPDAYSNC, does not agree with Excel. It also does not agree with LibreOffice. It does, however, agree with Gnumeric, and with my (hardly guaranteed) hand calculation of what the result should be. So, I'm going with it (and have added an appropriate comment to the test data). I'm glad to discuss the matter with anyone more familiar than I with how this is supposed to work - those 360-day years are killers.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


MarkBaker (approved):
>  those 360-day years are killers.

From memory, there's also a couple of the financial functions where the daysInYear mode is passed as an argument, but completely ignored by MS Excel.... I suspect they pass it in for consistency, so a functions look similar to  user in terms of arguments, but it's not documented that it isn't actually used

PR #2137, opened 2021-06-02
PHP8.1 Deprecation Passing Null to String Function

oleibman:
For each of the files in this PR, one or more statements can pass a null to string functions like strlower. This is deprecated in PHP8.1, and, when deprecated messages are enabled, causes many tests to error out. In every case, use coercion to pass null string rather than null.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Calculation/TextData/Extract.php
```diff
@@ -72,6 +72,6 @@ public static function right($value = '', $chars = 1): string
             $value = ($value) ? Calculation::getTRUE() : Calculation::getFALSE();
         }
 
-        return mb_substr($value, mb_strlen($value, 'UTF-8') - $chars, $chars, 'UTF-8');
+        return mb_substr($value ?? '', mb_strlen($value ?? '', 'UTF-8') - $chars, $chars, 'UTF-8');
```
MarkBaker, line 75: The null coalescence should probably be done before this point to avoid using it twice in this line.... for scrutinizer, each null coalescence increases the path count

MarkBaker (changes requested):

src/PhpSpreadsheet/Calculation/TextData/Extract.php
```diff
@@ -26,7 +26,7 @@ public static function left($value = '', $chars = 1): string
             $value = ($value) ? Calculation::getTRUE() : Calculation::getFALSE();
         }
 
-        return mb_substr($value, 0, $chars, 'UTF-8');
+        return mb_substr($value ?? '', 0, $chars, 'UTF-8');
```
MarkBaker, line 29: In the case of the Excel function implementations (particularly in the statistical functions), I typically handled these "optional" values in the initial extraction of values:

```php
public static function left($value = '', $chars = 1): string
{
    $value = ($value === null) ? '' : Functions::flattenSingleValue($value);
    $chars = ($chars === null) ? 1 : Functions::flattenSingleValue($chars);
    ...
}
```
Perhaps a similar approach would help create more consistency in the code

MarkBaker (changes requested):

src/PhpSpreadsheet/Calculation/TextData/Extract.php
oleibman: Are you certain about Scrutinizer's behavior here? I believe that PhpUnit's code coverage component treats `??` (but not `?:`) as any other operator, and does not add to the complexity rating when it is used. Does Scrutinizer behave differently?

I agree that we can avoid doing it twice in the same line. For this PR, I wanted to make all the changes exactly the same for easy review. In responding to your next comment, I am proposing to look at all of TextData. Can I fix this minor inefficiency at that time?

src/PhpSpreadsheet/Calculation/TextData/Extract.php
oleibman: For consistency's sake, I think these functions should use the same model as Engineering, MathTrig, DateTime, etc. - put the flattens and validations inside a try-catch block. I would be willing to look at this for all of TextData, but not as part of this PR.

My attempt to do something like this resulted in test failures, because your recoding above isn't quite right. Omitting the chars parameter from LEFT does act as if chars was supplied as one. However, supplying a null (e.g. `LEFT("ABC", A2)`, where A2 is an empty cell) makes LEFT act as if chars was supplied as zero. It is easy to fix your sample to reflect this, but I suspect there may be similar problems in other TextData functions. Again, I would be willing to investigate further, but not as part of this PR.

Also, the chars parameter is optional for LEFT and RIGHT, but chars and start are both required for MID, which the existing code does not recognize.

src/PhpSpreadsheet/Calculation/TextData/Extract.php
MarkBaker: I could be wrong about Scrutinizer; but technically it's the equivalent of an if/else statement with two paths through the code.

Happy to accept it as is now, with a look at simplifying it later

src/PhpSpreadsheet/Calculation/TextData/Extract.php
MarkBaker: Excel is a PITA sometimes, particularly with the way it allows "missing" arguments, that aren't trailing values.... the calculation engine passes a null in those cases

Now that we've done the bulk work on moving the Excel functions out of the mega classes, it will be a lot more practical to rationalise all validations down to a common validation class; and apply the flattens within the try/catch as you suggest
Further investigation could be awkward, as I'm not sure that OpenOffice and Gnumeric necessarily follow Excel's behaviour; and I don't believe that Excel itself has been completely consistent from version to version

MarkBaker (approved):

MarkBaker:
Idly wondering how much overhead it would add to the calculation engine to use Reflection to read the method signature, and pass in any appropriate default value as defined in that signature if an argument had been left blank, rather than sending a null value to the method.... that way, any null values received by the method could only be the result of a null value cell or an explicit null argument

It might be worth an experiment, to see how much overhead it might add in terms of performance... I'll have a go sometime in the coming weeks, just to see if it works out

MarkBaker:
Spent the weekend working on this as [PR #2143](https://github.com/PHPOffice/PhpSpreadsheet/pull/2143) (PR Is still in draft, because I want to clean up the code a bit). It works well, and with the change any null value passed to a function implementation is a genuine null, and not an empty argument... empty arguments are pre-populated with the defined default values for the argument.

I still need to run some performance tests, but the overhead doesn't seem excessive

Issue #1977, opened 2021-04-01
[Bug] When a cell with an empty string is used in a formula, #VALUE! is returned, while in Excel and other calc software, the empty string is considered as 0 and the formula works.


JulioPablo:
This is:

```
- [x] a bug report
```

### What is the expected behavior?
The cell holding an empty string value `''` \ `""` is treated as a `0`

### What is the current behavior?
The cell holding the empty string is considered invalid and when the formula cell value is attempted to be calculated ` #VALUE! ` is returned.

### What are the steps to reproduce?
If you want to reproduce this, try to get the calculated values of the cells by calling `getCalculatedValue()`.
You can also open it on your favorite calc program and check the behavior there and how it does not match.
[Formula_Test.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/6241089/Formula_Test.xlsx)

For clarity:

![image](https://user-images.githubusercontent.com/33587721/113245942-34d2ef00-9275-11eb-9a98-55bcbcfea8a4.png)

The first row does not work in PHPSpreadsheet
The second row **does** work in PHPSpreadsheet

Both rows work in Excel/Calc

### Proposed solution

The culprit of this behavior is this function 

https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Calculation/Calculation.php#L4767-L4803

My proposal is modifying the following `if` statement 

https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Calculation/Calculation.php#L4777-L4799

To look as follows

```
        //    Numbers, matrices and booleans can pass straight through, as they're already valid
        if (is_string($operand)) {
            //    We only need special validations for the operand if it is a string
            //    Start by stripping off the quotation marks we use to identify true excel string values internally
            if ($operand > '' && $operand[0] == self::FORMULA_STRING_QUOTE) {
                $operand = self::unwrapResult($operand);
            }

            //    If the string is empty, treat it as a 0, as per Excel's behavior
            $operand = $operand > '' ? $operand : 0; 

            //    If the string is a numeric value, we treat it as a numeric, so no further testing
            if (!is_numeric($operand)) {
                //    If not a numeric, test to see if the value is an Excel error, and so can't be used in normal binary operations
                if ($operand[0] == '#') {
                    $stack->push('Value', $operand);
[...]

PR #2138, opened 2021-06-02
Updates to documentation concerning Value Binders

MarkBaker:
Fix a couple of checks on characters at positions within a string before checking the length of the string when the offset may not exist

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [X] documentation
```

Checklist:

- [ ] Code style is respected

Update the documentation on Value Binders to include the case where numeric strings won't be converted to numbers, and reference the StringValueBinder

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Added
 
-- Nothing.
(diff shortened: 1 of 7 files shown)


oleibman:
Now that you are adding StringValueBinder to the docs, could you make it clear that String Value Binder also converts null and boolean to strings, and that a custom binder might be needed if that isn't the desired behavior? I got into a lengthy discussion with someone about that. I'll find that issue number later.

oleibman:
Discussion to which I referred was in issue #1977.

PR #2162, opened 2021-06-15
Autofilter Part 2

oleibman:
Most of the remaining 32-bit-unsafe date handling that remains in PhpSpreadsheet is in AutoFilter. Cleaning this up demonstrated that there are a lot of problems with AutoFilter, and I will do it in two pieces. Part 1 was PR #2141 which I have just merged.

In this PR:
- Fix remaining 32-bit dates in filterTestInDateGroupSet.
- Also in some of the existing AutoFilter samples. Note that the comments in two of those said the filter was being set for the first day of each month, but the code specifies the last day - I have corrected the comments.
- Remove mocking in unit tests for AutoFilter in favor of 'real' tests.
- Code coverage is now 100% in all of AutoFilter, AutoFilter/Column, and AutoFilter/Common/Rule.
- No remaining AutoFilter(/Column(/Rule)) exceptions in Phpstan baseline.
- Documentation for escaping of asterisk, question mark, and tilde in text filters included spurious backslashes which are now removed.
- Text filter escaping of question mark did not work. There had been no unit tests for any text filtering.
- Likewise there had been no testing for TopTen.
- Above- and below- average filters were not working because they acquired their Calculation instance incorrectly. There had been no tests.
- Several unchanging private static arrays in Rule were changed to private const arrays.
- Clones are now tested.
- RuleTest is moved to same directory as other tests.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Worksheet/AutoFilter.php
```diff
@@ -453,12 +454,10 @@ private static function filterTestInPeriodDateSet($cellValue, $monthSet)
...
     /**
      * Search/Replace arrays to convert Excel wildcard syntax to a regexp syntax for preg_matching.
-     *
-     * @var array
      */
-    private static $fromReplace = ['\*', '\?', '~~', '~.*', '~.?'];
+    private const FROM_REPLACE = ['~~', '~\\*', '\\*', '~\\?', '\\?', "\x1c"];
 
-    private static $toReplace = ['.*', '.', '~', '\*', '\?'];
+    private const TO_REPLACE = ["\x1c", '[*]', '.*', '[?]', '.', '~'];
```
MarkBaker, line 460: > Documentation for escaping of asterisk, question mark, and tilde in text filters included spurious backslashes which are now removed.

Note that there is now a helper class for handling this, the `WildcardMatch` class that's used by the Database functions. You can find examples of its use in the `buildCondition()` method of `DatabaseAbstract` or in the `ExcelMatch` class used for the Excel `MATCH()` function.
It handles a lot more edge cases than this version: I'd forgotten that it was also needed for AutoFilter as well.

MarkBaker (changes requested):
Id recommend switching to use of the WildCard class for handling wildcards in text searches, because it handles a lot of edge cases that this original simplistic method didn't take into account

src/PhpSpreadsheet/Worksheet/AutoFilter.php
oleibman: WildcardMatch does not handle `~~` correctly. I will change it to do so, and expect to have an update to this PR shortly.

oleibman: Latest push corrects, simplifies (one str_replace in lieu of two str_replaces and a preg_replace), and comments WildcardMatch, and changes AutoFilter to use it. AutoFilter was already testing for `~~`; I added an additional test for that to MATCH as well.

MarkBaker (approved):

MarkBaker:
Note that we have a few issues with PHP8.1 nightly that will need addressing before the planned November release of that PHP version; although some of these issues are with phpunit dependencies

PR #2177, opened 2021-06-22
fix 'Illegal offset type' error - authorId is SimpleXmlElement

jasverix:
This is:

- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

Got an 'illegal offset type' trying to read Excel file with comments.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -896,7 +896,7 @@ class Xlsx extends BaseReader
                                     foreach ($commentsFile->commentList->comment as $comment) {
                                         $commentModel = $docSheet->getComment((string) $comment['ref']);
                                         if (!empty($comment['authorId'])) {
-                                            $commentModel->setAuthor($authors[$comment['authorId']]);
+                                            $commentModel->setAuthor($authors[(string) $comment['authorId']]);
                                         }
                                         $commentModel->setText($this->parseRichText($comment->text));
                                     }


MarkBaker:
The problem is really a bit more deep-seated than this

Microsoft recently added "threaded comments" as a new feature.... the worst thing about it seems to be a renaming between Comments (threaded) and Notes (which used to be called Comments).... among other changes that this entails, the author of a comment is no longer stored with the comment itself in the new files, but instead the comment holds a pointer reference to a list of authors.
While this code eliminates an error message, it doesn't actually address the problem

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -896,7 +896,7 @@ public function load(string $pFilename, int $flags = 0)
                                     foreach ($commentsFile->commentList->comment as $comment) {
                                         $commentModel = $docSheet->getComment((string) $comment['ref']);
                                         if (!empty($comment['authorId'])) {
-                                            $commentModel->setAuthor($authors[$comment['authorId']]);
+                                            $commentModel->setAuthor($authors[(string) $comment['authorId']]);
```
MarkBaker, line 899: The autoincremented index for the standard PHP enumerated `$authors` array is an integer value, so you really should be casting to `int` here, not to `string`.

MarkBaker (commented):
How immediate is this as a problem for you?

While this simple solution may resolve any loader errors, it will not resolve the problems that MS have introduced with their changes to the types of comments that can be made, or with the way threaded comments are now handled.... threaded comments no longer store the author in the authors list, but in a separate xml file, and the comment itself is no longer stored in the comments list here for threaded comments... even the substitute message stored here isn't stored in the same way

In other words, can you wait until we provide support for those changes to the comments structure, which should be within the next 2-3 weeks

jasverix:
I have added this PR in my fork in "repositories" in composer.json, so I don't have any immediate need for this to be merged. I am just going to remove the "repositories" entry when you have found a solution. Thank you in advance!

PR #2178, opened 2021-06-22
Reader/Gnumeric vs. Scrutinizer

oleibman:
Just reviewing Scrutinizer's list of "bugs". There are 19 ascribed to me. For some, I will definitely take no action (e.g. use of bitwise operators in AND, OR, and XOR functions). However, where I can clean things up so that Scrutinizer is satisfied and the resulting code is not too contorted, I will make an attempt.

I believe this is the only one with which will involve more than 2 or 3 changes. It fixes 5 items ascribed to me, and 4 to others.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Reader/Gnumeric.php
```diff
@@ -266,9 +266,10 @@ public function loadIntoExisting(string $pFilename, Spreadsheet $spreadsheet): S
         (new Properties($this->spreadsheet))->readProperties($xml, $gnmXML);
 
         $worksheetID = 0;
-        foreach ($gnmXML->Sheets->Sheet as $sheet) {
+        foreach ($gnmXML->Sheets->Sheet as $sheetOrNull) {
+            $sheet = self::testSimpleXml($sheetOrNull);
             $worksheetName = (string) $sheet->Name;
-            if ((isset($this->loadSheetsOnly)) && (!in_array($worksheetName, $this->loadSheetsOnly))) {
+            if (is_array($this->loadSheetsOnly) && !in_array($worksheetName, $this->loadSheetsOnly)) {
```
MarkBaker, line 272: The `in_array()` check should also use the 3rd argument for strict checking

MarkBaker (changes requested):

src/PhpSpreadsheet/Reader/Gnumeric.php
oleibman: Thanks, hadn't known about the extra argument.

MarkBaker: Most of this codebase was written before it was introduced to PHP, but I try to apply it whenever I'm modifying anything and notice an `in_array()`

PR #2181, opened 2021-06-23
Check for existance of cellCollection before we try to use it

thunderquiet:
Will update these as I go along after we confirm the fix is good in our internal projects.

This is:

```
- [X ] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected


MarkBaker (commented):
What problem is this trying to solve? What error does it fix?

If a worksheet exists, then a cell collection should also exist, because it's defined in the sheet's constructor.

We need unit tests to prove that this works if it is a fix for some problem

thunderquiet:
I was able to fix the issue in the code we maintain. Closing this one out as its no longer required.

For reference, we were calling disconnect on a sheet but some formulas in other sheets were still refering to it. Hence cell collection was null and trying to access it would blow up.

PR #2188, opened 2021-06-25
DOCS: Gridlines when printing

woytam:
This is:
```
- [X] a bugfix of docs
```

I think that right function to show/hide grid lines when printing is `setPrintGridlines()` and not `setShowGridlines()` as said in docs.


oleibman:

Issue #1663, opened 2020-10-01
PHP8 breaks Writer/Xlsx/LocaleFloatsTest; Probably Not Fixable


oleibman:
This is:

```
- [X] a bug report
- [ ] a feature request
- [X] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Unit test sets locale to France, sets cell to 1.1, and reads contents of cell expecting "1,1".
Test succeeds under PHP7; it fails ("1.1") with PHP8. The PHP RFC authorizing the breaking change is:
https://wiki.php.net/rfc/locale_independent_float_to_string
Quoting from there, "In our opinion, the backward compatibility break won't be very serious in practice ..."
It isn't all that difficult to "fix" the test - skip it for PHP8, or delete it as you probably will have to eventually,
or use sprintf (except that will be testing PHP itself, not PhpSpreadsheet), probably others.
The problem is that, even if the test is changed, users depending on the current behavior will view this as a
PhpSpreadsheet problem.
In North America, I am pretty much insulated from this problem. I cannot guess how seriously it might be seen in Europe.

### What is the current behavior?
"1,1" with PHP7, "1.1" with PHP8.

### What are the steps to reproduce?
Existing unit test Writer/Xlsx/LocaleFloatsTest.php demonstrates the problem.

### Which versions of PhpSpreadsheet and PHP are affected?
PHP8, all versions of PhpSpreadsheet.

oleibman:
Undoing stalebot.

oleibman:
My initial tests had been with a Release Candidate version of PHP8. Current testing suggests that var_dump has changed since then to produce locale-aware output, which would mean that no changes are required to unit tests. I don't see anything in the change logs which suggest that this behavior has been changed, so I will leave the issue open for now and continue to monitor. The affected test is skipped in the PhpUnit tests on GitHub, because trying to set the locale fails, presumably because only one locale is installed on the test system (or, at least, fr_FR.UTF-8 isn't available). This appears to be the case for all PHP releases which are part of the unit test, not just 8.0/8.1. This isn't ideal; I don't know what can be done to change that. The tests are executed on my local systems.

oleibman:
[...]

PR #2191, opened 2021-06-26
Correct Some Problems Which Will Show Up for PHP8.1

oleibman:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Correct Some Problems Which Will Show Up for PHP8.1

PHP8.1 wants to issue a message when you use a float where it thinks you ought to be using an int (it wants its implicit casts made explicit). This is causing unit tests to fail. The following corrections are made in this PR:
- Calculation.php tests `isset(self::binaryOperators[$token])`, where token can be a float. No numeric values are members of that array, so we can test for numeric before isset.
- SharedOle.php packs a float, intending it as an int, in 2 places. I simplified the logic here, and added explicit casts to avoid the problem. This is used by Xls Reader and Writer; as added confirmation, I added some timestamps from before 1970 (i.e. negative values) to Document/EpochTest. Because of this, the test suite has been verified for 32-bit PHP as well as PHP 8.1.
- Writer/Xlsx/StringTable tests `isset($aFlippedStringTable[$cellValue])`. This is the same problem as in Calculation, but requires a different solution. The same if statement here also tests that the datatype is string, but it does so after the isset test. Changing the order of these tests avoids the problem.

There remain PHP8.1 errors having to do with vendor code used as a polyfill for Enum. @MarkBaker has reported the issue to the vendor.


oleibman:
Apologies for the multiple commits. I started out from the wrong baseline.

MarkBaker:
Enum polyfill issue has now been resolved

MarkBaker:
Three issues now remain for PHP 8.1, all in mPdf

oleibman:
FWIW, I reported the Mpdf Php8.1 issues a while back. They closed my ticket immediately, and were, I thought, quite rude about it. The fix on their side is really easy, but I'm not going to attempt it again. Perhaps you'll have better luck.

There is still an issue with LocaleFloatsTest. It fortuitously is skipped because there is no "French" environment set up when Github runs the unit tests. However, if there were one, the test would fail, for the same reason as I thought it would fail for 8.0 (issue #1663).

PR #2195, opened 2021-06-29
Use certer for vertical-align in Excel when meet vertical-align:middel in CSS

nkjackzhang:
…Excel

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

When convert HTML to Excel, and meet `vertical-align: middle` for `td` element in CSS, we should not set "middle" to `vertical` field in `Alignment`, which doesn't work, instead, we should set "center" to it.


src/PhpSpreadsheet/Style/Alignment.php
```diff
@@ -240,6 +240,10 @@ public function setVertical($pValue)
             $pValue = self::VERTICAL_BOTTOM;
         }
 
+        if ($pValue == 'middle') {
```
MarkBaker, line 243: It would be better if the method actually validated that the argument passed in was a valid entry from the possible values in the set of defined constants `self::VERTICAL_*` rather than arbitrary case-sensitive values; providing the default `VERTICAL_BOTTOM` if the value wasn't valid.

Likewise, it would also be good if a similar check was performed on setting horizontal alignment

MarkBaker (changes requested):

src/PhpSpreadsheet/Style/Alignment.php
nkjackzhang: I find a function in [Writer/Html](https://github.com/PHPOffice/PhpSpreadsheet/blob/2ae948a31915c19e2748d26845eee8f4bb012302/src/PhpSpreadsheet/Writer/Html.php#L240-L243), which is used to map vertical alignment in Excel to `vertical-align` in CSS, but it maps both `center` and `justify` in Excel to `middle` in CSS. Can i only map `middle` to `certer` and ignore `justify`?

MarkBaker: The solution then should be in the mapping for the HTML Writer, providing a mechanism that allows the defined valign mapping VALIGN_ARR to be overridden with a custom mapping

oleibman:
Converting to draft. Change is superseded by PR #3048. If that change is merged, I will close this one.

oleibman:
Superseded by 3048, which has just been merged. Closing this PR.

Issue #1897, opened 2021-03-04
Sheet Password Protection not working with default algorithm and password length greater than 24


Wazzledi:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

If i use the Password Protection in PHPSpreadsheet i should not unprotect the File in Excel without using the password.

### What is the current behavior?

When using the default hash algorithm for password protection with a password longer than 24 characters and write to a XLSX file, i can easily unprotect the worksheet without a password.

This error is only affecting XLSX files in my tests.

### What are the steps to reproduce?

Run the code below to create an unprotected file and a protected file.

After creating the files open the notProtected.xlsx file in excel and try to remove the protection with the unprotect sheet option in the review tab.

In my tests the sheet gets unprotected without entering a password.

As comparison: When trying to remove the protection in the protected.xlsx file i have to enter a password to unlock the file.

```php
<?php


use PhpOffice\PhpSpreadsheet\Spreadsheet;
use PhpOffice\PhpSpreadsheet\Writer\Xlsx;

require_once __DIR__ . "/vendor/autoload.php";

// Create new Spreadsheet object for protected spreadsheet
$spreadsheetProtected = new Spreadsheet();

$protection = $spreadsheetProtected->getActiveSheet()->getProtection();
//Password length <= 24
$protection->setPassword("ABCDEFGHIJKLMNOPQRSTUVWX");
$protection->setSheet(true);

$writer = new Xlsx($spreadsheetProtected);
$writer->save("protected.xls");

// Create new Spreadsheet object for unprotected spreadsheet
$spreadSheetNotProtected = new Spreadsheet();

$protection = $spreadSheetNotProtected->getActiveSheet()->getProtection();
//Password length > 24
$protection->setPassword("ABCDEFGHIJKLMNOPQRSTUVWXY");
$protection->setSheet(true);

//Only broken for xlsx
$writer = new Xlsx($spreadSheetNotProtected);
$writer->save("notProtected.xls");
```

### Which versions of PhpSpreadsheet and PHP are affected?

Used PHPSpreadsheet Version: 1.17.1

Used PHP Version: 7.4.10

oleibman:
[...]

PR #2197, opened 2021-06-29
Fix Worksheet Passwords

oleibman:
Fix for issue #1897.

The existing hashing code seems to work correctly almost all the time, but there are exceptions. It is replaced by an exact implementation of the spec, including a link to the spec in the comments. Cases known to fail are added to the unit test suite.

The spec expects the string to be at most 255 bytes (yes, bytes not characters). The program had permitted any length; it will now throw an exception when the maximum length is exceeded.

Xls does not support any hashing algorithm except basic. The Xls writer had, nevertheless, accepted the results of any of the other possible algorithms. This leads to (a) a worksheet that can't be unprotected, and (b) deprecation notices during the write (because it is using hexdec, which expects only hex characters, and the other algorithms generate non-hex characters). I have changed Xls writer to ignore passwords generated by other algorithms. An alternative would be to have the password hasher generate both an algorithmic password (for use by Xlsx) and a basic password (for use by Xls); I think that is too complex a solution, but can look into it if you think it worthwhile. The documentation mentions that the other algorithms are usable only for Xlsx, so I don't think a doc change is required.

I do not see any current support for Worksheet passwords in ODS Reader or Writer. I did not add support in this PR.

I added a new test to confirm the password for reading a spreadsheet is consistent with the one used for writing it. As you can see from the comments for the new test, it had an unusual problem with a somewhat unusual solution.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


oleibman:
I found the test (SettingsTest) which created the unusual problem for my new test. Expect a new commit/push to this PR later today.

Issue #2203, opened 2021-07-01
Reader/Xlsx Chart Problems, Probably Due to Namespacing Changes


oleibman:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Samples spreadsheets 32* look wrong when created with master. They appear okay with 18.0. Strongly suspect the problem is due to the namespacing changes in Reader/Xlsx. I am researching.

### What is the current behavior?
Charts are wrong.

### What are the steps to reproduce?
Run Samples/32_Chart_read_write.php

### Which versions of PhpSpreadsheet and PHP are affected?
18.0 unaffected, master affected.

oleibman:
The chart is constructed with the correct set of parameters in Reader/Xlsx (using 32readwriteAreaChart1.xlsx). When Writer recalculates the values, they wind up being reset to null string. No idea why. When I stop the recalculation, it still doesn't work. So multiple problems are happening. I'll start on it again tomorrow, probably with a different spreadsheet that doesn't use random numbers, in order to simplify things.

oleibman:
Interesting - new code works correctly with non-random data. But it seems to fail if a formula is used in any cell.

oleibman:
It seems that reading the chart is causing all the formulas on the sheet to go bad, whether or not involved in the chart.

oleibman:
It appears the code is not handling shared formulas properly. I need to study.

oleibman:
Fixed by #2204. Charts turn out to have nothing to do with the problem; the sample chart spreadsheets just happened to expose the real problem, which was that shared formulas were not being handled correctly.

PR #2204, opened 2021-07-02
Xlsx Reader Better Namespace Handling Phase 1 First Bugfix

oleibman:
See issue #2203. An undotted i uncrossed t. When using namespaces, need to call attributes() to access the attributes before trying to access them directly. Failure to do so in castToFormula caused problem for shared formulas.

Surprisingly, this didn't show up in unit tests. Perhaps sharing the same formula between two cells isn't common. It did show up in Chart Samples. I've added a test.

I was really inclined to merge this right away. Not to worry - I can control myself. It should be moved fairly quickly nevertheless.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


MarkBaker (approved):
With such a major change, there were always going to be small parts that had been missed, especially as it was a high-risk change.... but the major benefits make it worthwhile

Issue #549, opened 2018-06-13
MACCENTRALEUROPE instead MAC-CENTRALEUROPE


rokirokita:
This is:

```
- [X ] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

Normal encoding.

iconv -l return:

(...)
JIS_C62291984B, JOHAB, JP-OCR-B, JP, JS, JUS_I.B1.002, KOI-7, KOI-8, KOI8-R,
  KOI8-RU, KOI8-T, KOI8-U, KOI8, KOI8R, KOI8U, KSC5636, L1, L2, L3, L4, L5, L6,
  L7, L8, L10, LATIN-9, LATIN-GREEK-1, LATIN-GREEK, LATIN1, LATIN2, LATIN3,
  LATIN4, LATIN5, LATIN6, LATIN7, LATIN8, LATIN9, LATIN10, LATINGREEK,
  LATINGREEK1, **MAC-CENTRALEUROPE**, MAC-CYRILLIC, MAC-IS, MAC-SAMI, MAC-UK, MAC,
  MACCYRILLIC, MACINTOSH, MACIS, MACUK, MACUKRAINIAN, MIK, MS-ANSI, MS-ARAB,
  MS-CYRL, MS-EE, MS-GREEK, MS-HEBR, MS-MAC-CYRILLIC, MS-TURK, MS932, MS936,
  MSCP949, MSCP1361, MSMACCYRILLIC, MSZ_7795.3, MS_KANJI, NAPLPS, NATS-DANO,
  NATS-SEFI, NATSDANO, NATSSEFI, NC_NC0010, NC_NC00-10, NC_NC00-10:81,
  NF_Z_62-010, NF_Z_62-010_(1973), NF_Z_62-010_1973, NF_Z_62010,

(...)

### What is the current behavior?
PhpSpreadsheet has problem with file which has MAC-CENTRALEUROPE encoding
Notice: iconv(): Wrong charset, conversion from `MACCENTRALEUROPE' to `UTF-8//IGNORE//TRANS
LIT' is not allowed

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

//file with MAC-CENTRALEUROPE encoding
$file = 'file.xls';
[...]

PR #2213, opened 2021-07-07
Xls Reader Handle MACCENTRALEUROPE With or Without Hyphen

oleibman:
Fixes issue #549 and https://github.com/Maatwebsite/Laravel-Excel/issues/989 (which is the source of the new test file). Some systems accept MACCENTRALEUROPE as the name for the appropriate encoding, and some accept MAC-CENTRALEUROPE. I fortunately have access to at least one of each type, and have run the tests on each.

CodePage.php has an array of translations from codepage number to string. I now allow the value to itself be an array; if so, the code will test each in turn to see if it can be used in iconv. I did not go fishing for other similar problems. If such show up, they can be dealt with in the same manner as this one. I don't really expect others, since this is a problem not merely for Xls, but, even then, it applies only to BIFF5 and earlier.

I also moved XlsTest from Reader to Reader/Xls.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Shared/CodePage.php
```diff
@@ -82,8 +83,19 @@ public static function validate(string $codePage): bool
      */
     public static function numberToName(int $codePage): string
     {
-        if (array_key_exists($codePage, self::$pageArray)) {
-            return self::$pageArray[$codePage];
+        if (array_key_exists($codePage, self::PAGE_ARRAY)) {
+            $value = self::PAGE_ARRAY[$codePage];
+            if (is_array($value)) {
+                foreach ($value as $encoding) {
+                    if (@iconv('UTF-8', $encoding, ' ') !== false) {
```
MarkBaker, line 90: Is there any way that we could identify which of the possible charsets for a codepage value is being used on the first request for a file, and then cache that result so that we don't have to do the check for every single request? It's not critical, because it's never going to be more than 3 requests per file, but just wondering.

MarkBaker (approved):

src/PhpSpreadsheet/Shared/CodePage.php
oleibman: Done. Tradeoff is making array static rather than const.

MarkBaker (approved):

PR #2214, opened 2021-07-07
Disable iconv default options if they fail ( as seen on alpine images)

pokyunn:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

When using this code snippet to promote all "PHP Warning" to  "ErrorException", inside an alpine docker image.

set_error_handler(function ($severity, $message, $file, $line) {
    throw new ErrorException($message, 0, $severity, $file, $line);
});

The operator @ do not silence warnings anymore as they are promoted into ErrorException.
Since in composer is specified "require php ^7.2 || ^8.0", i choose Throwable interface, as this is implemented by exceptions, so if some one promote to a custom exception, this will work too, but I've kept the @ to preserve the current behavior when this promotion snippet is not present


oleibman:
This seems to open a can of worms, potentially requiring similar code everywhere that messages are, with good reason, suppressed. Could you not achieve the same effect with an alteration to your error_handler code?
```php
set_error_handler(
    function ($severity, $message, $file, $line) {
        if (!(error_reporting() & $severity)) {
            return false;
        }
        throw new ErrorException($message, 0, $severity, $file, $line);
    }
);
```

pokyunn:
Hi thanks for your reply. I didn't understand the bitwise operation at first, but searching for it, i've found your suggestion on [php8 migrations notes](https://www.php.net/manual/en/migration80.incompatible.php), i remember checking that more than twice, but for some reason i didn't catch it.

And yes after read it again and some tests, i agree this error_handler was good before 8 because the @ was modifying the return code to 0, and that was not triggering it, but now it's a bit dangerous as it is.

So i'll close this pull request, and thanks again.

PR #2223, opened 2021-07-15
Skip inner loops for empty/default read filter

rvanvelzen:
This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] a performance improvement
```

Checklist:

- [x] Code style is respected

The optimization from #773 was not copied along in #1033, so restore it


src/PhpSpreadsheet/Reader/Xlsx/ColumnAndRowAttributes.php
```diff
@@ -91,6 +92,10 @@ public function load(?IReadFilter $readFilter = null, $readDataOnly = false): vo
             $rowsAttributes = $this->readRowAttributes($this->worksheetXml->sheetData->row, $readDataOnly);
         }
 
+        if ($readFilter !== null && get_class($readFilter) === DefaultReadFilter::class) {
+            $readFilter = null;
```
MarkBaker, line 96: If it makes a measurable difference, wouldn't it be better to simply have the defaultFilter as null in the BaseReader rather than defaulting it to the DefaultReadFilter?

MarkBaker (commented):
Does this change actually make a measurable difference to load speed? If so, how much?

src/PhpSpreadsheet/Reader/Xlsx/ColumnAndRowAttributes.php
rvanvelzen: Since (as far as I could tell) every call-site of the read filter checks if it is null, that's probably a safe way to go.

rvanvelzen:
> Does this change actually make a measurable difference to load speed? If so, how much?

I have a couple of sheets take take ~7 seconds to load without this change and ~200 milliseconds after the change. I can't share those exact sheets but I might be able to generate a similar thing I you wish.

rvanvelzen:
Unfortunately setting `readFilter` to `null` by default breaks a bunch of other tests, so that doesn't appear to be a viable path. It would also be a BC break for other consumers that expect it to be set.

PR #2228, opened 2021-07-19
Decrease memory consumption for 1.18.0

gilberg-vrn:
This is:

- [x] improvement

Checklist:

- [x] Code style is respected

On write large tables (500k rows x 20+ cols) PhpSpreadsheet fails on extra memory allocation (checked on max_memory = 4G)
With that PR we got:
- stringTable now use temporary file
- stringTable store inverted index in cache
- cells index rewrite for use 4-bytes blocks and Generator instead in-memory
- rewrite getHighest* methods in Cells
- XMLWriter now return resource instead file content when used STORAGE_DISK


gilberg-vrn:
Can fix 2107

vassil-velichkov:
Any progress on this PR?

src/PhpSpreadsheet/Writer/Xlsx/StringTable.php
```diff
@@ -42,35 +29,33 @@ public function createStringTable(Worksheet $pSheet, $pExistingTable = null)
                 !is_object($cellValue) &&
                 ($cellValue !== null) &&
                 $cellValue !== '' &&
-                ($cell->getDataType() == DataType::TYPE_STRING || $cell->getDataType() == DataType::TYPE_STRING2 || $cell->getDataType() == DataType::TYPE_NULL) &&
-                !isset($aFlippedStringTable[$cellValue])
+                !$aFlippedStringTable->has(md5($cellValue)) &&
```
fischaz, line 32: Hello, 

I've being trying to use this branch and noticed that my unit tests writing and reading XLSX were failing with an index out of bound error. 

Drilling through, I eventually found out that the "cache" (in the unit test  a shared array in memory, but in production a shared redis), would acttually potentially clash when writing multiple XLSX file that re-use string between the spreadsheet eg:

- because the cache exists, it would not write the text in the cache and increment the stringTableRecordsCount value. (the sharedStrings.xml would not contain the entry either)
- when writing the sheet.xml it would then find the entry in the cache again, and point to a non existent offset.

in my branch, I've been generating a 'uniqid()' string to avoid any cache clash and pass it around the same way the CacheInterface is passed around. (and the cache will eventually purge itself anyway per cache underlying configuration).

src/PhpSpreadsheet/Writer/Xlsx/StringTable.php
```diff
@@ -42,35 +29,33 @@ public function createStringTable(Worksheet $pSheet, $pExistingTable = null)
...
+                ($cell->getDataType() == DataType::TYPE_STRING || $cell->getDataType() == DataType::TYPE_STRING2 || $cell->getDataType() == DataType::TYPE_NULL)
             ) {
-                $aStringTable[] = $cellValue;
-                $aFlippedStringTable[$cellValue] = true;
+                fwrite($stringTableFD, serialize($cellValue) . PHP_EOL);
```
fischaz, line 35: The second issue I found is that serialize of a string with a newline does **not** encode the newline char at all, so it will clash with the `PHP_EOL` and will fail during the `writeStringTable` function below when running fgets (only get up to the internal newline and unserialize will fail).

I added in my branch a use of `base64_encode` (and decode) to work around this issue.

PowerKiKi (changes requested):
Thank you for this contribution. There are a lot of interesting things in there, but there also a bit too mixed. As you clearly explained yourself, there are at least 5 differents things, some of those which cannot be merged as-is, as mentioned by @fischaz.

Could you please split this PR in 5 new PRs, to make them easier to review and merge separately ?

- stringTable now use temporary file
- stringTable store inverted index in cache
- cells index rewrite for use 4-bytes blocks and Generator instead in-memory
- rewrite getHighest* methods in Cells
- XMLWriter now return resource instead file content when used STORAGE_DISK

src/PhpSpreadsheet/Collection/Cells.php
```diff
@@ -96,8 +96,61 @@ public function has($pCoord)
             return true;
         }
 
+        $column = '';
+        $row = 0;
+        sscanf($pCoord, '%[A-Z]%d', $column, $row);
```
eigan, line 101: This might save some bytes, but sscanf adds some speed penalties. Es

eigan: "few bytes" is an understatement. The logic in this file made `getSortedCoordinates` use _a lot_ less ram. In my case with a sheet of 1000 strings in the stringtable and some 350K cells did it go from 80MB to zero.

eigan:
@gilberg-vrn I opened #2529 with your changes from Cells. Put you as author of the commit. Please let me know if I should close it or whatever 🙂

Issue #1159, opened 2019-09-16
Add Support for twoCellAnchor Object Positioning for Drawing


sgarwood:
This is:

```
- [ ] a bug report
- [x] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behaviour?

When creating an image in Excel there is the option to use 2 cell anchoring or 1 cell anchoring. 

This allows users in Excel to scale the image with the cell in a more robust manner and makes
editing the spreadsheet manually after export less cumbersome.

**OpenXML specificaion**
http://officeopenxml.com/drwPicInSpread-twoCell.php

**Example of desired behaviour**
```php
$drawing->setName('Paid');
$drawing->setDescription('Paid');
$drawing->setPath('./images/paid.png');

// New features below

$drawing->setTopRightCell('B15'); // alias to $drawing->setCoordinates('B15');
$drawing->setBottomRightCell('D20');

$drawing->editAs('oneCell'); // possible values are, oneCell, twoCell or absolute
$drawing->setOffsetX(110);
```

Exact semantic's of API can be discussed in #1160 

### What is the current behaviour?

Currently, there is only support for [One Cell positioning]([http://officeopenxml.com/drwPicInSpread-oneCell.php)

### What are the steps to reproduce?

Please see above

### Which versions of PhpSpreadsheet and PHP are affected?
All current releases

sgarwood:
Expands on #643 and #126

oleibman:
Implemented by PR #2532 and subsequent changes in March 2022.

PR #2237, opened 2021-07-26
Allow "Move but don't size with cells" to be set on images.

AdamGaskins:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

See #1159 (and the abandoned PR #1160)

By default when adding a Drawing to a spreadsheet, it is added inside a `oneCellAnchor`. This makes it easier to specify a width and height, however in some cases developers may want to change the default value of the option pictured below in Excel. This PR allows developers to add images into spreadsheets with the setting "Move and size with cells."

image

Example use case: imagine a spreadsheet of product data where each row has a column called "Images." When a product row is deleted, it would make sense for the photo to be deleted as well. The only way for this to happen is to have "Move and size with cells" set.

All existing code will work the same, as `oneCellAnchor` remains the default. See `samples/Basic/47_Image_move_and_size_with_Cells.php` for an example of the new process using `twoCellAnchor`.
- Change the `Drawing` to `twoCellAnchor` mode.
- Specify the coordinates and offset as usual.
- Specify the coordinates and offset for the bottom right corner of the image using the new `setCoordinates2()`, `setOffsetX2()` and `setOffsetY2()` methods.
    - (`setWidth()` and `setHeight()` only work with `oneCellAnchor` mode)
- Add to the worksheet like normal

One other note: I also added `getImageWidth()` and `getImageHeight()` functions to drawing that return the original pixel width/height of the image. `getWidth()` and `getHeight()` contain this value when the image is initially loaded into the `Drawing`, but if you resize it in the worksheet this info is lost. Having these isn't necessary, but may be helpful due to the increased complexity of inserting images in `twoCellAnchor` mode.


PowerKiKi (changes requested):
This looks like a nice improvement. There is a few things to improve though.

- the PR would need to be rebased on master
- unit tests are missing
- all new code must be type-hinted (not phpdoc-ed)
- parameters like `$pValue` should not be used anymore, instead use proper names, such as `$anchorType` and so on

maximerenou:
Hello @AdamGaskins, are you still available to help on this feature? It would be a nice improvement.

AdamGaskins:
Sure I can make the requested changes 👌🏻

oleibman:
Superseded by PR #2532 and PR #2674.

PR #2250, opened 2021-08-04
Fixing Switch function case comparison with an expression

zcser:
Example bug: =SWITCH(TRUE,1=0,"case1","default") will be evaluated as "case1" because before in PhpOffice\PhpSpreadsheet\Calculation\Logical\Conditional line 92 $targetValue(bool)==$arguments[$index*2+1](array) will be true.

This is:

```
- [ ] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected


oleibman:
I do not duplicate your results with the latest release or master version of PhpSpreadsheet:
```php
$sheet->getCell('A1')->setValue('=SWITCH(TRUE,1=0,"case1","default")');
echo $sheet->getCell('A1')->getCalculatedValue(), "\n";
```
The result I see is 'default', which seems to match Excel's behavior. Can you share your code which demonstrates the problem you are trying to fix? (And also PhpSpreadsheet version and Php version.)

oleibman:
7 months with no reply, still don't know what problem this is solving, closing.

Issue #2262, opened 2021-08-09
Function hasValidValue Breaks Coordinates of Tested Cell


oleibman:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
`$testCell->hasValidValue()` should not alter testCell.

### What is the current behavior?
`$testCell->hasValidValue()` can alter testCell; in particular, it may wind up pointing to a cell with different coordinates than it had been pointing to.

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

DataValidatorTest already demonstrates the problem as illustrated by statements such as:
```php
$testCell = $sheet->getCell('A1'); // redefine $testCell, because it has broken coordinates after using other cells
```
This is an easy enough workaround, but it should not be needed. The problem unfortunately happens way too deep in the Calculation engine for me to unravel.

### Which versions of PhpSpreadsheet and PHP are affected?
All.

oleibman:
Closing. Related to many other issues where cell should not be stored in a variable. Just an unexpected fact of life for PhpSpreadsheet usage.

PR #2150, opened 2021-06-09
Fix for #2149 / Read data validations for drop down list in another sheet.

otargetj2s:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Changes are covered by unit tests
- [x] Code style is respected
- [x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ ] CHANGELOG.md contains a short summary of the change
- [ ] Documentation is updated as necessary

### Why this change is needed?
This modification makes it possible to read the drop-down lists linked to other sheets. 
See issue #2149

Issue #2149, opened 2021-06-09
Xlsx reader not recognize data validations that references another sheet - Again


otargetj2s:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Load the data validations that references lists of another sheet.

### What is the current behavior?
return data validations with none type. Data not loaded for external list.
This regression was reported in early 2018. Never merged.

Excel sample:
[sample.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/6622841/sample.xlsx)

See links :
https://github.com/PHPOffice/PhpSpreadsheet/issues/388
https://github.com/PHPOffice/PhpSpreadsheet/issues/991
https://github.com/PHPOffice/PhpSpreadsheet/issues/1432

### What are the steps to reproduce?
Create a xlsx file with a column with data validation that reference a list of another sheet, then load it in your code.

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Reader and Spreadsheet object
$reader   = \PhpOffice\PhpSpreadsheet\IOFactory::createReader('Xlsx');
$spreadsheet = $reader->load('sample.xlsx');

// add code that show the issue here...
$spreadsheet->setActiveSheetIndexByName('Feuil1');
$worksheet  = $spreadsheet->getActiveSheet();

// internal drop down list: success (type "list")
echo 'B5:DrowDownListType: ', $worksheet->getCell('B5')->getDataValidation()->getType(), "\n";
// external drop down list: failure, data not loaded (type "none" instead of "list")
echo 'B14:DrowDownListType: ', $worksheet->getCell('B14')->getDataValidation()->getType(), "\n";
```

### Which versions of PhpSpreadsheet and PHP are affected?
phpspreadsheet >= 1.6.0
php >= 7.2

otargetj2s:
See the following link for code correction :
https://github.com/PHPOffice/PhpSpreadsheet/issues/1432#issuecomment-640028401

Can you do what's nessessary? I absolutely need it for my development.
Thanks.

PowerKiKi:
[...]

Issue #1432, opened 2020-04-03
Data validation not working when opening & saving a file


tntsoft:
This is:

```
- [x ] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What am I trying to accomplish?
I want to open an XLSX file that has some List Dropdown validation rules in it, write some data in it and save it.

### What is the expected behavior?
When opening and saving an Excel file, it should retain all the data validation inside it. 

### What is the current behavior?
It loses all data validation fields ( dropdowns, explanations etc)

### What are the steps to reproduce?
Open the bellow excel file with data validation -> save it -> it loses validation.

### Sample XLSX file to help reproduce
[sample.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/4427243/sample.xlsx)

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

use PhpOffice\PhpSpreadsheet\IOFactory;

// Declare the path of the file
$inputFileName = __DIR__ . '/sample.xlsx';

// Load the file
$spreadsheet = IOFactory::load($inputFileName);

// Get first sheet in preparation for changes
$sheetData = $spreadsheet->getSheet(0);

// Create an XLSX writer
$writer = IOFactory::createWriter($spreadsheet, "Xlsx");

// Save the file
$writer->save("05featuredemo.xlsx");
```

### Which versions of PhpSpreadsheet and PHP are affected?
"phpoffice/phpspreadsheet": "^1.11"
PHP 7.3

tntsoft:
No suggestions, anybody?

zzhovo:
1.12.0

still doesn't work

PowerKiKi:
Try to open-and-save without doing anything at all in between. Then unzip both xlsx files and compare the content of each files. Can you pinpoint what changed?

Second question, are you able to create a file from scratch, create the validation and save to for properly?

zzhovo:
Look at this line, https://github.com/PHPOffice/PhpSpreadsheet/blob/ac7fb4a31d16cec9ee8313cd7ffe357bc9039447/src/PhpSpreadsheet/Reader/Xlsx.php#L787

[...]

PR #2265, opened 2021-08-10
Data Validations Referencing Another Sheet

oleibman:
See issues #1432 and #2149. Data validations on an Xlsx worksheet can be specified in two manners - one (henceforth "internal") if a list is specified from the same sheet, and a different one (henceforth "external") if a list is specified from a different sheet. Xlsx worksheet reader formerly processed only the internal format; PR #2150 fixed this so that both would be processed correctly on read. However, Xlsx worksheet writer outputs data validators only in the internal format, and that does not work for external data validations; it appears, however, that internal data validations can be specified in external format.

This PR changes Xlsx worksheet writer to use only the external format. Somewhat surprisingly, this must come after most of the other XML tags that constitute a worksheet. It shares this characteristic (and XML tag) with conditional formatting. The new test case DataValidator2Test includes a worksheet which has both internal and external data validation, as well as conditional formatting.

There is some additional namespacing work supporting Data Validations that needs to happen on Xlsx reader. Since that is substantially unchanged with this PR, that work will happen in a future namespacing phase, probably phase 2. However, there are some non-namespace-related changes to Xlsx reader in this PR:
- Cell DataValidation adds support for a new property sqref, which is initialized through Xlsx reader using a setSqref method. If not initialized at write time, the code will work as it did before the introduction of this property. In particular, before this change, data validation applied to an entire column (as in the sample spreadsheet) would be applied only through the last populated row. In addition, this also allows a user to extend a Data Validation over a range of cells rather than just a single cell; the new method is added to the documentation.
- The topLeft property had formerly been used only for worksheets which use "freeze panes". However, as luck would have it, the sample dataset provided to demonstrate the Data Validations problem uses topLeft without freeze panes, slightly affecting the view when the spreadsheet is initially opened; PhpSpreadsheet will now do so as well.

It is worth noting issue #2262, which documents a problem with the hasValidValue method involving the calculation engine. That problem existed before this PR, and I do not yet have a handle on how it might be fixed.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


oleibman:
Existing code does not test for  validity of other than lists (e.g. whole number). That should not delay the fixing of the bug associated with this PR. I have added fixing the omission to my to-dos.

PR #2292, opened 2021-09-08
fixed null conversation for strToUpper

brainfoolong:
fixed in the same way as it already has been done for strToLower

This is:

```
- [x] a bugfix
```

Checklist:

- [ ] Code style is respected

Just a fix applied to `strToUpper` in the same way as it has already been done for `strToLower`.


brainfoolong:
Unfortunetely i don't have an example. The error happens for me, when i load an existing excel file (that was creating by unknown software) that i can't publish here. I was unable to reproduce it in a way that i can post here. As soon as i manipulate the excel and re-save it without sensible information, the error is gone.

I tried, but without success. Also i don't had time to invest more time.

Sorry.

PR #2312, opened 2021-10-03
Fix offset value on xls

hannanyusop:
"class": "ErrorException",
    "message": "Trying to access array offset on value of type null",
    "code": 0,
    "file": vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xls.php:7015"


hannanyusop:
$data = $this->externalNames[$index - 1]['name'] ?? "";

on line 7020, file Reader/Xls.php

oleibman:
I'm sorry - I do not understand this change at all. It says it has changed 89 files in 35 commits. From your comment, I believe you are trying to change only 1 line in 1 module. If that is so, I think we need to repackage this one. Also, do you have a test file which demonstrates the problem with unchanged code, and proves the fix with your change?

hannanyusop:
i've choose the wrong branch. i just need the change 1 line code

hannanyusop:
on line 7050

oleibman:
Converting this PR to draft status for now. Please resubmit using the correct branch, adding a test case.

oleibman:
Now it's 36 commits and 97 files. Perhaps you should just clone and create a new branch with the one change (and test files).

PowerKiKi:
Not sure what happened here, but this won't be merged as is. Please re-create a new PR with correct branch.

PowerKiKi:
Actually, this is already recreated in #2338

Issue #2273, opened 2021-08-15
Calibri fontnames are all uppercase in Shared/Font.php


cwolcott:
This is:

> A bug report

### What is the expected behavior?
When adding the following lines to my code and pair it with setAuto things should work.
```php       
        Font::setTrueTypeFontPath("/usr/share/fonts/msttcore/");
        Font::setAutoSizeMethod(Font::AUTOSIZE_METHOD_EXACT);
        ...
        // Apply auto size to all columns
        foreach (range('A',$workSheetMaxColumns) as $col) { $worksheet->getColumnDimension($col)->setAutoSize(true); }
```

### What is the current behavior?
When running this on my web server (CentOS) I had 2 issues:
- TrueType fonts were not installed on the CentOS system.  I was able to have our admin install the msttcore fonts.
- Now the real issue is that starting on line 43 some of the fonts are in all uppercase.  Since Linux is case sensitive an exception is raised at line 514 since the font can not be found.  In the msttcore directory all the *.tff files are lowercase.

### Which versions of PhpSpreadsheet and PHP are affected?
version 1.18.0 and master branch

### Suggested Fixes
- Lowercase the CALIBRI font names (Line 48-51) (most other font filenames are all lowercase)
```php       
    const CALIBRI = 'CALIBRI.TTF';
    const CALIBRI_BOLD = 'CALIBRIB.TTF';
    const CALIBRI_ITALIC = 'CALIBRII.TTF';
    const CALIBRI_BOLD_ITALIC = 'CALIBRIZ.TTF';
```
- Change line 512 to lowercase the $fontFile string (probably not the best solution)
`$fontFile = self::$trueTypeFontPath . strtolower($fontFile);`

PowerKiKi:
Fixed in [1.19.0](https://github.com/PHPOffice/PhpSpreadsheet/releases/tag/1.19.0)

PR #2325, opened 2021-10-08
[TASK] Lowercase calibri fontnames

ayacoo:
Related #2273


oleibman:
I usually like to see a unit test demonstrating that the change works. However, such a test here would require that the testing system contain fonts which are often not present on the system, and I see no reason to make the presence of these fonts a requirement for using the package. Indeed, there are no current unit tests for this functionality, probably for this reason. Perhaps it is possible to come up with reasonable tests, but this change need not be delayed awaiting those. It seems very simple, and looks correct.

oleibman:

Issue #2249, opened 2021-08-04
Ability to stream to an Amazon S3 bucket


thmnrtn:
This is:

```
- [] a bug report
- [x] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Able to stream a file to an Amazon S3 bucket using the writer object

### What is the current behavior?
It currently fails on the line below in \PhpOffice\PhpSpreadsheet\Writer\BaseWriter because Amazon S3 buckets don't support fopen in wb+ mode ([see Amazon S3 docs](https://docs.aws.amazon.com/sdk-for-php/v3/developer-guide/s3-stream-wrapper.html)).
```php
<?php
$fileHandle = $filename ? fopen($filename, 'wb+') : false;
```

The error thrown is from Amazon S3 is:

```
Mode not supported: wb+. Use one 'r', 'w', 'a', or 'x'.
```

Using wb mode instead works but I don't have a detailed enough understanding of PhpSpreadsheet to be sure that it wouldn't break anything else.

### What are the steps to reproduce?
Create an spreadsheet object and attempt to wrote to an Amazon S3 bucket using the writer object. 

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

$spreadsheet->getActiveSheet()->setCellValue('Test');

$writer = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet);

$write->save('s3://mybucket/myfile.xlsx');
```

### Which versions of PhpSpreadsheet and PHP are affected?
I'm using PHP 7.3.4 and PHPSpreadsheet 1.18.0

PowerKiKi:
Fixed in [1.19.0](https://github.com/PHPOffice/PhpSpreadsheet/releases/tag/1.19.0)

Issue #1931, opened 2021-03-17
Ability to load files by string or stream


andrei-dascalu:
This is:

```
- [ ] a bug report
- [ x] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

Allow loading by file content or passing a stream and transfer the load responsibility to the user, perhaps as an alternative to the current "by file" loading.

It would be greatly appreciated.

### What is the current behavior?

Currently, in order to load an excel/ods spreadsheet, one needs to provide a full path to a local file. This is inconvenient in a number of situations.

For example, when using a filesystem abstraction that works with cloud storage systems (eg: Flysystem), there is no local file. A workaround can be to make a local copy, work with it and then delete it, but this is inconvenient/risky when using containers that have no local persistence and may have limited to no local storage (consider a long-running service in Kubernetes, you may limit per container storage and exceeding that may break the application).

### Which versions of PhpSpreadsheet and PHP are affected?

PHPSpreadsheet: 1.17
PHP: 7.4

nerdgeneration:
I would also love this feature. The writer will take a file handle, the reader has a bunch of assertFile and path checks. I disagree with the memory overhead when php://temp is an ideal solution to all the above concerns.

toriqsetiawan:
what is the status of this feature? :)

MarkBaker:
> what is the status of this feature? :)

Unlikely

lkmorlan:
Using PhpSpreadsheet from Drupal, I found it supports stream wrappers (protocol handlers) for some formats, such as CSV and XLS. But it does not work in XLSX. It throws "Unable to identify a reader for this file".

This appears to be caused by `ZipArchive::open()` not supporting stream wrappers.

It would be nice to at least get a better error message. The current error makes it sound like the file is the problem.

It might also be an improvement to prevent stream wrappers from working on any formats so that there is consistency. As it is, someone could implement and test on a few formats, everything seems to work fine, and then it breaks when someone uses a different format. However, this would be a breaking change for some users.

Issue #1147, opened 2019-08-27
Load by string


simoheinonen:
Would be nice to be able to create the reader from string. Would be great for example when using a filesystem abstraction like flysystem. (Currently need to write to a temp file and then load that)

4n70w4:
The same problem!

v20100v:
Arg...

no php stream

npelov:
It seams an obvious requirement to be able to load spreadsheet from string/stream

Perturbatio:
FYI, you can capture the file contents using output buffering:

```
use League\Flysystem\Filesystem;
use League\Flysystem\InMemory\InMemoryFilesystemAdapter;
use PhpOffice\PhpSpreadsheet\Spreadsheet;
use PhpOffice\PhpSpreadsheet\Writer\Ods;

$adapter = new InMemoryFilesystemAdapter();
$filesystem = new Filesystem($adapter);

$writer = new Ods($spreadsheet);

ob_start();
$writer->save('php://output');
$content = ob_get_contents();
ob_end_clean();

$filesystem->write('test-memory.ods', $content);
if ($filesystem->has('test-memory.ods')){
    $doc = $filesystem->read('test-memory.ods');
    // do something with the document
} else {
    echo "File not found";
}
```

npelov:
Yes, that's a nice workaround for a feature that should have been there.

MarkBaker:
> Yes, that's a nice workaround for a feature that should have been there.

Should have been there! But simply not possible. Streams, by their nature are linear, read from start to finish; but aside from HTML and CSV, none of the spreadsheet file formats are linear. You can't start reading the file at the beginning and process it to the end. Those files have to be accessed in a way that allows you to jump backwards and forwards within the file, reading different bits of content from different parts of the file.

The only way that we could accept string or stream content would be to create a temporary file internally, and then process it; simply moving that "workround" from outside of the library to inside it.

npelov:
Hmm. I see. I haven't looked in in details how save works. But there is one better workaround:
```
    $fd = fopen("php://memory", "w+b");
    $writer->save($fd);
    rewind($fd);
    $xlsx = stream_get_contents($fd);
    fclose($fd);
```
Using the output buffer is not ideal. I think php://memory (since php 5.1) stream is a bit better. You could include that in the library. It doesn't require writable space on disk and doesn't mess with output buffer. I'm not sure if having gzip filters for output buffer would affect the php://output solution...

[...]

PR #2326, opened 2021-10-08
[FEATURE] Ability to stream to an Amazon S3 bucket

ayacoo:
Related #2249

This is:

```
- [ ] a bugfix
- [x] a new feature
```

**Requirements**
`composer required aws/aws-sdk-php`

**Example**
```
require '../vendor/autoload.php';

$client = new Aws\S3\S3Client([
    'credentials' => [
        'key' => 'KEY',
        'secret' => 'SECRET',
    ],
    'default_cache_config' => '',
    'certificate_authority' => true,
    'region' => 'REGION',
    'version' => 'latest'
]);

// Register the stream wrapper from an S3Client object
$client->registerStreamWrapper();

$writer->save('s3://BUCKETNAME/myfile.xlsx');
```

### DIFF
--- a/src/PhpSpreadsheet/Shared/Font.php
+++ b/src/PhpSpreadsheet/Shared/Font.php
@@ -45,10 +45,10 @@ class Font
     const ARIAL_ITALIC = 'ariali.ttf';
     const ARIAL_BOLD_ITALIC = 'arialbi.ttf';
 
-    const CALIBRI = 'CALIBRI.TTF';
-    const CALIBRI_BOLD = 'CALIBRIB.TTF';
-    const CALIBRI_ITALIC = 'CALIBRII.TTF';
-    const CALIBRI_BOLD_ITALIC = 'CALIBRIZ.TTF';
+    const CALIBRI = 'calibri.ttf';
+    const CALIBRI_BOLD = 'calibrib.ttf';
+    const CALIBRI_ITALIC = 'calibrii.ttf';
+    const CALIBRI_BOLD_ITALIC = 'calibriz.ttf';
 
     const COMIC_SANS_MS = 'comic.ttf';
     const COMIC_SANS_MS_BOLD = 'comicbd.ttf';


oleibman:
I usually like to see a unit test demonstrating that the change works. However, such a test seems impractical here - it requires access to s3, which probably introduces security and availability challenges. It seems clear to me that the change will have no adverse effect for non-s3 access.

alexanderkraemer:
This change makes a lot of sense! Looking forward for it getting merged!

warespace-silas-winter:
Hi everyone, sorry for reopening the pull request.

I see that the S3 protocol works for the writer. Could s3:// be a workaround for https://github.com/PHPOffice/PhpSpreadsheet/issues/1147 and https://github.com/PHPOffice/PhpSpreadsheet/issues/1931? At least for me it would be great.

oleibman:
I think s3:// would work in much the same way as http://. Unfortunately, http:// doesn't work for read. It will fail in `Shared\File::assertFile` because `is_file` will return false. And, even if I change the code to bypass that check, it will will fail in `Reader\Xlsx::loadSpreadsheetFromFile` because `$zip->open` will fail and return `ZipArchive::ER_READ`.

However, the following code may be useful to you (substitute your own s3 filename - I have no way to test that):
```php
        $filename = 'http://localhost:8000/discussion.4090.xlsx';
        $contents = file_get_contents($filename);
        $tmpfile = tmpfile();
        $path = stream_get_meta_data($tmpfile)['uri'];
        echo "path is $path\n";
        fwrite($tmpfile, $contents);
        fseek($tmpfile, 0);
        $spreadsheet = $reader->load($path);
        fclose($tmpfile);
```
The temporary file should be deleted by the fclose, and you can always write some sort of handler to force its deletion even if the spreadsheet load fails.

oleibman:
A possible slight improvement to the code above:
```php
        $filename = 'http://localhost:8000/filename.xlsx';
        $contents = file_get_contents($filename);
        $tmpfile = tmpfile();
        $path = stream_get_meta_data($tmpfile)['uri'];
        echo "path is $path\n";
        try {
            fwrite($tmpfile, $contents);
            fseek($tmpfile, 0);
            $spreadsheet = $reader->load($path);
        } finally {
            fclose($tmpfile);
        }
```

PR #2343, opened 2021-10-19
Fix array formula values in XLSX writer

RomanDevman:
Restored commented out feature to write array formula.

This is:

```
- [ ] a bugfix
- [+] a new feature
```

Checklist:

- [ ] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
@@ -1205,27 +1205,26 @@ class Worksheet extends WriterPart
             $objWriter->writeAttribute('t', 'b');
             $calculatedValue = (int) $calculatedValue;
         }
-        // array values are not yet supported
-        //$attributes = $pCell->getFormulaAttributes();
-        //if (($attributes['t'] ?? null) === 'array') {
-        //    $objWriter->startElement('f');
-        //    $objWriter->writeAttribute('t', 'array');
-        //    $objWriter->writeAttribute('ref', $pCellAddress);
-        //    $objWriter->writeAttribute('aca', '1');
-        //    $objWriter->writeAttribute('ca', '1');
-        //    $objWriter->text(substr($cellValue, 1));
-        //    $objWriter->endElement();
-        //} else {
-        //    $objWriter->writeElement('f', Xlfn::addXlfnStripEquals($cellValue));
-        //}
-        $objWriter->writeElement('f', Xlfn::addXlfnStripEquals($cellValue));
-        self::writeElementIf(
-            $objWriter,
-            $this->getParentWriter()->getOffice2003Compatibility() === false,
-            'v',
-            ($this->getParentWriter()->getPreCalculateFormulas() && !is_array($calculatedValue) && substr($calculatedValue ?? '', 0, 1) !== '#')
-                ? StringHelper::formatNumber($calculatedValue) : '0'
-        );
+
+        $attributes = $cell->getFormulaAttributes();
+        if (($attributes['t'] ?? null) === 'array') {
+            $objWriter->startElement('f');
+            $objWriter->writeAttribute('t', 'array');
+            $objWriter->writeAttribute('ref', $cell->getCoordinate());
+            $objWriter->writeAttribute('aca', '1');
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php
```diff
@@ -1260,27 +1260,25 @@ private function writeCellFormula(XMLWriter $objWriter, string $cellValue, Cell
...
-        //    $objWriter->startElement('f');
-        //    $objWriter->writeAttribute('t', 'array');
-        //    $objWriter->writeAttribute('ref', $pCellAddress);
-        //    $objWriter->writeAttribute('aca', '1');
-        //    $objWriter->writeAttribute('ca', '1');
-        //    $objWriter->text(substr($cellValue, 1));
-        //    $objWriter->endElement();
-        //} else {
-        //    $objWriter->writeElement('f', Xlfn::addXlfnStripEquals($cellValue));
-        //}
```
PowerKiKi, line 1275: @oleibman could you please let us know why you commented out this part ? was it a mistake ?

oleibman: Yike, this change was made over a year and a half ago and I haven't thought about it in the interim. Nevertheless ...

It was certainly a mistake to not describe this particular aspect of the change in the commit message. I believe (and this could certainly be self-serving) that I noticed a problem early on and intended the commenting out to let me work on the bulk of the change, intending to return to it later, but neglecting to do so. Nevertheless, the results of uncommenting the statements will often not be successful. In Excel, if I enter `=MINVERSE({2,0;0,1})` into cell A1, you will get a `dynamic array` (which we do not yet support) - A1 will contain 0.5, A2 and B1 will contain 0, and B2 will contain 1. There are also `spill` implications for such a formula. The XML for these cells will be:
``` xml
<row r="1" spans="1:2" x14ac:dyDescent="0.3">
<c r="A1" cm="1">
<f t="array" ref="A1:B2">MINVERSE({2,0;0,1})</f>
<v>0.5</v>
</c>
<c r="B1">
<v>0</v>
</c>
</row>
<row r="2" spans="1:2" x14ac:dyDescent="0.3">
<c r="A2">
<v>0</v>
</c>
<c r="B2">
<v>1</v>
</c>
</row>
```

I believe that the PhpSpreadsheet equivalent of doing this (with the statements uncommented) is:
```php
        $calculation = Calculation::getInstance($spreadsheet);
        $calculation::setArrayReturnType(Calculation::RETURN_ARRAY_AS_ARRAY);
        $sheet->getCell('A1')->setValue('=MINVERSE({2,0;0,1})');
        $oufil = 'issue.2343.xlsx';
        $writer->save($oufil);
```
But our output file only fills in A1:
```xml
<row r="1" spans="1:1">
<c r="A1">
<f>MINVERSE({2,0;0,1})</f>
<v>0.5</v>
</c>
</row>
```
And, even though A1 has its correct value, note that its `f` tag does not have a `t` attribute. This is because we never set any formula attributes, except in Xlsx Reader (see next paragraph), so we do not encounter the `'array'` condtion for a formula newly added to a spreadsheet.

We do slightly better when we read the first file (as opposed to creating a new spreadsheet), but we succeed only by accident. Because B1, A2, and B2 are assigned values in the XML, all 4 cells will have the expected values. But they are now independent of each other, not part of a dynamic array. When we write this out, it is almost correct:
```xml
<row r="1" spans="1:2">
<c r="A1">
<f>MINVERSE({2,0;0,1})</f>
<v>0.5</v>
</c>
<c r="B1">
<v>0</v>
</c>
</row>
<row r="2" spans="1:2">
<c r="A2">
<v>0</v>
</c>
<c r="B2">
<v>1</v>
</c>
</row>
```
Again, the `f` tag has no `t` attribute, and it doesn't seem to matter whether we set RETURN_TYPE_ARRAY or not. I think this particular aspect of the problem might be relatively easy to fix.

If there is a problem which uncommenting this code solves, I certainly have no objection. But it should at least have a test case. And anything which we do now will probably have to be redone if and when we tackle dynamic arrays.

PowerKiKi:
I merged this for the time being, even though it is not perfect, and array formula really are not supported in PhpSpreadsheet...

Issue #2347, opened 2021-10-21
PHPToExcel::timestampToExcel uses GMT timezone 


Bhoft:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

I am using the `Date::PHPToExcel($ts);` with a unix timestamp and was wondering why the timezone which i set wasn't used.

### What is the current behavior?

My current timezone set in my application is 'Europe\Berlin' i add some cells with datetimes to the Excel file and the Timezone of those fields isn't correct when i used unix timestamps.

### What are the steps to reproduce?

```php
<?php
require_once __DIR__ . '/vendor/autoload.php';

ob_start();
// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

// add code that show the issue here...
date_default_timezone_set('Europe/Berlin');

\PhpOffice\PhpSpreadsheet\Shared\Date::setDefaultTimezone(date_default_timezone_get());
\PhpOffice\PhpSpreadsheet\Shared\TimeZone::setTimeZone(date_default_timezone_get());

$dateString = '2020-04-01 00:00:00';
$unixTimestamp = strtotime($dateString);
var_dump($unixTimestamp);

$dateObject = DateTime::createFromFormat("U", $unixTimestamp)->setTimeZone(new DateTimeZone(date_default_timezone_get()));
var_dump($dateObject);

#draw the specified date time to excel
$worksheet = $spreadsheet->setActiveSheetIndex(0);


#displays 1/4/20 0:00
$worksheet
    ->getCellByColumnAndRow(1, 1)
    ->setValue(\PhpOffice\PhpSpreadsheet\Shared\Date::PHPToExcel($dateString))
    ->getStyle()
    ->getNumberFormat()
    ->setFormatCode(\PhpOffice\PhpSpreadsheet\Style\NumberFormat::FORMAT_DATE_DATETIME);

#displays: 1/4/20 0:00
$worksheet
    ->getCellByColumnAndRow(1, 2)
    ->setValue(\PhpOffice\PhpSpreadsheet\Shared\Date::PHPToExcel($dateObject))
    ->getStyle()
    ->getNumberFormat()
[...]

PR #2350, opened 2021-10-24
Update Doc Blocks to Discourage Use of Unix Timestamps

oleibman:
This was suggested by issue #2347. Unix timestamps have clear disadvantages compared with the alternate methods of supplying date and time to PhpSpreadsheet - DateTime objects, Excel date time fields, and strings. In particular, Unix timestamp is not Y2038-safe on a 32-bit system, and it reflects the time in UTC, which may come as a surprise to the end-user (as it did in the cited issue). The alternate methods do not come with such baggage. This change updates some doc blocks to note that Unix timestamps are discoburage (N.B. - not deprecated). No executable code is changed.

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] doc block updates
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Shared/Date.php
```diff
@@ -206,11 +206,12 @@ public static function excelToDateTimeObject($excelTimestamp, $timeZone = null)
 
     /**
      * Convert a MS serialized datetime value from Excel to a unix timestamp.
+     * The use of Unix timestamps, and therefore this function, is discouraged.
```
PowerKiKi, line 209: While the commit message explain the situation well, I'd expect to have a least a succinct explanation in the code itself, as it will most likely be the only thing that end-users see

PowerKiKi (changes requested):

src/PhpSpreadsheet/Shared/Date.php
oleibman: Done.

PR #2383, opened 2021-11-09
Added setter for HTMLPurifier config in HTML writer

vladdnepr:
This is:

```
- [ ] a bugfix
- [* ] a new feature
```

Checklist:

- [*] Code style is respected
- [*] CHANGELOG.md contains a short summary of the change
- [*] Documentation is updated as necessary

Sometimes default config values of HTMLPurifier are not right.

For example default cache path in prod env is not writeable. But we have app cache folder, where we can write cache data.

Снимок экрана 2021-11-09 в 11 53 56

Now we can set custom config and change what we need.


vladdnepr:
Please describe why Scrutinizer did fail?

vladdnepr:
@MarkBaker @PowerKiKi hi. Can somebody merge this PR?

oleibman:
This seems unnecessarily complicated; also, if the default code potentially writes to a protected directory, as your explanation makes clear that it might, we are better off solving that problem for everyone:
```php
            $sanitizer = new HTMLPurifier();
            $cachePath = File::sysGetTempDir() . '/phpsppur';
            if (is_dir($cachePath) || mkdir($cachePath)) {
                $sanitizer->config->set('Cache.SerializerPath', $cachePath);
            }
```
We use a similar approach for Mpdf. I think this would solve your problem without the need for changes at the application end, and without the need for extra properties and methods and documentation.

vladdnepr:
> This seems unnecessarily complicated; also, if the default code potentially writes to a protected directory, as your explanation makes clear that it might, we are better off solving that problem for everyone:
> 
> ```
>             $sanitizer = new HTMLPurifier();
>             $cachePath = File::sysGetTempDir() . '/phpsppur';
>             if (is_dir($cachePath) || mkdir($cachePath)) {
>                 $sanitizer->config->set('Cache.SerializerPath', $cachePath);
>             }
> ```
> 
> We use a similar approach for Mpdf. I think this would solve your problem without the need for changes at the application end, and without the need for extra properties and methods and documentation.

Done. Please check

oleibman:
Getting closer, but ... I would prefer that you not mess around with the PDF code. It works just fine now, and has nothing to do with your change. There is no reason to think PDF and Purifier use, and always will use, the temporary directory in compatible manners. Html does not have a tempdir because it almost never needs one; any spreadsheet lacking comments will still have to test the existence of and possibly allocate a new directory that it will not use, because of the code you've added to the constructor.

vladdnepr:
> Getting closer, but ... I would prefer that you not mess around with the PDF code. It works just fine now, and has nothing to do with your change. There is no reason to think PDF and Purifier use, and always will use, the temporary directory in compatible manners. Html does not have a tempdir because it almost never needs one; any spreadsheet lacking comments will still have to test the existence of and possibly allocate a new directory that it will not use, because of the code you've added to the constructor.

Done, please check

oleibman:
Thank you. That looks good. I will merge it some time over the next few days.

Issue #2378, opened 2021-11-06
AUTOFILTER_COLUMN_RULE_NOTEQUAL is actual showing EQUAL


cwolcott:
This is:

```
- [X] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Creating an autofilter rule of **NOTEQUAL** should hide the rows that match the rule.
```
$autoFilter = $worksheet->getAutoFilter();
$columnFilter = $autoFilter->getColumn('G');
$columnFilter->setFilterType(Column::AUTOFILTER_FILTERTYPE_FILTER);
$columnFilter->createRule()->setRule(Rule::AUTOFILTER_COLUMN_RULE_NOTEQUAL, 'LC');
```
### What is the current behavior?
It is showing the rows that are **EQUAL** to 'LC'

### What are the steps to reproduce?
Create a spreadsheet and attempt to filter column NOTEQUAL to a specific text value.

### Which versions of PhpSpreadsheet and PHP are affected?
1.18.0

### Images
Spreadsheet view after performing **NOTEQUAL**, it is actually showing **EQUAL**
```$columnFilter->createRule()->setRule(Rule::AUTOFILTER_COLUMN_RULE_NOTEQUAL, 'LC');```
<img width="368" alt="AUTOFILTER_COLUMN_RULE_NOTEQUAL" src="https://user-images.githubusercontent.com/3777526/140607285-4917e7d8-865b-4c16-aee0-f9252ad6356c.png">

This is what the spreadsheet looks like with no AutoFilter applied (Note: EW rows are further down in the spreadsheet and aren't show in the screen capture)
<img width="368" alt="No AUTOFILTER applied" src="https://user-images.githubusercontent.com/3777526/140607350-118af2e7-0295-417f-9f28-b91bb6b38149.png">

oleibman:
On its own, when Excel creates a sheet with an EQUAL filter, the XML uses a `<filters>` tag followed by `<filter>` without attribute `operator`; when it creates a sheet with a NOTEQUAL filter, it uses a `<customFilters>` tag followed by `<customFilter operator='notEqual' ...>`. PhpSpreadsheet is currently generating `<filters>` in both cases, leading to your result. I need to research when exactly `<customFilters>` is required.

oleibman:
[...]

PR #2392, opened 2021-11-11
AutoFilter Improvements

oleibman:
Fix issue #2378. The following changes are made:
- NotEqual tests must be part of a custom filter. Documentation has been changed to indicate that.
- Method setAndOr was replaced by setJoin some time ago. Documentation now reflects that change.
- Documentation to indicate that string filters are not case-sensitive, same as in Excel.
- Filters testing against numeric value now include a numeric test (not numeric for not equal, numeric for all others).
- String filter had previously treated everything as a test for "equal". It now handles "not equal" and the variants of "greater/less" with or without "equal".
- Documentation correctly stated that no more than 2 rules are allowed in a custom filter. Code did not enforce this restriction. It now does, throwing an exception if an attempt is made to add a third rule.
- Deleted a lot of comments in Rule.php to make it easier to see what is not yet implemented (between, begins with, etc.). I may take these on in future.
- Added a number of tests for the new functionality.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/docs/topics/autofilters.md
+++ b/docs/topics/autofilters.md
@@ -99,6 +99,8 @@ results are unpredictable.
 Other filter expression types (such as cell colour filters) are not yet
 supported.
 
+String comparisons in filters are case-insensitive.
+
 ### Simple filters
 
 In MS Excel, Simple Filters are a dropdown list of all values used in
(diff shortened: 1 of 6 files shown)


oleibman:
Another undetected merge conflict! Looks like I will have to rebase and try again, probably tomorrow.

oleibman:
Closing, superseded by #2393.

PR #2399, opened 2021-11-17
Rounding in NumberFormatter

### LINKED ISSUE #2385: Loaded excel file didn't round decimal value correctly.
This is:

```
- [ x ] a bug report
- [ ] a feature request
```

Decimal value of 2,5 in excel got rounded to 3 when using Number cell format.

When excel data is being retrieved using toArray(), the decimal value of 2,5 got rounded to 2 instead of 3. The rounding only works when the digit before the comma is an odd number (eg. 1,5 will be rounded to 2 and 23,5 will be rounded to 24 while 2,5 will be rounded to 2 and 22,5 will be rounded to 22 instead).

All I did was simply load the excel file
```
$spreadsheet = $reader->load($request->file('file'));

$sheetData = $spreadsheet->getActiveSheet()->toArray();

var_dump($sheetData);
```

And the cell which is formatted to Number (Format code is "0") with the decimal value came out wrong as described in the current behavior :
```
[13]=> string(1) "2"
```

Here's the excel if needed. The value I mention is on cell N7.
[template_Bestorq (#21542) RING.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/7509960/template_Bestorq.21542.RING.xlsx)

PhpSpreadsheet 1.19.0 and using PHP 7.4

oleibman:
Fix #2385. NumberFormatter is using sprintf on a float, and is seeing inconsistent rounding as a result (it will also occasionally result in `-0`). Change to round the number before presenting it to sprintf.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


nohn:
Tests that seem to provoke that strange rounding behaviour are in https://github.com/nohn/PhpSpreadsheet/tree/strange_rounding_format_value_with_mask

PR #2433, opened 2021-12-02
Respect DataType in insertNewBefore

mfuerni:
`ReferenceHelper::insertNewBefore` copies data from one cell to another when adding/removing rows or columns.
It now also respects the data type set for that cell and does not use value binder again.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

When inserting or removing new rows, the data type of a cell is currently not respected. This can lead to wrong data e.g. in cells of type string, that are interpreted as number by using `Cell::setValue`.


oleibman:
Can you add unit test(s) which (a) would not have worked as desired before your change and (b) will work as desired with it.

mfuerni:
> Can you add unit test(s) which (a) would not have worked as desired before your change and (b) will work as desired with it.

I added a simple test, is that enough?
Hope I did it in the right way.

oleibman:

PR #2435, opened 2021-12-04
Phpstan and Php8 Phase 2

oleibman:
PR #2428 reduced the count of Phpstan errors reported under Php8 from 218 to 13. This PR takes care of those 13, which could not necessarily be handled merely by tweaks to the configuration files.

The commonest problem was that Phpstan under Php7 thinks date_parse can return an array or false, but, under Php8, it thinks it can return only array. The documentation agrees with the Php7 interpretation, but I have not been able to construct a case, in any release of Php, where date_parse returns false. In case of error, it sets elements 'error' and 'error_count' in the returned array. Code is changed to handle either possibility in a manner of which Phpstan will approve.

Other problems are handled through type-hinting.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


oleibman:
No intention of acting on Scrutinizer failure. Added 3 methods to Helpers, each with a complexity rating of A. This pushes overall module complexity from B to C. Could split Helpers up, but can't see any compelling reason to do so; that would make things less maintainable, not more so.

PR #2458, opened 2021-12-21
Richtext colors are not being read correctly after namespace change

daedeloth:
Related to:
https://github.com/PHPOffice/PhpSpreadsheet/pull/2303

This is:
- [x] a bugfix
- [ ] a new feature

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx.php
@@ -1689,7 +1689,7 @@ class Xlsx extends BaseReader
                             $objText->getFont()->setSize((float) $attr['val']);
                         }
                         $attr = $run->rPr->color->attributes();
-                        if (count($attr) > 0) {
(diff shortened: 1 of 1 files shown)


oleibman:
I agree with your analysis that the namespace change caused the problem, and that the handling of attributes is the specific problem. However, your change is misplaced; RichText is not the only user of ReadColor. The fix belongs in Xlsx/Reader/Styles.php function readColor. On entry, it should have something like:
```php
$attr = Xlsx::getAttributes($color)
````
Then replace all further uses of `$color` with `$attr` in that function. I believe this would solve your problem, and eliminate the failure that occurred when running unit tests on your package.

You should also add a unit test that verifies that the code is working correctly.

**Comment by oleibman** [comment]
It looks like a better initialization in function readColor is:
```php
$attr = ($color && $color->attributes()) ? $color->attributes() : [];
```

PR #2466, opened 2021-12-26
Allow for any version of psr/simple-cache

tacman:
This is:

```
- [ X ] a bugfix
```

Checklist:

- [ ] Code style is respected

This package should support the latest versions of the cache dependency.


oleibman:
Not possible till we no longer need to support Php7.

ruudk:
Wouldn't it be better to drop support for PHP 7 altogether? PHP 7.3 is EOL and PHP 7.4 is Security Support only.

PR #2485, opened 2022-01-04
Handle Explicit "Date" Type for Cell

### LINKED ISSUE #2373: PhpOffice\PhpSpreadsheet\Exception: Invalid datatype: d
Hello

I am using v1.19.0
I am openning an existing Xlsx file and I want to write something in a new sheet

The following error occurs:
```
type: PhpOffice\PhpSpreadsheet\Exception
file: /opt/PhpSpreadsheet/src/PhpSpreadsheet/Cell/Cell.php
line: 233
message: Invalid datatype: d
trace: 
#0 /opt/PhpSpreadsheet/src/PhpSpreadsheet/Reader/Xlsx.php(828): PhpOffice\PhpSpreadsheet\Cell\Cell->setValueExplicit('2021-03-29', 'd')
#1 myscript.php(xxx): PhpOffice\PhpSpreadsheet\Reader\Xlsx->load('...')
```

I can share the file in private

oleibman:
Fix #2373. Excel can handle DateTime/Date/Time as a string if the datatype of the cell is set to "d". The string is, apparently, supposed to follow the ISO8601 spec. Openpyxl can be configured to generate a file with such values, so I've added support and set up unit tests. Excel, naturally, converts such a string input into its numeric representation of the date/time stamp. So will PhpSpreadsheet, so a call to setValueExplicit specifying Date format will actually see the cell wind up with Numeric format - there is no way (and no reason) for the Date type to 'stick'.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


oleibman:
No plans to address Scrutinizer "complexity" problem. Added a new datatype that requires processing; some additional complexity is to be expected.

PR #2487, opened 2022-01-06
Refinement for XIRR

### LINKED ISSUE #2469: XIRR financial function not calculating as in Excel
This is:

```
- [X] a bug report
- [ ] a feature request
```

XIRR should be able to return a value -0.6422991 for this simple use case

XIRR returns #NUM!

```php
<?php

use PhpOffice\PhpSpreadsheet\Calculation\Financial\CashFlow\Variable\NonPeriodic;

require __DIR__."/vendor/autoload.php";

ini_set('memory_limit', '8192M');
ini_set('display_errors', 1);
ini_set('log_errors', true);

$cashFlows = [55600, -51094.83];
$cashFlowDates = ['24/11/2021', '24/12/2021'];

# Excel returns -0.6422991
echo NonPeriodic::rate($cashFlows, $cashFlowDates).PHP_EOL;
```

PhpSpreadsheet 1.20.0 (tested with 1.15 also using pre namespace changes)
PHP 7.4 and 8.0

oleibman:
Fix #2469. The algorithm used for XIRR is known not to converge in some cases, some of which are because the value is legitimately unsolvable; for others, using a different guess might help.

The algorithm uses continual guesses at a rate to hopefully converge on the solution. The code in Python package xirr (https://github.com/tarioch/xirr/) suggests a refinement when this rate falls below -1. Adopting this refinement solves the problem for the data in issue 2469 without any adverse effect on the existing tests. My thanks to @tarioch for that refinement. The Python project cites https://github.com/RayDeCampo/java-xirr as a reference. Thanks also to @RayDeCampo.

The data from 2469 is, of course, added to the test cases. The user also mentions that an initial guess equal to the actual result doesn't converge either. A test is also added to confirm that that case now works.

The test cases are changed to run in the context of a spreadsheet rather than by direct calls to XIRR calculation routine. This revealed some data validation errors which are also cleaned up with this PR. This suggests that other financial tests might benefit from the same change; I will look into that.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


blacknell:
Great work. For what it's worth, I re-ran my workflow on my dataset with the fixes and it worked a charm. Previously, I was getting 285 failed convergences out of a total of 7,177 calls to the XIRR method. Now, there are zero.

Furthermore, these were masking an additional 15 failures caused by bad data on my part which I've now been able to fix.

oleibman:
Wonderful, thank you for the confirmation. I hope to be able to merge this by the weekend.

PR #2529, opened 2022-01-26
Reduce memory usage of Cells collection

eigan:
- Store coordinates in 4-byte array
- Cells::getSortedCoordinates() will now not increase memory usage

I have extracted the Cells part of #2228 and done some modifications to it:
- Removed two extra methods (hasAnyCellInRow, getCellsByRow)
- Early return  in Cells::add() to avoid the `sscanf` call when not needed.
- getSortedCoordinates and getCoordinates() is no longer generators. Felt this was an unnecessary breaking change. Besides, it didnt reduce memory usage for me while testing anyway.
- Revert rename of method signature variables $cellCoordinate to $pCoord and $worksheet to $parent.

This is:

```
- [x a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Collection/Cells.php
+++ b/src/PhpSpreadsheet/Collection/Cells.php
@@ -45,9 +45,9 @@ class Cells
     private $currentCellIsDirty = false;
 
     /**
-     * An index of existing cells. Booleans indexed by their coordinate.
+     * An index of existing cells. Cells indexed by 4-bytes (64 cells) blocks.
      *
-     * @var bool[]
+     * @var int[][]
      */
     private $index = [];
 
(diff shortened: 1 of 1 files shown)


eigan:
To test these changes did I add some memory logging before and after `createStringTable()` which uses ` Cells::getSortedCoordinates`:

### Before changes
``` 
36.5MB: Make string table  
42.5MB: End make string table 
```

### After changes
```
36.5MB: Make string table  
36.5MB: End make string table 
```

The overall peak memory was reduces with the same amount (6MB).

eigan:
I noticed a side-effect. 

`Cells::getCoordinates(false)` was previously returning in the order the cells was added, now the cells are returned by column, while `Cells::getCoordinates(true)` returns by row (as before).


#### Before
```php
Cells::getCoordinates(false) === ['A1', 'B1', 'C1', 'A2', 'C2', 'C3'];
``` 

#### After
```php
Cells::getCoordinates(false) === ['A1', 'A2', 'B1', 'C1', 'C2', 'C3'];
``` 

While it might not break the contract(?) will it reduce the usefulness of the cache I wanted to create. The cache would hold a number of rows in memory, while the rest was stored on disk. This only works if PHPSpreadsheet handles the cells in the same order as I am working with them (row-by-row).

PR #2530, opened 2022-01-26
Always access cells in the same order as they was inserted

eigan:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

This will make it easier to create an optimized cache.

As a sideeffect, not calling Cells::getSortedCoordinates() reduces some
memory usage as the current state of that method allocates a lot of memory (see #2529)

⚠️ I have not created the cache impl yet, so not sure how well it will perform regardless of these changes. But the memory savings is nice.

### DIFF
--- a/src/PhpSpreadsheet/Writer/Xls.php
+++ b/src/PhpSpreadsheet/Writer/Xls.php
@@ -161,7 +161,7 @@ class Xls extends BaseWriter
 
         // add fonts from rich text eleemnts
         for ($i = 0; $i < $countSheets; ++$i) {
-            foreach ($this->writerWorksheets[$i]->phpSheet->getCoordinates() as $coordinate) {
+            foreach ($this->writerWorksheets[$i]->phpSheet->getCoordinates(false) as $coordinate) {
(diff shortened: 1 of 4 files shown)


MarkBaker:
Hmm!

Trying to run the following very simple code against this branch:
```php
$data1 = [
    ['Key', 'Value'],
    [1, 2],
    [6, 64],
    ['=SUM($A$2:$A3)', '=SUM($B$2:$B3)'],
];

$worksheet->fromArray($data1, null, 'A1', true);

$data2 = [
    [2, 4],
    [3, 8],
    [4, 16],
    [5, 32],
];

$worksheet->insertNewRowBefore(3,4);
$worksheet->fromArray($data2, null, 'A3', true);
```
and then writing to an `Xlsx` file results in a file that won't load cleanly, and that is missing data when I tell it to recover what it can anyway. It isn't a problem with other formats; but this is an issue for the `Xlsx` Writer.

And that is typical of user code; reading a template file with formulas for the totals already in the bottom row, then inserting new rows of data to be included in those totals.

eigan:
Feel free to close this. I refactored code to use box/Spout because we didn't really need all the stuff in this library.

PR #2539, opened 2022-01-30
Calculation engine array formulae initial work

MarkBaker:
This is:

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [X] Code style is respected

Initial support for reading/writing array formulae with the Xlsx Reader/Writer

Work to allow setting array formulae from userland code looks like it will create a BC break. It requires extra arguments for the Cell `setValue()` and `setValueExplicit()` methods - the equivalent of pressing Ctrl-Enter when entering an array formula in MS Excel - (unless I can come up with some creative alternative, like executing the formula at the point when it is set to identify if it is an array formula or not; and if so, what the spillage cell range should be); and these additional arguments must be passed to the Cell Binder creating a BC break for the BinderInterface.


oleibman:
Unrelated to this PR, I have come upon something that could be categorized as a security problem. I do not see how to fix it for the moment, and, more importantly, I am not sure how to report it. What would be the best way to communicate about it?

MarkBaker:
EMail me (mark at angel-demon dot eu)

oleibman:
550: Unrouteable address

MarkBaker:
Sorry, it was  demon-angel, not angel-demon

Issue #2535, opened 2022-01-28
[Bug] Removing column shifts removed column's value into next column if next column value is empty


orkhanahmadov:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

When removing a column from the worksheet, all remaining column and row values should stay in place, including empty and `null` values.

### What is the current behavior?

When removing a column before the last column, if the next column's value is empty or null instead of keeping that column's value it shifts the removed column's value into that column.

### What are the steps to reproduce?

Take following data as example:

```php
// file.csv
[
  ['a1', 'b1', ''],
  ['a2', 'b2', 'c2'],
]
```

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$reader = new \PhpOffice\PhpSpreadsheet\Reader\Csv();
$reader->load('file.csv');

$worksheet = $reader->getActiveSheet();

// remove column B
$worksheet->removeColumn('B');

// Output result
var_dump($worksheet->toArray());
```

The result ends like this:

```php
[
  ['a1', 'b1'],
  ['a2', 'c2'],
]
```

But it must be like this:

```php
[
  ['a1', ''],
  ['a2', 'c2'],
]
```

### Which versions of PhpSpreadsheet and PHP are affected?

- `phpoffice/phpspreadsheet@1.21`

### Cause

With CSV reader when loading file with `loadIntoExisting()` method, following line checks if "cell" value is empty, and prevents inserting if it is empty

https://github.com/PHPOffice/PhpSpreadsheet/blob/e7b0497237080284f35d91a3a086de52035b3cfc/src/PhpSpreadsheet/Reader/Csv.php#L325

### Possible solution

We can make a property and setter like:

```php
protected $preventSettingEmptyValues = true;

public function setPreventSettingEmptyValues($preventSettingEmptyValues)
{
  $this->preventSettingEmptyValues = (bool) $preventSettingEmptyValues;

  return $this;
}
```

And change that `if` statement to:

```php
if (($this->preventSettingEmptyValues && $rowDatum !== '') && $this->readFilter->readCell($columnLetter, $currentRow)) {
  ...
}
```

I'll be happy to create a PR with the solution if you agree with it.

[...]

PR #2541, opened 2022-01-31
`ReferenceHelper@insertNewBefore` checks for missing coordinates before replacing values

orkhanahmadov:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected
- [x] Documentation is updated as necessary - not necessary

Relates to [this issue](https://github.com/PHPOffice/PhpSpreadsheet/issues/2535). When deleting a column before the last column with the null value it incorrectly keeps deleted columns value instead of copying the null/empty value from the last column. This PR fixes this issue by checking for missing coordinates and merging them with `$allCoordiantes` in `insertNewBefore()`.


orkhanahmadov:
> For completeness, can you change your test to assert all the values in A1-C2 after removeCol rather than just B2?.

Good point. I made changes to my PR, added assertions for cells in the test case. I also switched to array value checking instead of using `getCell()`, discovered that `getCell()` creates the cell if it is not available in the worksheet. If I used `getCell()` in a test case that solved the bugfix on its own, without any actual code changes. I modified my implementation to use `createNewCell` on the worksheet to create missing cells, but this required changing the visibility of the method from `private` to `public`

oleibman:

PR #2564, opened 2022-02-07
Fix documentation, instantiation example

jr-k:
This is:

- [X] a bugfix
- [ ] a new feature

Checklist:

- [ ] Code style is respected

Minor documentation fix


MarkBaker (approved):
Hmm! Yes, thanks for correcting this.

I'm glad that some people actually read the documentation, especially when they notice silly errors like this

Issue #353, opened 2018-01-30
Inline strings are not written into spreadsheet.


PowerGamer1:
### What is the expected behavior?
The generated spreadsheet contains text 'inline str' in A1 cell. The text appears directly in the cell inside `sheet1.xml` (and not in `sharedStrings.xml`).

### What is the current behavior?
A1 cell of generated spreadsheet is empty.

### What are the steps to reproduce?
```PHP
<?php
require __DIR__ . '/vendor/autoload.php';
$doc = new \PhpOffice\PhpSpreadsheet\Spreadsheet();
$sheet = $doc->getSheet(0);
$sheet->getCellByColumnAndRow(1, 1)->setValueExplicit('inline str', \PhpOffice\PhpSpreadsheet\Cell\DataType::TYPE_INLINE);
$sheet->getCellByColumnAndRow(2, 1)->setValueExplicit('shared str', \PhpOffice\PhpSpreadsheet\Cell\DataType::TYPE_STRING);
$writer = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($doc, 'Xlsx');
$writer->save('1.xlsx');
```
### Which versions of PhpSpreadsheet and PHP are affected?
1.0.0.

When a cell value is set with a `DataType::TYPE_INLINE` the writer should write such string into the spreadsheet as inline string (which goes directly into the content of a cell rather than shared string table). See 18.3.1.4 and 18.18.11 in ISO/IEC 29500-1 for more info.

PowerKiKi:
I can confirm what you described. But what is the use-case for using inline strings instead of shared one ?

PowerGamer1:
1. According to spec inline string MUST be placed into the cell itself, not string table, so as long as your library allows setting string type as inline it must adhere to spec when writing such strings into XML - and use case does not matter.

2. The whole point of shared string table is to save memory for string storage in case your worksheet has many instances of the same string in different cells. Putting there a string that appears in a single cell only and never repeats in any other cell would not only be pointless but also unnecessary increase shared string table size which might increase the lookup speed in shared string table.

PowerKiKi:
That seems fair. If this is important to you, please consider contributing a PR for it. In the meantime we can probably workaround it by using shared strings.

eltharin:
Hi, 
It's possible to want to open files using inline strings and for the moment the return is null.

PowerKiKi:
[...]

PR #2569, opened 2022-02-09
Fixed handling of inline strings in XLSX

mweimerskirch:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [ ] Code style is respected

I was working with inline strings and noticed that the cells showed up empty in both LibreOffice and Excel.
Upon inspecting the generated XML code, I noticed that it looked like this:
```xml
<c r="A1" t="inlineStr">
    <t>Test</t>
</c>
```

instead of this:
```xml
<c r="A1" t="inlineStr">
    <is>
        <t>Test</t>
    </is>
</c>
```

To fill the cells, I used $cell->setValueExplicit('Test', DataType::TYPE_INLINE);

After applying my patch, the cells appear correctly filled in LibreOffice and Excel.

Official documentation from Microsoft: https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.inlinestring?view=openxml-2.8.1

This might solve #353. Not sure if it's exactly the same issue I was having.


oleibman:
It would be nice to include some evidence that your change works correctly. It is difficult to do this with a unit test, since the old code produces a syntactically correct spreadsheet; the missing data is evident only when the spreadsheet is opened. Instead, might you be able to add a new row using inlineStr to samples/Basic/02_Types.php? Note that this won't work initially for Xls; you need to make a one-line addition to Writer/Xls/Worksheet.php to add support for inline strings there.

mweimerskirch:
Here you go. I added an example value to the 02_Types.php file.
I also added "support" for inline strings in XLS, by handling them the exact same way as regular strings.

If you run the sample file before and after my patch, you can see that in the unpatched version, the C20 cell is empty while in the patched version it appears filled, like in the following screenshot:

image

oleibman:

Issue #2551, opened 2022-02-04
Enable Excel function implementations for use in "array formulae"


MarkBaker:
This is:

```
- [ ] a bug report
- [X] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

## Preparation of Excel function implementations allowing them to be used in "array formulae".

Typically in MS Excel, arguments passed to a function are simple "scalar" values, e.g.
```=ABS(-3)```

But MS Excel does allow for that argument passed to a function also to be a matrix/array of values:
```=ABS({-3, 2.5, -1; 0, -1, -2})```
This is still considered a single argument, but the argument is a matrix rather than a simple "scalar". In this case, MS Excel will return a matrix with the same dimensions, where the function has been applied to each individual "scalar" value in that matrix. In the above case, this will result in a response of
```{3, 2.5, 1; 0, 1, 2}```
That result may then be processed by other Excel functions in the formula (e.g. `MAX()`) to reduce it to a single value; or it may be displayed across multiple cells.

### Example of an existing function (The Math/Trig ABS() function)

This is relatively straightforward to implement for functions that only accept a single argument: the following shows the changes required to a typical function.

#### Current implementation:
```php
class Absolute
{
    /**
     * ABS.
     *
     * Returns the result of builtin function abs after validating args.
     *
     * @param mixed $number Should be numeric
     *
     * @return float|int|string Rounded number
     */
    public static function evaluate($number)
    {
        try {
            $number = Helpers::validateNumericNullBool($number);
        } catch (Exception $e) {
            return $e->getMessage();
        }

        return abs($number);
    }
}
```
#### Updated (array function enabled implementation)
```php
class Absolute
{
    private static function evaluateArray(array $numbers): array
    {
        $result = [];
        foreach ($numbers as $number) {
            $result[] = self::evaluate($number);
        }

        return $result;
    }

    /**
     * ABS.
     *
     * Returns the result of builtin function abs after validating args.
     *
     * @param mixed $number Should be numeric, or can be an array of numbers
     *
     * @return array|float|int|string rounded number
     *         If an array of numbers is passed as the argument, then the returned result will also be an array
     *            with the same dimensions
     */
    public static function evaluate($number)
    {
        if (is_array($number)) {
            return self::evaluateArray($number);
        }

        try {
[...]

PR #2618, opened 2022-02-22
Array-enable the Database Functions

MarkBaker:
This is:

```
- [ ] a bugfix
- [X] a new feature
```

Checklist:

- [X] Code style is respected

Part of the [Issue #2551](https://github.com/PHPOffice/PhpSpreadsheet/issues/2551) update to make all relevant functions array-argument-enabled

### DIFF
--- a/tests/PhpSpreadsheetTests/Calculation/Functions/Engineering/BesselITest.php
+++ b/tests/PhpSpreadsheetTests/Calculation/Functions/Engineering/BesselITest.php
@@ -20,10 +20,12 @@ class BesselITest extends TestCase
      * @dataProvider providerBESSELI
      *
      * @param mixed $expectedResult
+     * @param mixed $x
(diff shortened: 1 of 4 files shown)


tristanbailey:
Just a question, as this looks like it passes, is it just waiting until the package is ready to move to V2?

MarkBaker:
> Just a question, as this looks like it passes, is it just waiting until the package is ready to move to V2?

Most functions have already been array enabled (meaning that you can pass an array of values instead of a scalar argument, and get an array of results back), and merged into master. Database functions are an exception, because they're quite a bit more complicated.

oleibman:
No activity in over 3 years. PR does not do what its title suggests. Merge conflicts which can be resolved, but I think the resolution would be to use the existing code in master. These all suggest closing this PR.

PR #2619, opened 2022-02-22
Fix XLSX broken vertical align font style

hoogi91:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Because vertical align font style settings for XLSX documents is broken and thus `getFont()->isSuperscript()` and `getFont()->isSubscript()` will not return their expected values

The issue was introduced by commit change https://github.com/PHPOffice/PhpSpreadsheet/commit/ad5532e2f453f39f8fa041b99461935c9234d1df#diff-84c568bcc9feb3dcb6b54f6c4dc9d951782b43a3a7100f033837ed5a4e870524R122 and exists since version 1.22.0


oleibman:
Interestingly, when subscript or superscript is specified as part of a Rich Text cell, it is handled correctly (different area of code without this bug). I'll need to look over the tests to see what else is missing. No reason to delay this PR, however.

oleibman:

PR #2532, opened 2022-01-26
Add two cell anchor drawing

naotake51:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Changes are covered by unit tests
- [x] Code style is respected
- [x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [x] CHANGELOG.md contains a short summary of the change
- [ ] Documentation is updated as necessary

### Why this change is needed?

PhpSpreadsheetを使用して、セルの中央に画像を配置したいです。
twoCellAnchorタグによる画像の配置ができると助かります。

I want to use PhpSpreadsheet to center the image in a cell.
It would be helpful if there was an image arrangement by twoCellAnchor tag.

sample code
```php
// A position relative to a cell (top-left) and sized relative to another cell (bottom right)
$drawing = new Drawing();
$drawing->setName('Green Square');
$drawing->setPath('tests/data/Writer/XLSX/green_square.gif');
$drawing->setCoordinates('A1');
$drawing->setOffsetX(30);
$drawing->setOffsetY(10);
$drawing->setCoordinates2('E8');
$drawing->setOffsetX2(-50);
$drawing->setOffsetY2(-20);
$drawing->setWorksheet($sheet);
```

スクリーンショット 2022-01-27 1 25 07

output xl/drawings/drawing1.xml
```xml
<xdr:twoCellAnchor>
    <xdr:from>
        <xdr:col>0</xdr:col>
        <xdr:colOff>285750</xdr:colOff>
        <xdr:row>0</xdr:row>
        <xdr:rowOff>95250</xdr:rowOff>
    </xdr:from>
    <xdr:to>
        <xdr:col>4</xdr:col>
        <xdr:colOff>-476250</xdr:colOff>
        <xdr:row>7</xdr:row>
        <xdr:rowOff>-190500</xdr:rowOff>
    </xdr:to>
    <xdr:pic>
        ...
        <xdr:spPr>
            <a:xfrm>
                <a:off x="25400" y="812800" />
                <a:ext cx="4940300" cy="3765550" />
            </a:xfrm>
            <a:prstGeom prst="rect">
                <a:avLst />
            </a:prstGeom>
        </xdr:spPr>
    </xdr:pic>
    <xdr:clientData />
</xdr:twoCellAnchor>
```
---
[...]

Issue #126, opened 2017-03-29
Support Shapes in twoCellAnchors


ntoniazzi:
Are there plans to support [Shapes](http://www.officeopenxml.com/drwShape.php)? Currently, only [pictures](https://github.com/PHPOffice/PhpSpreadsheet/blob/develop/src/PhpSpreadsheet/Reader/Xlsx.php#L1602) seems to be supported.
Thanks.

PowerKiKi:
AFAIK there is no plan to work on that, but if you or somebody else is able to contribute a PR for that, it would be appreciated.

PR #2632, opened 2022-03-01
Two Cell Anchor Drawing

oleibman:
This change was introduced by @naotake51 as PR #2532, but it was closed before I had a chance to review it, and I have been unable to communicate with the author since. I will gladly defer to the original. Please see the original PR for details.

Similarly, PR #2237 (@AdamGaskins) is open, but with changes requested for several months. It covers a lot of the same ground as 2532. It makes sense to me to combine them.

To those changes I added support for the 'editAs' attribute - I imagine that there are instances of 'absolute' out in the wild.

There have been several other tickets referencing two cell anchors, including issue #1159 and PR #1160 (@sgarwood, who also added support for editAs), PR #643, and issue #126, all now closed but not necessarily entirely resolved. I will try to ensure that those tickets are addressed with this one.

And, in trying to make sure 1160 is covered, I stumbled upon a bug. If you use the same image resource to create two+ memory drawings, the MemoryDrawing destructor for the first will cause the rest to generate a very long warning message. This is not a problem for Php8+, only for Php7-. I have suppressed the message in the MemoryDrawing constructor. 1160 went stale due to an unresolved test error, but I don't think this was the problem. At any rate, its test works now.

This is:

```
- [x] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -4506,26 +4506,6 @@ parameters:
 			path: src/PhpSpreadsheet/Style/NumberFormat/PercentageFormatter.php
 
 		-
-			message: "#^Cannot call method getCell\\(\\) on PhpOffice\\\\PhpSpreadsheet\\\\Worksheet\\\\Worksheet\\|null\\.$#"
-			count: 1
-			path: src/PhpSpreadsheet/Worksheet/BaseDrawing.php
-
-		-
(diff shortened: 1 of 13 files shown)


oleibman:
Interesting. 20_Read_Xls and 30_Template both fail in PHP8.1. This is new, but ... they both create corrupt Xlsx files in the current release. I need to figure out those problems.

oleibman:
Some functions that need to be changed have moved to different locations. This leads to merge conflicts which are too difficult to resolve. I will rebase and try again later.

PR #2642, opened 2022-03-03
Fix font index problem

redforks:
Commit fea0e34e90d2d8 remove the follow comment:

```
  // this has to do with that index 4 is omitted in all BIFF versions for some strange reason
  // check the OpenOffice documentation of the FONT record
```

It seems both problem exist:

1. index 4 may omitted
2. some excel tools generate invalid font-index, caused fea0e34e90d2d8 problem.

Without this fix, may cause read incorrect font.


oleibman:
You should be able to avoid the Phpstan problem by coding after your `getSheetByName` call:
```php
self::assertNotNull($sheet);
```

oleibman:

PR #2643, opened 2022-03-03
docs: fix typo on formula wizard example

rmobis:
This is:

```
- [ ] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Just a simple typo fix on the documentation for the expression wizard. I was really confused until I realized it was a typo because on the image below you can see the formula.

### DIFF
--- a/docs/topics/conditional-formatting.md
+++ b/docs/topics/conditional-formatting.md
@@ -653,11 +653,11 @@ $wizardFactory = new Wizard($cellRange);
 /** @var Wizard\Expression $expressionWizard */
 $expressionWizard = $wizardFactory->newRule(Wizard::EXPRESSION);
 
-$expressionWizard->expression('ISODD(A1)')
+$expressionWizard->expression('ISODD(A2)')
     ->setStyle($greenStyle);
 $conditionalStyles[] = $expressionWizard->getConditional();
 
-$expressionWizard->expression('ISEVEN(A1)')
+$expressionWizard->expression('ISEVEN(A2)')
     ->setStyle($yellowStyle);
 $conditionalStyles[] = $expressionWizard->getConditional();
 


docs/topics/conditional-formatting.md
```diff
@@ -653,11 +653,11 @@ $wizardFactory = new Wizard($cellRange);
 /** @var Wizard\Expression $expressionWizard */
 $expressionWizard = $wizardFactory->newRule(Wizard::EXPRESSION);
 
-$expressionWizard->expression('ISODD(A1)')
+$expressionWizard->expression('ISODD(A2)')
```
MarkBaker, line 656: In Conditional Formulae, we always use cell A1 to reference the top-left cell in the formula range, whatever that range might be.
> Because we use cell `A1` as the baseline cell for relative references, the Wizard is able to handle the necessary adjustments for cell references and formulae to match the range of cells that it is being applied to when `getConditional()` is called, so it returns the correct expression.

So I'm afraid this is actually changing the formula to reference a different cell... basically saying if the cell in the row below is even or odd

MarkBaker (commented):
Thanks for the PR: however, this isn't actually a typo... several times in this document, there's mention that `A1` is used to reference the top-left cell in the conditional range, whatever cell reference that might be in the actual worksheet

> Notice also that we treat the first cell in our range as cell `A1`: the relative row number will be adjusted automatically to match our defined range; that is, the range that we specified when we instantiated the Wizard (passed in through the Wizard Factory) when we make the call to `getConditional()`.

rmobis:
Oh, sorry about that then. Guess I missed that last bit from the docs.

**PS:** How do I reference a cell to the left/up of my starting range, then?

MarkBaker:
> PS: How do I reference a cell to the left/up of my starting range, then?
Depends whether you need it pinned or not. If it's pinned, the you just use the `$` to pin row/column as needed
If it's unpinned, then you'll need to create the rule manually. The Wizard simplifies things for most cases; and allows a new range to be set without changing the formula so that the same wizard and rule can be used multiple times for different worksheet ranges (which would become a lot more complicated without the A1 top-left reference)

MarkBaker:
Don't let this put you off submitting other PRs though, either to PhpSpreadsheet, or to other projects

PR #2654, opened 2022-03-06
Handle Booleans in Conditional Styles

oleibman:
You can set up a conditional style to, say, apply to cells equal to boolean values. For such conditions, the Excel XML specifies `TRUE` or `FALSE`. It is noteworthy that false matches empty cells as well as FALSE, but not 0; similarly TRUE does not match 1. The Xlsx Writer just casts these values to string, which will not work properly. The Xlsx Reader treats the values as strings, so it won't work properly either. This PR corrects both. Also the doc blocks in Style/Conditional allow bool in some places, but not in others; these are corrected but no executable code is changed there.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Reader/Xlsx/ConditionalStyles.php
```diff
@@ -196,12 +196,19 @@ private function readStyleRules($cfRules, $extLst)
                 $objConditional->setStopIfTrue(true);
             }
 
-            if (count($cfRule->formula) > 1) {
-                foreach ($cfRule->formula as $formula) {
-                    $objConditional->addCondition((string) $formula);
+            if (count($cfRule->formula) >= 1) {
+                foreach ($cfRule->formula as $formulax) {
+                    $formula = (string) $formulax;
+                    if ($formula === 'TRUE') {
```
MarkBaker, line 202: I'm longing for the day when we've reached a minimum level of PHP that we can do:
```php
$objConditional->addCondition(
    match((string) $formulax) {
        'TRUE' => true,
        'FALSE' => false,
        default => $formula
    }
);
```
So much easier to read than `if`/`elseif`/`else`.

Does this need to make allowance for locale variants in TRUE/FALSE strings (VRAI/FAUX or VERO/FALSE)?

Issue #2372, opened 2021-11-05
Writer/BaseWriter.php: failed to open stream: Permission denied


magicua:
Hello.

This is a bug report

I am getting the following error:           
[2021-11-05 15:18:37] production.ERROR: fopen(/var/www/applications/.../storage/framework/laravel-excel/laravel-excel-scZOjOuy25S7MOmMyXnRYzntkoCoLJxZ.xlsx): failed to open stream: Permission denied {"exception":"[object] (ErrorException(code: 0): fopen(/var/www/applications/.../storage/framework/laravel-excel/laravel-excel-scZOjOuy25S7MOmMyXnRYzntkoCoLJxZ.xlsx): failed to open stream: Permission denied at /var/www/applications/.../vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Writer/BaseWriter.php:123)
[stacktrace]
#0 [internal function]: Illuminate\\Foundation\\Bootstrap\\HandleExceptions->handleError(2, 'fopen(/var/www/...', '/var/www/applic...', 123, Array)
#1 /var/www/applications/.../vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Writer/BaseWriter.php(123): fopen('/var/www/applic...', 'wb+')

but the rights to the file are correct:
-rw-r--r-- 1 www-data www-data 0 Nov  5 13:15 /var/www/applications/.../storage/framework/laravel-excel/laravel-excel-scZOjOuy25S7MOmMyXnRYzntkoCoLJxZ.xlsx

probably the problem is in the file phpoffice/phpspreadsheet/src/PhpSpreadsheet/Writer/BaseWriter.php in the line 118
`$mode = 'wb+';`

I think it is necessary to write:
`$mode = 'w+b';`

because the documentation states:
https://www.php.net/manual/en/function.fopen.php
To use these flags, specify either 'b' or 't' as the **last** character of the mode parameter.

scarneros:
Facing the same issue on Ubuntu 20.04 server, this solution worked for me:

https://unix.stackexchange.com/questions/503111/group-permissions-for-root-not-working-in-tmp

oleibman:
PR 2657, which I hope will resolve this problem, was just merged. If you have a way to test against master, please do so.

oleibman:
No update in 4 months. I believe it is fixed. Closing.

pmediavictor:
Any updates here? I am having same issue.

[...]

PR #2657, opened 2022-03-07
Remove + From Mode in BaseWriter

oleibman:
BaseWriter openFileHandle opens its output file with mode `wb+`. We don't appear to attempt any reads on the file after it is opened in write mode, so `wb` should be sufficient. This may have been a factor in the need for PR #2326, and is likely to be responsible for issue #2372.

This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected


oleibman:
@MarkBaker @PowerKiKi - I have put this in draft form because I want another opinion. This is a really simple change, it passed all the unit tests, and I can't find any reason for the unusual mode declaration in the first place. But perhaps there's some history behind it, or my search was flawed. Do you know of any reason not to move this forward?

MarkBaker:
My recollection was that it was necessary when running PHP on Windows in the early PHP5 releases, otherwise binary content such as embedded images couldn't load

oleibman:
That's good - I do almost all my testing on Windows, so, if that was still a problem, I'd see it. I'll double-check all the sample outputs before marking this ready.

PowerKiKi:
I am not aware of problems that this change could cause... :shrug:

oleibman:
I see no problems created in the sample files, so have marked this ticket as ready.

Issue #1816, opened 2021-01-31
format as table


oscaroxy:
Hi, do exist a method in order to format as table the excel?
not only filter in header, but a true table, that is in microsoft excel "home->format as table", transform the rows in table, including the tiger color rows.
thanks

MarkBaker:
c.f. [PR #2671](https://github.com/PHPOffice/PhpSpreadsheet/pull/2671) merged for release in version 1.23.0

PR #2671, opened 2022-03-12
Initial implementation of Excel's tables feature

aswinkumar863:
This is:

```
- [ ] a bugfix
- [x] a new feature
```

Checklist:

- [x] Code style is respected

> Excel's tables feature (i.e. Select `Home > Format as Table` in Excel App).

Tables are similar to AutoFilter but tables have several other advantages like named ranges, easy formatting, totals row and header row with filter. Tables can also be converted to charts and pivot tables easily.

Unlike AutoFilter, Tables are not restricted to only one per worksheet.

 **Issue:** https://github.com/PHPOffice/PhpSpreadsheet/issues/1816, https://github.com/PHPOffice/PhpSpreadsheet/discussions/2337

#### Usage:
```php
$table = new Table();
$table->setName('Sales_Data');
$table->setRange('A1:D17');
$spreadsheet->getActiveSheet()->addTable($table);
```

#### In this Commit:
- Added Table API with initial support for header and totals row.
- Added complete styling options for Table.
- Added Xlsx Writer for Table.
- Added samples.
- Covered with unit tests.

#### To be done:
- Filter expressions similar to AutoFilter.
- Precalucate formulas for totals row (Check sample 2).
- Table named ranges in formulas and calculation.


MarkBaker:
Sorry that it's taken me a while to look at this.

Looking at the PR, I have a couple of questions:
 - Should the table be protected to prevent adding new rows/columns or deleting columns/rows within the table range? If not, then it probably needs a change to the ReferenceHelper to adjust the table range
 - What styling is applied by default to the table? Is it only the pre-defined styles, or can this be changed by developers? Does it need any new styles writing to the Xlsx? 
 - Does the table need an entry in the Defined Names section?

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,407 @@
...
+    }
+
+    /**
+     * Set Table name.
+     *
+     * @return $this
+     */
+    public function setName(string $name)
+    {
+        $this->name = preg_replace('/\s+/', '_', trim($name)) ?? '';
```
MarkBaker, line 91: There's some rules for table names such as max length 255 characters, must begin with a letter, underscore or backslash, can't be the single letter `c` or `r`, can only contain letters, numbers, periods, and underscore, mustn't look like a cell reference, etc.
Should this be validated here?

MarkBaker:
And I already know that structured references are going to be a real pain in the Calculation Engine

src/PhpSpreadsheet/Worksheet/Table/Column.php
```diff
@@ -0,0 +1,202 @@
+<?php
+
+namespace PhpOffice\PhpSpreadsheet\Worksheet\Table;
+
+use PhpOffice\PhpSpreadsheet\Worksheet\Table;
+
+class Column
+{
+    /**
```
MarkBaker, line 9: Do you have any plan to add column style id and calculated columns?
e.g.
```xml
<tableColumn name="Unit Price" xr3:uid="{CF759E84-F4C3-4E91-8817-327D2F6A5BD9}" id="4" dataDxfId="1"/>
```
With the `dataDxfId` here pointing to a money style mask

and 

```xml
<tableColumn name="Value" xr3:uid="{A067D829-D65F-4AA3-996D-6F464FDADAD0}" id="5" dataDxfId="0">
    <calculatedColumnFormula>Table2[[#This Row],[Quantity]]*Table2[[#This Row],[Unit Price]]</calculatedColumnFormula>
</tableColumn>
```
Which allow the formula and style to be applied automatically for any new rows added to the table range

src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -2007,6 +2017,56 @@ public function removeAutoFilter(): self
...
+    {
+        $cellRange = Coordinate::stringFromColumnIndex($columnIndex1) . $row1 . ':' . Coordinate::stringFromColumnIndex($columnIndex2) . $row2;
+
+        return $this->addTable(new Table($cellRange, $this));
+    }
+
+    /**
+     * Remove collection of Tables.
+     */
+    public function removeTableCollection(): self
```
MarkBaker, line 2063: Is it worth having options to remove individual tables (e.g. `removeTableByCellRange()` or `removeTableByName()`) or is it better to just get the table collection, remove the table that you want removing, and then `setTableCollection()` again?

MarkBaker (commented):
I've made a few comments, just to ensure that you're considering those factors

aswinkumar863:
> Sorry that it's taken me a while to look at this.
> 
> Looking at the PR, I have a couple of questions:
> 
> * Should the table be protected to prevent adding new rows/columns or deleting columns/rows within the table range? If not, then it probably needs a change to the ReferenceHelper to adjust the table range
> * What styling is applied by default to the table? Is it only the pre-defined styles, or can this be changed by developers? Does it need any new styles writing to the Xlsx? (Custom styling created a lot of problems for me the last time I tried an implementation of pivot tables). e.g. adding a number format mask to your totals row, or for monetary value columns.
> * Does the table need an entry in the Defined Names section? (e.g. to handle your formula in the totals example). There's no defined name entry in the XML for the file, but it does appear in the Name Manager in MS Excel itself

- No, the table is not protected. Let me check the ReferenceHelper.
- Excel's pre-defined `TableStyleMedium2` style is applied by default. Only the predefined styles can be used by developers. I have to plan about custom column formatting via `dataDxfId`. [Custom table style](https://support.microsoft.com/en-us/office/format-an-excel-table-6789619f-c889-495c-99c2-2f971c0e2370#:~:text=Create%20or%20delete%20a%20custom%20table%20style)  is not yet planned (More priority to Filter expressions and Precalculation).
- No, the table doesn't need an entry in the Defined Names. But the calculation engine must consider the table and its column as defined names to perform precalculation.

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Yes. This should be validated here. I missed it. WIP

src/PhpSpreadsheet/Worksheet/Table/Column.php
aswinkumar863: Yes. I have a plan to add both column style id and calculated columns.

src/PhpSpreadsheet/Worksheet/Worksheet.php
aswinkumar863: Definitely, its worthy. I prefer `removeTableByName()` in Worksheet and `removeTablefromWorksheet()` in Table.

src/PhpSpreadsheet/Worksheet/Table.php
MarkBaker: The other aspect to watch out for here is case-insensitivity: as far as Excel is concerned, a name of `SalesData` is the same as `SALESdata`. I don't have a full list of the rules to hand; but MS must have documented it somewhere.

MarkBaker:
> * No, the table is not protected. Let me check the ReferenceHelper.

The ReferenceHelper should have the example of AutoFilters to provide you with an example; because deleting rows in a Table area is straightforward (especially as column style and calculated columns are for a future stage), but if columns are deleted you have to check remove those entries and adjust offsets. Hopefully it won't be too difficult.

MarkBaker:
> * No, the table doesn't need an entry in the Defined Names. But the calculation engine must consider the table and its column as defined names to perform precalculation.

If you can reach the point where formulae can be applied and written; I'll try to work out how to process structured references in the Calculation engine... until then, I think it's easiest if the `save()` method disables preCalculation if there are Tables on a worksheet, rather than relying on developers remembering that they need to do so.

MarkBaker:
Thank you for the PR and the work that you're doing.

This is probably the single most requested new feature for PhpSpreadsheet; and it's one that I've attempted (and failed) To implement twice in the past. Custom column formatting (if the style wasn't used elsewhere in the worksheet) was my biggest problem, because the Writer wasn't writing it... hopefully, if you're aware of that pitfall, you'll be able to avoid it.

src/PhpSpreadsheet/Worksheet/Table/Column.php
aswinkumar863: Tried to add column style but I failed. Because I couldn't find a way to get the id (zero-based index) of the style in `dxfs` of `styles.xml`.

ColumnFormula is added successfully. Note that only the new rows make use of the column formula, for existing rows the formula needs to be added to each cell in the column.

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Necessary validations added 3c3d949

src/PhpSpreadsheet/Worksheet/Worksheet.php
aswinkumar863: `removeTableByName()` added 50b91e8

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,471 @@
...
+            throw new PhpSpreadsheetException('The table name cannot be longer than 255 characters');
+        }
+        // Check for A1 or R1C1 cell reference notation
+        if (
+            preg_match(Coordinate::A1_COORDINATE_REGEX, $name) ||
+            preg_match('/^R\[?\-?[0-9]*\]?C\[?\-?[0-9]*\]?$/i', $name)
+        ) {
+            throw new PhpSpreadsheetException('The table name can\'t be the same as a cell reference');
+        }
+        if (!preg_match('/^[A-Z_\\\\]/i', $name)) {
```
MarkBaker, line 106: Note that "letters" includes the full range of UTF-8 letters, so `Γειά_σου_Κόσμε` would be a valid table name; so `\p{L}` would be better than `A-Z` in this and the next regexp (and using `/iu` for the flags, although `\p{L}`  should work even without the `u` flag)

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    /**
+     * Set Table's Worksheet.
+     *
+     * @return $this
+     */
+    public function setWorksheet(?Worksheet $worksheet = null)
+    {
+        if ($this->name != '' && $worksheet != null) {
```
MarkBaker, line 241: Strict comparisons would be better here than loose comparisons

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,471 @@
...
+            $spreadsheet = $worksheet->getParent();
+
+            foreach ($spreadsheet->getWorksheetIterator() as $sheet) {
+                foreach ($sheet->getTableCollection() as $table) {
+                    if ($table->getName() == $this->name) {
```
MarkBaker, line 246: strict comparison again

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,471 @@
...
+
+        return $this;
+    }
+
+    /**
+     * Get show Header Row.
+     *
+     * @return bool
+     */
+    public function getShowHeaderRow()
```
MarkBaker, line 123: Where possible, we're trying to improve the codebase by using return typehints rather than an `@return` entr in the docblock; though this has only been possible since PHP started supporting return typehints, so a lot of the older code still hasn't been updated

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,471 @@
...
+    private $style;
+
+    /**
+     * Create a new Table.
+     *
+     * @param string $range (e.g. A1:D4)
+     *
+     * @return $this
+     */
+    public function __construct(string $range = '', ?Worksheet $worksheet = null)
```
MarkBaker, line 67: Should the constructor also allow the table name as an argument? As it's a mandatory value for a table?

src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -2007,6 +2017,74 @@ public function removeAutoFilter(): self
...
+    public function addTable(Table $table): self
+    {
+        $table->setWorksheet($this);
+        $this->tableCollection[] = $table;
+
+        return $this;
+    }
+
+    /**
+     * Add Table Range by using numeric cell coordinates.
```
MarkBaker, line 2044: Note that I'm trying to deprecate the "ByColumnAndRow" methods in Worksheet

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    /**
+     * Validate that the specified column is in the Table range.
+     *
+     * @param string $column Column name (e.g. A)
+     *
+     * @return int The column offset within the table range
+     */
+    public function testColumnInRange($column)
```
MarkBaker, line 275: Should this be a public method, accessible from userland code?

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,471 @@
...
+    }
+
+    /**
+     * Set Table Cell Range.
+     *
+     * @return $this
+     */
+    public function setRange(string $range): self
+    {
+        // extract coordinate
```
MarkBaker, line 179: Should there be an check in this method for the minimum size of the range, e.g. if a user supplies (`A1:A1`)?

MarkBaker (changes requested):
I've made a few more comments here. Some of these issues might seem to be nit-picking; but this is very close to being ready now

src/PhpSpreadsheet/Worksheet/Table.php
```diff
@@ -0,0 +1,471 @@
...
+        if (
+            preg_match(Coordinate::A1_COORDINATE_REGEX, $name) ||
+            preg_match('/^R\[?\-?[0-9]*\]?C\[?\-?[0-9]*\]?$/i', $name)
+        ) {
+            throw new PhpSpreadsheetException('The table name can\'t be the same as a cell reference');
+        }
+        if (!preg_match('/^[A-Z_\\\\]/i', $name)) {
+            throw new PhpSpreadsheetException('The table name must begin a name with a letter, an underscore character (_), or a backslash (\)');
+        }
+        if (!preg_match('/^[A-Z_\\\\][A-Z0-9\._]+$/i', $name)) {
```
MarkBaker, line 109: A literal `.` is also a valid character in a table name

MarkBaker, line 109: A literal `.` is also a valid character in a table name

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Table name as a constructor argument is a nice addition. Also, the second argument `$worksheet` _(grabbed from autofilter)_ may get confused with `$worksheet->addTable()`. So I would try to replace it with optional `tableName`.

And the table name is not mandatory. Like the excel application, the default name will be used while writing the Xlsx file.

```php
public function __construct(string $range = '', string $name = '')
```

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: To support languages with combined characters. I used `\p{M}` along with `\p{L}`. 8889ecf0446f2e996de55a5247a32d58462ea1fd

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Changed wherever possible 3a6ebc0ce6de43c6f37c6331177767e12245b4e5

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Changed wherever possible  44d63f027aedc865342e1f87e30234c91a5df359

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Implemented 530e6642bf26522176504fccc81385d7358936d2

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Refactored 3c4a51acb5b50a2d04817db03d063d34b8a5302f

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Validated ea3263650bb10d8646df91a847ac536376de6e46

src/PhpSpreadsheet/Worksheet/Table.php
aswinkumar863: Yes, it is already there in the character class ```[A-Z0-9\._]```. I have included a name with a dot in the valid table name test case.

src/PhpSpreadsheet/Worksheet/Worksheet.php
aswinkumar863: Should I need to remove the method? Or replace it with `addTableByCellAddress`?

aswinkumar863: My thought is to remove the method completely. And it would be better to add support for `string|CellAddress` while instantiating the table.

MarkBaker: That would be the best solution I think. We've never been consistent in our use of the "ByColumnAndRow" methods, so I'm trying to make things consistent for v2.0 by removing those methods, and deprecating those existing uses now. Using string|array<int>|CellRange allows a lot more flexibility as well; if you just copy the way the AutoFilter has been changed, it should work well.

MarkBaker (changes requested):
Just the one more change that I've noted;

You're well ahead of my work on Structured References in the Calculation Engine: I've started that, and have made some progress; but am having "fun" with the  capturing regexp

src/PhpSpreadsheet/Worksheet/Worksheet.php
aswinkumar863:
``` php
$cellRange = Functions::trimSheetFromCellReference($this->validateCellRange($autoFilterOrRange));
```

I found this in the `setAutoFilter` method of `Worksheet.php`. And I need your help to use it in Table class. I have a few questions. 
- Is there any helper function similar to `validateCellRange` (outside worksheet.php)?
- Why `validateCellRange` visibility is protected and is not static (like `extractSheetTitle`) ?

MarkBaker: There's nothing similar to `validateCellRange()` outside of the Worksheet object because it was introduced as validation for methods that existed only in the Worksheet class and not elsewhere.
It's not static, because the Worksheet object must have been instantiated already to use the non-static methods that reference it; and also (potentially in the future) so that it can access the Worksheet title for additional validation.
It's protected rather than public because there was no need to call it from outside of the Worksheet class; and any public method becomes accessible to user code, which isn't necessarily a good thing.

It isn't yet in any released code because it's only been created since the last release; so if you want to call it from your Table class, I can extract it into a separate `Worksheet/Validations` class; and I can make those methods static and public. Just let me know if you want me to do that.

aswinkumar863: Thank you for clarifying that. 

Please extract it into a separate `Worksheet/Validations` class for me. Thank you once again.

MarkBaker: I've extracted those validation methods into a separate class, and made them all public static. I've also applied those validations to the cell range checking in AutoFilter (though only minimally, that code could be cleaned up a bit more)... that should also help provide an example for your own usage.

aswinkumar863: Implemented in 534cbc04c040079c26c8855670ba8ed94566c176

And I have noticed that the `$worksheet` is nullable, but the default argument value `null` is missing in the [CellAddress constructor](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Cell/CellAddress.php#L35).

```diff
-public function __construct(string $cellAddress, ?Worksheet $worksheet)
+public function __construct(string $cellAddress, ?Worksheet $worksheet = null)
```

aswinkumar863:
> Names are case insensitive; so you should probably force both to upper or lower case before doing a strict comparison; remembering that they are UTF-8 strings, so using `strToUpper()` or `strToLower()` from the Shared StringHelper, which is UTF-8 aware

Implemented in d414f139f144bd3f387c7c5f223a8d024913c6ff

src/PhpSpreadsheet/Worksheet/Worksheet.php
MarkBaker: Thanks, I'll fix that

MarkBaker (approved):
Thanks for this contribution; it's much appreciated. I'll add the details to the Change Log.

m-thalmann-athesia:
Hi all! I have a question regarding this feature: Is this documented anywhere? I can't seem to find this in the documentation...

MarkBaker:
@m-thalmann-athesia 
This is documented only with samples, and was part of the `1.23.0` release in April this year.

It's a complex feature, and being implemented in stages, with creation/writing first, and Reading basic tables is the second part included in the next release. Calculated table values won't be fully supported until I can get the Calculation Engine handling Structured References, which has been driving me insane for the past 9 months. Until then, calculated values can't be calculated after reading; and it's necessary to disable calculation when writing (although the formulae can still be written).

Full documentation is likely to be created only after the calculation engine works with Structured References; though I might add some basic documentation over Christmas/New Year.

m-thalmann-athesia:
@MarkBaker oh I see! Okay, thank you very much for the hard & great work!

bwl21:
@MarkBaker with this code fragment

```php
        // Create Table Style
        $tableStyle = new TableStyle();
        $tableStyle->setTheme(TableStyle::TABLE_STYLE_MEDIUM23);
        $table->setStyle($tableStyle);
        $spreadsheet->getActiveSheet()->addTable($table);

```

I get 

image

with 

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
<recoveryLog xmlns="http://schemas.openxmlformats.org/spreadsheetml/2006/main">
<logFileName>Reparaturergebnis speichern in persons (27)0.xml</logFileName>
<summary>Fehler in Datei '/Users/beweiche/Downloads/persons (27).xlsx'</summary>
<repairedRecords summary="Die folgenden Reparaturen wurden durchgeführt:_x000d__x000a__x000d__x000a_">
<repairedRecord>Reparierte Datensätze: Tabelle von /xl/tables/table1.xml-Part (Tabelle)</repairedRecord>
</repairedRecords>
</recoveryLog>
```

What can I do?

oleibman:
@bwl21 Samples/Table/01_Table.php uses exactly the code you asked about, and does not create a corrupt file. (It does a little additional styling, but, even when I comment out those additional statements, there is no problem with the output.)  In order to see what's going wrong for you, I need more information - the script or the output file or both.

bwl21:
> @bwl21 Samples/Table/01_Table.php uses exactly the code you asked about, and does not create a corrupt file. (It does a little additional styling, but, even when I comment out those additional statements, there is no problem with the output.)  In order to see what's going wrong for you, I need more information - the script or the output file or both.

thanks for your investigatin. i will try to make a test case.

bwl21:
@oleibman I managed to find the reason:

1. the content of a value must not bei `null`
2. I observe that the headers must be unique

can you confirm this?

oleibman:
Confirmed. Although my following statement may not be precisely correct, it captures the gist. Excel will report a problem if a table has columns with the same name, or a column with no name. We might be able to do something to prevent us creating a spreadsheet with those conditions. Please open a new issue if you want us to look into that. In the meantime, you should have enough information to be able to get around your problem for now.

PR #2673, opened 2022-03-12
Resolve Conditional Formatting issues with Xls writer

MarkBaker:
This is:

```
- [X] a bugfix
- [ ] a new feature
```

Checklist:

- [X] Code style is respected

This PR resolves the problem with writing a CF Header for each CF range, no matter how many rules there are for that range, rather than writing a single header for all ranges.

It does not appear that this has ever worked correctly, and that only writing a single CF range was supported before, but recent changes to provide examples of multiple CF ranges in a single worksheet have highlighted the problem, while a modification to the Xls Writer to support multiple ranges actually broke the CF Writer so that Excel issues a warning when the file is opened.

Now the code can also handle:
 - Using cell references for the rule value
 - Using string literals for the rule value
 - Using formula expressions for the rule value
 
None of these previously worked. All cell reference, string literal or formula expressions used as rule values were silently converted to integer 0

In the event of s formula expressions that Xls itself doesn't support (e.g. when they use an Excel function that isn't known to the BIFF format), the code defaults the condition to `ptgInt` +0 (`=0`) to avoid breaking the structure of the file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,8 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
+- Fix bug in Conditional Formatting in the Xls Writer that resulted in a broken file when there were multiple conditional ranges in a worksheet.
+- Fix Conditional Formatting in the Xls Writer to work with rules that contain string literals, cell references and formulae.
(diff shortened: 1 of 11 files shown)


MarkBaker:
I guess I'll also need to create a series of unit tests for CF Writing both for Xlsx and Xls; create, save, reload and verify that they're as written.

PR #2683, opened 2022-03-16
Tweak to php-cs-fixer Parameters

oleibman:
For a recent change, I removed some errors from Phpstan baseline and instead added annotations in the source members. I did this work incrementally, and was surprised when php-cs-fixer required a change from `// @phpstan-ignore-next-line` to 
`/** @phpstan-ignore-next-line */`. No problem thinks I, and continue to modify several members using the new convention until php-cs-fixer required a change from `/** @phpstan-ignore-next-line */` to `// @phpstan-ignore-next-line`??? I did as directed, and continued to be surprised for the rest of that ticket.

Having had time to research, the problem is due to two options in the php-cs-fixer config file `'comment_to_phpdoc' => true` and `'phpdoc_to_comment' => true`. It seems that php-cs-fixer is treating these annotations the same as doc-blocks, expecting `/**` before a `structural element`, and `//` otherwise. For the statements where I had questions, it expects `/**` before a statement which you might be able to precede with `/** @var`, and `//` where you would not be able to precede it with `/** @var`. However, in this case, what it is doing is forcing what appear to be inconsistencies between otherwise identical statements, whereas php-cs-fixer is supposed to be supporting consistent syntax throughout the project. This PR changes both options to false, allowing (but not requiring) a consistent syntax for these examples. It contains an example of a change from each format to the other, changes which php-cs-fixer would previously have flagged.

An added bonus for this change is that Scrutinizer annotations can now be added to the code; these were often rejected by php-cs-fixer. These should, of course, be used very conservatively, but there are cases where Scrutinizer's analysis is either faulty or not helpful. This PR takes advantage of the change by adding annotations to eliminate the two existing problems which Scrutinizer classifies as 'Security', problems for which there is no sensible way to satisfy Scrutinizer's complaint.

No executable code is changed by this PR.

This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] tool configuration
```

Checklist:

- [x] Code style is respected


MarkBaker:
I had similar problems recently, with php-cs-fixer changing `// @var string var` to `/** @var string var */` that I'd added for phpstan. Contrary to phpstan's documented behaviour, phpstan was happy with `// @var string var`, but reported an expects string but got string|null error when the annotation was `/** @var string var */`... so I was in a situation where either php-cs-fixer reported an error, or phpstan reported an error.

I didn't delve further into the cause; but made some changes so that I didn't need to use the phpstan annotation at all; but I suspect it may have been related to this issue that you've discovered with the conflicting php-cs-fixer configuration.

PR #2701, opened 2022-03-19
fix #2700 : set xaxis/yaxis and chart title color

### LINKED ISSUE #2700: Set color for xaxis / yaxis title
This is: 

- [x] a feature request

I would be able to choose xaxis / yaxis color like that:

image

rycks:
This PR is a partial solution (only excel export) for #2700 : now you can choose font color for xaxis/yaxis/title charts.

### DIFF
--- a/src/PhpSpreadsheet/Chart/Title.php
+++ b/src/PhpSpreadsheet/Chart/Title.php
@@ -3,6 +3,7 @@
 namespace PhpOffice\PhpSpreadsheet\Chart;
 
 use PhpOffice\PhpSpreadsheet\RichText\RichText;
+use PhpOffice\PhpSpreadsheet\Style\Font;
 
 class Title
 {
@@ -21,6 +22,12 @@ class Title
     private $layout;
 
     /**
+     * Title Font
+     * @var Font
+     */
+    private $font;
+
+    /**
      * Create a new Title.
      *
      * @param array|RichText|string $caption
@@ -29,6 +36,7 @@ class Title
     {
         $this->caption = $caption;
         $this->layout = $layout;
+        $this->font = new Font();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


MarkBaker:
Is it possible to provide some unit tests to verify this? Charts is one area that is the least covered by tests, and it really needs some TLC there.

MarkBaker:
Also, please run a phpcs check to ensure coding complies with the PSR12 standard; phpcbf or php-cs-fixer should be able to fix all these warnings

src/PhpSpreadsheet/Chart/Title.php
```diff
@@ -87,4 +95,23 @@ public function getLayout()
...
+        return $this->font;
+    }
+
+    /**
+     * Set font
+     *
+     * @param Font $font
+     * @return Title
+     */
+    public function setFont(Font $font = null)  {
```
MarkBaker, line 113: If font defaults to null, the property should also allow null; and the writer should also check for null font when reading the colour

MarkBaker (changes requested):
Please address the requested change; and `composer fix` should resolve all the warnings

oleibman:
Converting to draft status for now. There is no need for a `font` property for `Title`, which already has a property `caption`, which can be a `RichText` item, which can already have font information. PR #2828, which I will probably merge this week, uses this information to set chart title color for both export and import.

That PR does not take care of the axis labels (I believe it will handle read correctly but that's only part of the equation). I will look into that soon. Axis labels are more complicated than chart titles because write (export) attempts to reduce the caption from array/string/RichText to a string, and it looks like it does so incorrectly. So it will need further changes, in addition to handling colors.

oleibman:
Closing - superseded by #2828 and #2841, which are now both merged.

Issue #2183, opened 2021-06-23
Specified colors do not work on bar charts


wmelnick-cmeonline:
This is:

```
- [X] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Colors specified for bar chart work like they do for donut chart

### What is the current behavior?
Default style colors show

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

// add code that show the issue here...
$colors = [ '00ff00', 'cfcfcf', '007f00' ];
$dataSeriesValues1 = [
    new DataSeriesValues(DataSeriesValues::DATASERIES_TYPE_NUMBER, 'Worksheet!$Z$1:$Z$2', null, 2, [], null, $colors),
];
// Then do the rest to create the DataSeries and then the Chart.
// with DataSeries::TYPE_DONUTCHART the chart colors match what is sent.
// with DataSeries::TYPE_BARCHART the chart colors are always the style colors.
```

### Which versions of PhpSpreadsheet and PHP are affected?
php 7.2, current PhpSpreadsheet

oleibman:
Should be solved by #2906, which is part of just released 1.24.

Issue #1433, opened 2020-04-04
Custom Color for Bars & Columns


arpit0515:
This is:

### What is the expected behavior?
I am trying to custom color the bars of the bar chart & column chart which is being exported using phpoffice/phpspreadsheet. I have used multiple codes and ways which are mentioned in the pie chart example code but it is not working.

### What is the current behavior?
default color scheme is being applied to the bar charts

### What are the steps to reproduce?
create any bar chart using phpoffice/phpspreadsheet and try to change the color either in the dataserieslabels or dataseriesvalues. Nothing is happening. 

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = new \PhpOffice\PhpSpreadsheet\Spreadsheet();

// add code that show the issue here...

$dataSeriesLabels = [
    new DataSeriesValues(DataSeriesValues::DATASERIES_TYPE_STRING, 'Data!$C$1', null, 1, [], null, ["#CCCCCC"]), // 'Budget'
    new DataSeriesValues(DataSeriesValues::DATASERIES_TYPE_STRING, 'Data!$D$1', null, 1, [], null, ["#D6D6D6"]), // 'Forecast'
    new DataSeriesValues(DataSeriesValues::DATASERIES_TYPE_STRING, 'Data!$E$1', null, 1, [], null, ["#B8292F"]), // 'Actual'
    new DataSeriesValues(DataSeriesValues::DATASERIES_TYPE_STRING, 'Data!$F$1', null, 1, [], null, ["#EB8500"]), // 'Actual'
];

$dataSeriesValues = [
    new DataSeriesValues(DataSeriesValues::DATASERIES_TYPE_NUMBER, 'Data!$C$2:$C$5', null, 4, [], null, array('#cccccc', '#00abb8', '#b8292f', '#eb8500')),
    new DataSeriesValues(DataSeriesValues::DATASERIES_TYPE_NUMBER, 'Data!$D$2:$D$5', null, 4, [], null, array('#cccccc', '#00abb8', '#b8292f', '#eb8500')),
[...]

PR #2724, opened 2022-04-01
Allow setting fill colour for bar chart data points

### LINKED ISSUE #1400: Bar chart fillcolor not working
This is:

```
- [* ] a bug report
- [ ] a feature request
```

If I set the colors for a BarChar, I'm expecting the lines to have those colors.

BarChar don't have those colors

Take the sample file 33_Chart_create_bar_stacked.php, and add a color.

just change this line 
 $dataSeriesValues = array(
                    new DataSeriesValues(DataSeriesValues::DATASERIES_TYPE_NUMBER, 'Worksheet!$D$2:$D$5', NULL, 4, array(), null, '000000')) ;

 the latest

DanWin:
This is:

- [X] a bugfix
- [ ] a new feature

Checklist:

- [X] Code style is respected

This will fix #1400 #1433 and #2183

### DIFF
--- a/src/PhpSpreadsheet/Writer/Xlsx/Chart.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx/Chart.php
@@ -1082,7 +1082,7 @@ class Chart extends WriterPart
             //    Values
             $plotSeriesValues = $plotGroup->getPlotValuesByIndex($plotSeriesRef);
 
-            if (in_array($groupType, [DataSeries::TYPE_PIECHART, DataSeries::TYPE_PIECHART_3D, DataSeries::TYPE_DONUTCHART, DataSeries::TYPE_BARCHART, DataSeries::TYPE_BARCHART_3D])) {
+            if (in_array($groupType, [DataSeries::TYPE_PIECHART, DataSeries::TYPE_PIECHART_3D, DataSeries::TYPE_DONUTCHART, DataSeries::TYPE_BARCHART, DataSeries::TYPE_BARCHART_3D], true)) {
                 $fillColorValues = $plotSeriesValues->getFillColor();
                 if ($fillColorValues !== null && is_array($fillColorValues)) {
                     foreach ($plotSeriesValues->getDataValues() as $dataKey => $dataValue) {
--- a/src/PhpSpreadsheet/Writer/Xlsx/Chart.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx/Chart.php
@@ -1082,7 +1082,7 @@ class Chart extends WriterPart
             //    Values
             $plotSeriesValues = $plotGroup->getPlotValuesByIndex($plotSeriesRef);
 
-            if (($groupType == DataSeries::TYPE_PIECHART) || ($groupType == DataSeries::TYPE_PIECHART_3D) || ($groupType == DataSeries::TYPE_DONUTCHART)) {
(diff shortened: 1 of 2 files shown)


src/PhpSpreadsheet/Writer/Xlsx/Chart.php
```diff
@@ -1082,7 +1082,7 @@ private function writePlotGroup(?DataSeries $plotGroup, $groupType, XMLWriter $o
             //    Values
             $plotSeriesValues = $plotGroup->getPlotValuesByIndex($plotSeriesRef);
 
-            if (($groupType == DataSeries::TYPE_PIECHART) || ($groupType == DataSeries::TYPE_PIECHART_3D) || ($groupType == DataSeries::TYPE_DONUTCHART)) {
+            if (in_array($groupType, [DataSeries::TYPE_PIECHART, DataSeries::TYPE_PIECHART_3D, DataSeries::TYPE_DONUTCHART, DataSeries::TYPE_BARCHART, DataSeries::TYPE_BARCHART_3D])) {
```
MarkBaker, line 1085: `in_array()` should always be used in strict mode

MarkBaker:
Is it possible to provide some unit tests to verify this for all chart types that are affected?

MarkBaker (changes requested):

MarkBaker (changes requested):
Please address requested change; and unit testing

DanWin:
Agreed, `in_array()` should be in strict mode. I saw it mostly being non-strict in other parts of the code and the previous checks were also non-strict and thus kept it the same way.
As for unit testing, I'll need help with that. I've never written those and I'm not sure how to write tests for this case.

MarkBaker:
> Agreed, `in_array()` should be in strict mode. I saw it mostly being non-strict in other parts of the code and the previous checks were also non-strict and thus kept it the same way.

Thanks. A lot of the code that uses `in_array()` was originally written for earlier versions of PHP, before the strict mode flag was added. Now, we always try to apply it when we're working on those parts of the codebase where it isn't yet applied (scouting rules).

> As for unit testing, I'll need help with that. I've never written those and I'm not sure how to write tests for this case.

There's some examples on how to approach testing in `/tests/PhpSpreadsheetTests/Writer/Xlsx/*` where we use the writer calls to return the XML to the test rather than actually saving to a file; and can then assert that the XML contains the markup that's expected.
e.g. `ConditionalTest.php` and the `testWriteSimpleCellConditionalFromWizard()` method.

Though you might want to create a `Charts` subfolder under `.../Writer/Xlsx/`, and put the test in there.
 
Something like:
```php
namespace PhpOffice\PhpSpreadsheetTests\Writer\Xlsx\Charts;

use PhpOffice\PhpSpreadsheetTests\Functional\AbstractFunctional;

class BarChartTest extends AbstractFunctional
{
    public function testWriteBarChartWithUserDefinedFillColors(): void
    {

        // Code to create a bar chart with user-defined fill colours goes here

        $writerWorksheet = new Xlsx\Worksheet($writer);
        $data = $writerWorksheet->writeWorksheet($worksheet, []);

        $expected = <<<XML
//xml for a bar chart with user-defined fill colours goes here
XML;
        self::assertStringContainsString($expected, $data);
    }
}
```

oleibman:
This PR will be superseded by #2906, including tests. Converting this to draft; and will close it if 2906 is merged.

oleibman:
Superseded by 2906, which was just merged.

PR #2731, opened 2022-04-07
Fix reading of files in the root of a zip

andres1gb:
This is:

```
- [x] a bugfix
- [ ] a new feature
```

Checklist:

- [x] Code style is respected

Because the project fails to read Xlsx files created by some generators (such as Google Campaign Manager) which puts xml files in the root of the zip container.


andres1gb:
It seems that Scrutinizer fails to start the test. Looks more like an initialization issue than a failure with the test itself.

`configure: error: Package requirements (openssl >= 1.0.2) were not met:

Requested 'openssl >= 1.0.2' but version of OpenSSL is 1.0.1f

Consider adjusting the PKG_CONFIG_PATH environment variable if you
installed software in a non-standard prefix.

Alternatively, you may set the environment variables OPENSSL_CFLAGS
and OPENSSL_LIBS to avoid the need to call pkg-config.
See the pkg-config man page for more details.`

oleibman:
You are correct about the Scrutinizer problem. If it remains unfixed, it will not stop your change from being merged. However, would it be possible to add a unit test which would fail without your change, but which succeeds with your change?

andres1gb:
> You are correct about the Scrutinizer problem. If it remains unfixed, it will not stop your change from being merged. However, would it be possible to add a unit test which would fail without your change, but which succeeds with your change?

I have CM360 generated Xlsx files that would fail, so I could replace business-related data (that I can't disclose) while keeping the ZipFile structure and write a test that simply read that file. Without the change it won't be able to read any cell.

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -368,6 +368,9 @@ private function getFromZipArchive(ZipArchive $archive, $fileName = '')
         if (strpos($fileName, '//') !== false) {
             $fileName = substr($fileName, strpos($fileName, '//') + 1);
         }
+        // Relative paths generated by dirname($filename) when $filename
+        // has no path (i.e.files in root of the zip archive)
+        $fileName = (string) preg_replace('/^\.\//', '$1', $fileName);
```
oleibman, line 373: Sorry for not noticing this earlier. Your regular expression works, but only accidentally. The value in `$1` should never be set given the regular expression you're using. I think setting the replacement value to the null string is what you want to do.

oleibman (changes requested):

src/PhpSpreadsheet/Reader/Xlsx.php
andres1gb: Absolutely. I simplified the former regex (which grouped apart the path), and left the replacement that happens to be always null. It's already fixed.

oleibman (approved):

oleibman:

PR #2742, opened 2022-04-11
Decrease xlsx reader memory consumption

czoIg:
This is:

```
- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
```

Checklist:

- [x] Code style is respected

While reading big xlsx file like 400k rows [test.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/8464853/test.xlsx) `simplexml_load_string()` uses almost 6 GB of memory.
After changing it to `XMLReader` each node is reading separately and xml string is added to "map". Each map's element is parsed with `simplexml_load_string()` separately which dramatically decreases memory usage (~1,2 GB for the mentioned earlier file).

### How to check

XlsxTest.php
```php
<?php

$file = './test.xlsx';

$inputFileType = IOFactory::identify($file);
$objReader = IOFactory::createReader($inputFileType);
$excel = $objReader->load($file);

echo 'Done';
```

```
/usr/bin/time -l php XlsxTest.php
```
(cannot use `memory_get_usage()` because of https://bugs.php.net/bug.php?id=62467)

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx/SheetStructure.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx/SheetStructure.php
@@ -1,0 +1,30 @@
+<?php
+
+namespace PhpOffice\PhpSpreadsheet\Reader\Xlsx;
+
+class SheetStructure
+{
+    public const CONDITIONAL_FORMATTING = 'conditionalFormatting';
+    public const SHEET_PROTECTION = 'sheetProtection';
+    public const SHEET_VIEW = 'sheetView';
+    public const MERGE_CELL = 'mergeCell';
+    public const EXT_LST = 'extLst';
+    public const DATA_VALIDATION = 'dataValidation';
+    public const LEGACY_DRAWING_HF = 'legacyDrawingHF';
+    public const DRAWING = 'drawing';
+    public const AUTO_FILTER = 'autoFilter';
+    public const TABLE_PART = 'tablePart';
+    public const HYPERLINK = 'hyperlink';
+    public const ROW = 'row';
+    public const COL = 'col';
+    public const PROTECTED_RANGE = 'protectedRange';
+    public const SHEET_PR = 'sheetPr';
+    public const SHEET_FORMAT_PR = 'sheetFormatPr';
+    public const PRINT_OPTIONS = 'printOptions';
+    public const PAGE_MARGINS = 'pageMargins';
+    public const PAGE_SETUP = 'pageSetup';
+    public const HEADER_FOOTER = 'headerFooter';
+    public const ROW_BREAKS = 'rowBreaks';
+    public const COL_BREAKS = 'colBreaks';
+    public const ALTERNATE_CONTENT = 'AlternateContent';
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
   (i.e. `"Vrai"` wil be converted to a boolean `true` if the Locale is set to `fr`.)
 - Allow `psr/simple-cache` 2.x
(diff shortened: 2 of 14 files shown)


MarkBaker:
Sorry that it's taken me a while to get round to looking at this: I've been doing a lot of work elsewhere in the application, most recently in reducing memory usage for the cell collection once a file is loaded.

I'm about to take a deeper dive into this PR (thank you for the submission), and it certainly looks positive.

I can't even load your sample test file with my basic test setup;
```
vagrant@homestead:~/projects/PHPOffice/PHPSpreadsheet/develop$ php samples/readerTest.php
Killed
```
But using your branch, the file can actually be loaded
```
vagrant@homestead:~/projects/PHPOffice/PHPSpreadsheet/develop$ php samples/readerTest.php

Call time to load spreadsheet file was 188.0361 seconds
 Current memory usage: 1064964 KB
    Peak memory usage: 1087496 KB
```
and that's a very powerful positive.

I'll need to do some more testing with smaller files, and with a range of features, so that I can make a proper comparison of speed and memory usage; and also take a look at how you've made the changes.

MarkBaker:
Just out of interest for a comparison: if I merge the memory-saving changes that I've already merged to master into your branch, I get the following memory usage:
```
vagrant@homestead:~/projects/PHPOffice/PHPSpreadsheet/develop$ php samples/readerTest.php

Call time to load spreadsheet file was 197.1121 seconds
 Current memory usage: 1001476 KB
    Peak memory usage: 1024008 KB
```
Reduced by another 60MB

src/PhpSpreadsheet/Reader/Xlsx/SheetStructure.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+namespace PhpOffice\PhpSpreadsheet\Reader\Xlsx;
+
+class SheetStructure
+{
+    public const CONDITIONAL_FORMATTING = 'conditionalFormatting';
```
MarkBaker, line 7: Definitely a good use of constants for all the different elements

MarkBaker:
There are definitely some trade-offs here:

Some tests with smaller files, so that I can do an easy comparison between master and your code. With a 10-column, 40k row spreadsheet with purely numeric cell values:

master
```
Call time to load spreadsheet file was 30.4329 seconds
 Current memory usage: 147456 KB
    Peak memory usage: 147456 KB
```
your PR
```
Call time to load spreadsheet file was 38.3924 seconds
 Current memory usage: 186368 KB
    Peak memory usage: 188420 KB
```

With a 10-column, 102k row spreadsheet with purely numeric cell values:

master
```
Call time to load spreadsheet file was 96.7067 seconds
 Current memory usage: 333824 KB
    Peak memory usage: 333824 KB
```
your PR
```
Call time to load spreadsheet file was 95.7461 seconds
 Current memory usage: 434176 KB
    Peak memory usage: 438276 KB
```

But if I increase that to 200k rows, then master crashes out but your PR will successfully read the file
```
Call time to load spreadsheet file was 208.5629 seconds
 Current memory usage: 864256 KB
    Peak memory usage: 872452 KB
```

Your approach performs about 25% slower with smaller spreadsheets until about the million cell mark; and is consistently higher memory consumption where I'm able to measure both; yet it's able to read the largest spreadsheets where the current master branch code fails.

That gives me something of a dilemma. I'd love to be able to handle those extremely large spreadsheets as well; but not at the expense of higher memory usage and slower performance for the majority of users working with smaller spreadsheets. Your PR is great for the exception cases; but not for the majority of cases.

Having two Xlsx Readers, one for the smaller files and one for extremely large files might be an option, although it would have a lot of code duplication between the two, and increase the maintenance work; but it's also difficult to advise the circumstances when each should be used... it's not based purely on number of cells, shared strings and shared formulae make a difference.

PR #2772, opened 2022-04-23
Time Interval Formatting

### LINKED ISSUE #2768: getFormattedValue does not honor brackets in time formats
This is:

```
- [x] a bug report
- [ ] a feature request
```

A table cell formatted with e.g. "[hh]:mm" and a value higher than 1 should display more the correct hours.
A value of 1.15 should yield a formatted value of "27:36".

The formatted value is "[03]:36".

```php
<?php

$sheet->getStyle('A1')->getNumberFormat()->setFormatCode('[hh]:mm');
$sheet->setCellValue('A1', 1.15);
echo $sheet->getCell('A1')->getFormattedValue();
```

1.22.0, PHP 7.4

oleibman:
Fix #2768. DateFormatter handles only one of six special formats for time intervals `[h] [hh] [m] [mm] [s] [ss]`. This PR extends support to the rest. There should be no more than one of these in any format string. Although it certainly could make sense to treat `[d] [dd]` in the same manner, Excel does not seem to support those.

Interesting observations - hours and minutes are truncated (presumably because they may be followed by minutes and seconds), but seconds are rounded. Also, there are some floating point issues, which fortunately showed up for the example in the original issue. There, the time interval was 1.15, which should evaluate to a minutes value of 1656 (as it does in Excel). However, on my system it evaluated to 1655 because of a rounding error in the 13th decimal place. To overcome this, values are rounded to 10 decimal places before truncating.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Style/NumberFormat/DateFormatter.php
```diff
@@ -112,23 +160,23 @@ public static function format($value, string $format): string
         // escape any quoted characters so that DateTime format() will render them correctly
         /** @var callable */
         $callback = ['self', 'escapeQuotesCallback'];
-        $format = preg_replace_callback('/"(.*)"/U', $callback, $format);
```
MarkBaker, line 115: While I'm reminded: callbacks in the format `['self', 'escapeQuotesCallback']` are being deprecated in PHP 8.2; the valid replacement format is `[self::class, 'escapeQuotesCallback']`

MarkBaker: Also worth noting:
```php
$format = preg_replace_callback('/"(.*)"/U', $callback, $format);
```
can return null or a string; and phpstan may subsequently complain when that value is passed as an argument to another function. Rather than doing `anotherFunction($format ?? '')`, it's probably better to enforce the string earlier with an explicit cast of the return from `preg_replace_callback()`
```php
$format = (string) preg_replace_callback('/"(.*)"/U', $callback, $format);
```

oleibman: Changes applied in latest commit. It looks like there are only 2 other places in PhpSpreadsheet where 'self' is used.

PR #2774, opened 2022-04-24
Real Errors Identified in Calculation by Scrutinizer

### LINKED ISSUE #2780: Warning: trim() expects parameter 1 to be string, array given
This is:

```
- [X] a bug report
- [ ] a feature request
```

no Warning on trim()

trim() trigger a Warning on an array use

use the Excel file wich has a cell with this type of formula =SUM(D1:INDIRECT(ADDRESS(ROW();COLUMN()-4)))
load then save = Warning!

<?php

	$file = 'c:\tmp\relativecell.xlsx';
	$showPHPExcel = \PhpOffice\PhpSpreadsheet\IOFactory::load($file);
	$sheets_array = $showPHPExcel->getSheetNames();
	$liassewriter = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($showPHPExcel, "Xlsx");
	$file = 'c:\tmp\file1.xlsx';
	$liassewriter->save($file);
?>

- [ ] Reader
- [X ] Writer
- [ ] Styles
- [ ] Data Validations
- [x] Formula Calulations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

[relativecell.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/8555977/relativecell.xlsx)

ALL

### Patch should be 
Patch : Calculation.php line 4536 & 4546
Existing : 
if ((trim($operand1Data['value']) != '') && (is_numeric($operand1Data['value']))) {
Should be  :
if (!is_array($operand1Data['value']) && (trim($operand1Data['value']) != '') && (is_numeric($operand1Data['value']))) {

oleibman:
Before Scrutinizer broke, I took a look at the remaining 43 errors which it categorized as 'major'. Most of these were false positives, but, in the case of Calculation and Reader/Xlsx/Chart, I was able to determine that its analysis of some of the problems was correct. There is little point addressing the false positives until it starts working again, but we should fix the real errors.

This PR addresses the real errors in Calculation.
- A test for `$pCellParent` should have been a test  for `$pCellWorksheet`.
- A test for `$operand1Data['reference']` should have been a test for `$operand1Data['value']`.
- A test for `$operand2Data['reference']` should have been a test for `$operand2Data['value']`.

Fix #2780. Warning message is being issued when getting calculated value of cell with value `=INDIRECT(ADDRESS(ROW(),COLUMN()-2))/$C$4`. Problem was in the same section of code that is being changed in this PR.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Calculation/Calculation.php
```diff
@@ -3155,7 +3155,8 @@ private static function translateFormula(array $from, array $to, string $formula
                 $i = false;
                 foreach ($temp as &$value) {
                     //    Only adjust in alternating array entries
-                    if ($i = !$i) {
+                    $i = !$i;
+                    if ($i) {
```
MarkBaker, line 3159: If we are going to change the instances of this, then using the more explicit `if ($i === true) {` would probably be better; and perhaps even renaming `$i` as something more meaningful like `$notWithinQuotes`

oleibman: Done.

Issue #1445, opened 2020-04-16
[tip] How to migrate from PHPExcel to PHPSpreadsheet 


TomasVotruba:
Hi!

**I want to share a tip for people stuck on [deprecated PHPExcel](https://github.com/PHPOffice/PHPExcel)**. It might take < 30 minutes to migrate to PHPSpreadsheet now.

I maintain @rectorphp open-source project, that handles instant migrations with AST (known in nikic/php-parser and PHPStan tools)

Recently I got into project that uses PHPExcel, read the [migration README](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/docs/topics/migration-from-PHPExcel.md) and made a migration set.

I wrote post about it: [How to Migrate From PHPExcel to PHPSpreadsheet with Rector in 30 minutes](https://getrector.org/blog/2020/04/16/how-to-migrate-from-phpexcel-to-phpspreadsheet-with-rector-in-30-minutes)

---

**tl;dr;**

```bash
composer require rector/rector --dev
vendor/bin/rector process src --set phpexcel-to-phpspreadsheet
```

And you're ready to go :wink:

PowerKiKi:
This is great thank you for sharing. I would like to integrate this info official documentation. Do you have any opinion on how to do that? Replace everything with Rector? Or replace only our regexp tool and keep description of other changes?

TomasVotruba:
> Replace everything with Rector?

:+1: That would the best. I tried to cover all the changes described in the migration guide.

If you find any mistake or would like to add more changes to automate, let me know. It should be doable :wink:

PowerKiKi:
@TomasVotruba could you please edit your blog post and replace all references from:

`https://github.com/PHPOffice/PhpSpreadsheet/blob/master/docs/topics/migration-from-PHPExcel.md` 

to:

`https://github.com/PHPOffice/PhpSpreadsheet/blob/50d78ce7898ee3a540cefd9693085b3636e578e6/docs/topics/migration-from-PHPExcel.md`

So the old manual way is still somewhat accessible.

TomasVotruba:
Sure, that's a good idea. There you go: https://github.com/rectorphp/getrector.org/commit/0596ce4bf406c09fc24f787f588d03cfd138c67f

If you share a tweet about this, just mention [@rectorphp](https://twitter.com/rectorphp) and I'll retweet it.

jeanmobayed:
[...]

PR #2861, opened 2022-05-31
Update excel migration with sets() method

TomasVotruba:
Hi, 
I'm here to update old information about the migration with Rector :)
The `--set` is no longer used. We have people comming to Rector to report it, as the docs here is still old:
https://github.com/PHPOffice/PhpSpreadsheet/issues/1445

This PR solves it and shows the new way to use Rector configuration :+1:


TomasVotruba:
Thank you :clap:

PR #2870, opened 2022-06-03
Fix: SUBTOTAL to handle #REF! errors

ndench:
This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

When Excel calculates a `SUBTOTAL` that contains `#REF!`, such as `=SUBTOTAL(9, #REF!)` it returns `#REF!`. PhpSpreadsheet causes a PHP warning to be raised due to an array unpacking in [Subtotal](https://github.com/hyraiq/PhpSpreadsheet/blob/d0ff3d9ca7fc1b44cc2136ec427ca219572215c6/src/PhpSpreadsheet/Calculation/MathTrig/Subtotal.php#L38).

The root cause is the call to [Functions::flattenArrayIndexed](https://github.com/hyraiq/PhpSpreadsheet/blob/d0ff3d9ca7fc1b44cc2136ec427ca219572215c6/src/PhpSpreadsheet/Calculation/MathTrig/Subtotal.php#L90) which returns something that looks like `[0 => '#REF!']` as opposed to `[0.W.29 => '#REF!']`. The `Subtotal.php` function then does an explode on the key to get the column and row values to check if they are hidden or contain other aggregate functions that the `SUBTOTAL` should ignore.

I'm unsure of the best way to handle this. 

1. Should the `FlattenArrayIndexed` be updated? 
    - I've steered clear of this because it's used in many places and I was unsure of flow on effects
2. Are there other `#REF!` calculation errors that should also be tracked down? 
    - A quick search shows that not many places use `explode` and then unpack the result immediately into more than property. So I think it's unlikely.
3. Can `Subtotal.php` handle other error type values?
    - There are already tests for the `#VALUE!` error. Excel won't let me write other errors into a column to see how it handles them with the SUBTOTAL function.

Submitting this PR to provide some failing test cases and my suggested solution (which has the fewest flow on effects possible).
Let me know if there's another way this should be handled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
+- #REF! value handling in the Excel Math/Trig SUBTOTAL() function [PR #2870](https://github.com/PHPOffice/PhpSpreadsheet/pull/2870)
 - Xls Reader resolving absolute named ranges to relative ranges [Issue #2826](https://github.com/PHPOffice/PhpSpreadsheet/issues/2826) [PR #2827](https://github.com/PHPOffice/PhpSpreadsheet/pull/2827)
 - Null value handling in the Excel Math/Trig PRODUCT() function [Issue #2833](https://github.com/PHPOffice/PhpSpreadsheet/issues/2833) [PR #2834](https://github.com/PHPOffice/PhpSpreadsheet/pull/2834)
 - Invalid Print Area defined in Xlsx corrupts internal storage of print area [Issue #2848](https://github.com/PHPOffice/PhpSpreadsheet/issues/2848) [PR #2849](https://github.com/PHPOffice/PhpSpreadsheet/pull/2849)
--- a/src/PhpSpreadsheet/Calculation/MathTrig/Subtotal.php
+++ b/src/PhpSpreadsheet/Calculation/MathTrig/Subtotal.php
@@ -89,6 +89,10 @@ class Subtotal
         $cellReference = array_pop($args);
         $aArgs = Functions::flattenArrayIndexed($args);
 
+        if (in_array(ExcelError::REF(), $args)) {
+            return ExcelError::REF();
+        }
+
         try {
             $subtotal = (int) Helpers::validateNumericNullBool($functionType);
         } catch (Exception $e) {
--- a/tests/PhpSpreadsheetTests/Calculation/Functions/MathTrig/SubTotalTest.php
+++ b/tests/PhpSpreadsheetTests/Calculation/Functions/MathTrig/SubTotalTest.php
@@ -141,4 +141,18 @@ class SubTotalTest extends AllSetupTeardown
         $sheet->getCell('A1')->setValue('=SUBTOTAL(9, B1:B9,#REF!,C1:C9)');
         self::assertEquals('#REF!', $sheet->getCell('A1')->getCalculatedValue());
     }
+
+    public function testNonStringSingleCellRefError(): void
+    {
+        $sheet = $this->getSheet();
+        $sheet->getCell('A1')->setValue('=SUBTOTAL(9, Sheet99!A1)');
+        self::assertEquals('#REF!', $sheet->getCell('A1')->getCalculatedValue());
+    }
+
+    public function testNonStringCellRangeRefError(): void
+    {
+        $sheet = $this->getSheet();
+        $sheet->getCell('A1')->setValue('=SUBTOTAL(9, Sheet99!A1)');
+        self::assertEquals('#REF!', $sheet->getCell('A1')->getCalculatedValue());
+    }
 }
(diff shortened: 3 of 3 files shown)


oleibman:
I am not duplicating your result in PhpSpreadsheet 1.23; in particular, I do not see a warning message. Please tell Php and PhpSpreadsheet release where you see a warning message (and the contents of the message).

There is at least one problem with your test case, so let's start with that. We currently cannot always distinguish the error situation `#REF!` from the string with the same content. You are testing the string case, and that's okay, but you should also be testing the error case. You might get that by using a cell from a non-existent sheet in your formula, e.g. `=Sheet99:A1`. I am a little surprised that the error condition and the string don't lead to the same result in 1.23, but the string leads to the same result as Excel `#REF!`, while the error condition leads to a different result (0) - your test should demonstrate that you have fixed that problem.

ndench:
### More details

Thanks for the comment @oleibman. I can confirm that I'm using PhpSpreadsheet 1.23.0 and PHP 8.1.6.
I see two different warning messages, both originating at [MathTrig\Subtotal#38](https://github.com/hyraiq/PhpSpreadsheet/blob/d0ff3d9ca7fc1b44cc2136ec427ca219572215c6/src/PhpSpreadsheet/Calculation/MathTrig/Subtotal.php#L38).

> Warning: Undefined array key 1
> Warning: Undefined array key 2


### New tests
I've added in 2 more tests as you requested, both of which fail with the `Undefined array key`, error (Excel returns `#REF!` for both of these). However I'm not sure of the best way to make `Subtotal.php` handle these cases, since the result from `Functions::flattenArrayIndexed($args)` is actually an array of `0` and `null` instead of `#REF!`. This leads me to believe that something needs to be fixed in the `flattenArrayIndexed` method. The only thing we could do in `Subtotal` is to something like this:

```
$foo = explode('.', $index);
if (\count($foo) !== 3) {
    return false;
}

[, $row, $column] = $foo;
```

But that leads to a `0` result instead of `#REF!`. I'd really appreciate some feedback on how you think this should be handled.


### RE: Unable to reproduce the issue:

I'm not sure why you can't reproduce it, but commenting out my changes to `Subtotal.php` gives be the following results with phpunit running on PHP 8.1.6:

```
root@b013d8eb1009:/var/www/html/vendor/phpoffice/phpspreadsheet# ./vendor/bin/phpunit --filter SubTotalTest
PHPUnit 9.5.20 #StandWithUkraine

..........................................EEEE                    46 / 46 (100%)

Time: 00:00.143, Memory: 76.00 MB

There were 4 errors:

1) PhpOffice\PhpSpreadsheetTests\Calculation\Functions\MathTrig\SubTotalTest::testRefError
PhpOffice\PhpSpreadsheet\Calculation\Exception: Worksheet!A1 -> Undefined array key 1

/var/www/html/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/Cell.php:293
/var/www/html/vendor/phpoffice/phpspreadsheet/tests/PhpSpreadsheetTests/Calculation/Functions/MathTrig/SubTotalTest.php:135

2) PhpOffice\PhpSpreadsheetTests\Calculation\Functions\MathTrig\SubTotalTest::testSecondaryRefError
PhpOffice\PhpSpreadsheet\Calculation\Exception: Worksheet!A1 -> Undefined array key 1

/var/www/html/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/Cell.php:293
/var/www/html/vendor/phpoffice/phpspreadsheet/tests/PhpSpreadsheetTests/Calculation/Functions/MathTrig/SubTotalTest.php:142

3) PhpOffice\PhpSpreadsheetTests\Calculation\Functions\MathTrig\SubTotalTest::testNonStringSingleCellRefError
PhpOffice\PhpSpreadsheet\Calculation\Exception: Worksheet!A1 -> Undefined array key 1

/var/www/html/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/Cell.php:293
/var/www/html/vendor/phpoffice/phpspreadsheet/tests/PhpSpreadsheetTests/Calculation/Functions/MathTrig/SubTotalTest.php:149

4) PhpOffice\PhpSpreadsheetTests\Calculation\Functions\MathTrig\SubTotalTest::testNonStringCellRangeRefError
PhpOffice\PhpSpreadsheet\Calculation\Exception: Worksheet!A1 -> Undefined array key 1

/var/www/html/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Cell/Cell.php:293
/var/www/html/vendor/phpoffice/phpspreadsheet/tests/PhpSpreadsheetTests/Calculation/Functions/MathTrig/SubTotalTest.php:156

ERRORS!
Tests: 46, Assertions: 42, Errors: 4.
```

And you can see the PHP 7.4 build is [now failing](https://github.com/PHPOffice/PhpSpreadsheet/runs/6874722162?check_suite_focus=true#step:10:241) with `Undefined offset` because of the new test cases.

oleibman:
Small changes are required in Calculation and SubTotal to get everything to work properly. Ironically, once these changes are made, your change to Subtotal is no longer necessary, as your tests will prove. For now, I am placing your change in draft status. If you are willing to expand your scope to include the change in Calculation (plus a few extra test cases that such a change will require), I can discuss that with you. Otherwise, I will open a PR to make the change. Let me know how you wish to proceed.

ndench:
Ok no problem. I'm not entirely sure which changes are required. If you're able to do up a PR im happy to test it 🙂

oleibman:
Closing, superseded by #2902. @ndench please test. I could certainly use some more test cases in RefErrorTest.

PR #2872, opened 2022-06-07
Add Support to Chart/Axis and Gridlines for Shadow

oleibman:
Continuing the work from #2865. Support is added for Shadow properties for Axis and Gridlines, and Glow and SoftEdges are extended to Gridlines. Tests are added. Some chart tests are moved from Reader/Xlsx and Writer/Xlsx so that most chart tests are under a single directory.

This is a minor breaking change. Since the support for these properties was just added, it can't really affect much in userland. Some properties had been stored in the form which the XML requires them rather than as the user would enter them to Excel. So, for example, setting the Glow size to 10 points would have caused it to be stored internally as 127,000. This change will store the size internally as 10, obviously making the appropriate conversion when reading from or writing to XML. This makes unit tests much simpler, and I think this is also what a user would expect, especially considering the difficulties in keeping track of the trailing zeros.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1171,17 +1171,7 @@ parameters:
 			path: src/PhpSpreadsheet/Chart/DataSeries.php
 
 		-
-			message: "#^Parameter \\#1 \\$angle of method PhpOffice\\\\PhpSpreadsheet\\\\Chart\\\\GridLines\\:\\:setShadowAngle\\(\\) expects int, int\\|null given\\.$#"
-			count: 1
-			path: src/PhpSpreadsheet/Chart/GridLines.php
-
-		-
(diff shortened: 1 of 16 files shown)


oleibman:
@MarkBaker, I think Gridlines are misconfigured. I believe they should be a property of Axis, not Chart. In the XML, they are a child of the valAx node. I also believe that the current setup allows only for horizontal gridlines and vertical is also permitted. I have some ideas on how handling a change in that area might proceed, but it would be a breaking change and a major distraction to the work I'm doing at the moment adding missing pieces and increasing coverage. Do you agree that the current setup might be a problem, and that it is okay to defer looking into it for now?

MarkBaker:
I agree that the gridlines shouldlogically be a property of the axis, and not of the chart... I'm just trying to remember why I wrote it that way (but it was many years ago, without the benefit of documentation, and I haven't really looked at the chart structures since then).

It should be possible to deprecate the existing structure/methods; and we have the 2.0dev branch that you can use branch from to experiment when you want to start work on the changes

PR #2875, opened 2022-06-08
Add removeCommentByColumnAndRow function

dgeppo:
This is:

```
- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

This function allows you to delete a comment

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -2588,9 +2588,9 @@ class Worksheet implements IComparable
      * @param array<int>|CellAddress|string $cellCoordinate Coordinate of the cell as a string, eg: 'C5';
      *               or as an array of [$columnIndex, $row] (e.g. [3, 5]), or a CellAddress object.
      *
-     * @return Comment
+     * @return $this
      */
-    public function removeComment($cellCoordinate):void
+    public function removeComment($cellCoordinate)
     {
         $cellAddress = Functions::trimSheetFromCellReference(Validations::validateCellAddress($cellCoordinate));
 
@@ -2605,6 +2605,8 @@ class Worksheet implements IComparable
         if (isset($this->comments[$cellAddress])) {
             unset($this->comments[$cellAddress]);
         }
+        
+        return $this;
     }
 
     /**
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -2590,7 +2590,7 @@ class Worksheet implements IComparable
      *
      * @return Comment
      */
-    public function removeComment($cellCoordinate)
+    public function removeComment($cellCoordinate):void
     {
         $cellAddress = Functions::trimSheetFromCellReference(Validations::validateCellAddress($cellCoordinate));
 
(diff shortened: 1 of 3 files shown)


src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -2581,6 +2581,22 @@ public function setComments(array $comments)
...
+    
+
+    /**
+     * Remove comment.
+     *
+     * @param int $columnIndex Numeric column coordinate of the cell
+     * @param int $row Numeric row coordinate of the cell
+     *
+     */
+    public function removeCommentByColumnAndRow($columnIndex, $row)
```
MarkBaker, line 2593: Useful to have a method to delete a cell comment, although you may have noticed that I'm trying to deprecate "byColumnAndRow" methods in favour of the cellAddress methods.
Every existing functionality has a cellAddress (or cellRange) method, not ever functionality has a "byColumnAndRow" method, so we already have inconsistency; and providing only a "byColumnAndRow" method for this functionality introduces even more inconsistency.
The `master` branch also has a slightly different approach for validating the address, which allows passing in a string with the cell address, an array of column/row ids, or a CellAddress object

So you might want to consider using the cellAddress approach instead (or at last as well as the "byColumnAndRow" method)

MarkBaker:
Useful functionality, but can you also provide unit tests to verify that the code works, thank you.

MarkBaker (changes requested):
See notes, and request for unit tests

dgeppo:
Thanks Mark for your feedback, I just updated and add test

src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -2582,6 +2582,31 @@ public function setComments(array $comments)
...
+        } elseif (strpos($cellAddress, '$') !== false) {
+            throw new Exception('Cell coordinate string must not be absolute.');
+        } elseif ($cellAddress == '') {
+            throw new Exception('Cell coordinate can not be zero-length string.');
+        }
+        // Check if we have a comment for this cell and delete it
+        if (isset($this->comments[$cellAddress])) {
+            unset($this->comments[$cellAddress]);
+        }
+    }
```
MarkBaker, line 2608: I think the only thing missing now is a return value... we don't generally return a `void` from the worksheet methods, but return `$this` to support the fluent interface

tests/PhpSpreadsheetTests/CommentTest.php
```diff
@@ -83,4 +84,12 @@ public function testSetText(): void
         $comment->setText($test);
         self::assertEquals('This is a test comment', (string) $comment);
     }
+    
+    public function testRemoveComment(): void {
+        $spreadsheet = new Spreadsheet();
+        $sheet = $spreadsheet->getActiveSheet();
+        $sheet->getComment('A2')->getText()->createText('Comment to delete');
+        $sheet->removeComment('A2');
```
MarkBaker, line 92: Probably also worth having an assert to ensure that `A2` has a comment after the create so that you can be sure that the `removeComment()` is removing something that was there

MarkBaker (changes requested):
Can you just address the couple of comments that I've made, thanks

dgeppo:
Thanks Mark for your feedback, I made the update following your remarks.

MarkBaker (commented):
Thank you for the submission, and the new functionality that you've added to the library.
I've fixed the cs issues that the pipeline highlighted: for future reference, running `./vendor/bin/phpcs --report=checkstyle src tests`, and fixed by running `./vendor/bin/phpcbf --report=checkstyle src tests` and `./vendor/bin/php-cs-fixer fix --format=checkstyle`

Issue #2885, opened 2022-06-14
Formatter - PercentageFormat - ArgumentCountError: 4 arguments are required, 2 given


FlameStorm:
This is a bug report according to reading loaded XLSX file raising a Fatal error in sprintf due to incorrect format string.

### What is the expected behavior?
Just normal working with loaded XLSX files.

### What is the current behavior?
Got a Fatal error during reading loaded XLSX file:

ArgumentCountError: 4 arguments are required, 2 given

/var/www/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Style/NumberFormat/PercentageFormatter.php:43
/var/www/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Style/NumberFormat/Formatter.php:146
/var/www/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Style/NumberFormat.php:404
/var/www/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Worksheet/Worksheet.php:2835

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$reader = \PhpOffice\PhpSpreadsheet\IOFactory::createReader("Xlsx");
$spreadsheet = $reader->load("./percents_test_case.xlsx");
$sheet = $spreadsheet->getActiveSheet();
$finishColumn = $sheet->getHighestColumn();
$rowsCount = $sheet->getHighestRow();
$rows = $sheet->rangeToArray("A1:{$finishColumn}{$rowsCount}");
// ^ Fatal error

```

The original (and external for us, we can't do anything with it's generation) loaded XLSX file contains format like `<numFmt numFmtId="84" formatCode="[$-010419]0%">` in `<numFmts ...>` block at /xl/styles.xml ("under zip"). 
Such custom/special formats with prepending `[$-...]` in formatCode are allowed as described in documentation for `numberingformat` at https://docs.microsoft.com/en-us/dotnet/api/documentformat.openxml.spreadsheet.numberingformat?view=openxml-2.8.1 and really can be set in MS Excel, LibreOffice Calc or another spreadsheet editor.

Sample file for showing the case:
[...]

PR #2886, opened 2022-06-14
Ignore square-$-brackets prefix in format string

FlameStorm:
This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [x] additional unit tests
```

Checklist:

- [x] Code style is respected

This PR fixes an issue https://github.com/PHPOffice/PhpSpreadsheet/issues/2885.
We can just ignore `[$-...]` part in `numberingformat` if specified, so this fix do this.


oleibman:
You will need to eliminate the Phpstan error. This should do the trick - change
```php
$format = preg_replace('/^\[\$-[^\]]*\]/', '', $format);
```
to
```php
$format = (string) preg_replace('/^\[\$-[^\]]*\]/', '', $format);
```
It would also be a good idea to add an assertion to show that the cell's format fits the pattern you're trying to fix.
```php
self::assertSame('[$-en-GB]0%', $sheet->getStyle('A1')->getNumberFormat()->getFormatCode();
```

FlameStorm:
> You will need to eliminate the Phpstan error. This should do the trick - change
> 
> ```
> $format = preg_replace('/^\[\$-[^\]]*\]/', '', $format);
> ```
> 
> to
> 
> ```
> $format = (string) preg_replace('/^\[\$-[^\]]*\]/', '', $format);
> ```
> 
> It would also be a good idea to add an assertion to show that the cell's format fits the pattern you're trying to fix.
> 
> ```
> self::assertSame('[$-en-GB]0%', $sheet->getStyle('A1')->getNumberFormat()->getFormatCode());
> ```

Thanks, oleibman! Fixed for Phpstan as you suggested. 
About an extra assert - I think that's not so much necessary, but added it too like you suggested, so seems it's ok now

oleibman:

PR #2902, opened 2022-06-20
Handling of #REF! Errors in Subtotal, and More

oleibman:
This PR derives from, and supersedes, PR #2870, submitted by @ndench. The problem reported in the original is that SUBTOTAL does not handle #REF! errors in its arguments properly; however, my investigation has enlarged the scope.

The main problem is in Calculation, and it has a simple fix. When the calculation engine finds a reference to an uninitialized cell, it uses `null` as the value. This is appropriate when the cell belongs to a defined sheet; however, for an undefined sheet, #REF! is more appropriate.

With that fix in place, SUBTOTAL still needs a small fix of its own. It tries to parse its cell reference arguments into an array, but, if the reference does not match the expected format (as #REF! will not), this results in referencing undefined array indexes, with attendant messages. That assignment is changed to be more flexible, eliminating the problem and the messages.

Those 2 fixes are sufficient to ensure that the original problem is resolved. It also resolves a similar problem with some other functions (e.g. SUM). However, it does not resolve it for all functions. Or, to be more particular, many functions will return #VALUE! rather than #REF! if this arises, and the same is true for other errors in the function arguments, e.g. #DIV/0!. This PR does not attempt to address all functions; I need to think of a systematic way to pursue that. However, at least for most MathTrig functions, which validate their arguments using a common method, it is relatively easy to get the function to propagate the proper error result.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [x] additional unit tests
```

Checklist:

- [x] Code style is respected


oleibman:
Yipe! Php80 is reporting error "Cannot use positional argument after named argument". This seems to be because SubTotal is using FlattenArrayIndexed, which creates an array of the form `stringkey1=>value1 intkey2=>value2`, and the string key is treated as a named argument, while the intkey is treated as positional. Changing SubTotal to use FlattenArray seems to work, but there are other functions which use FlattenArrayIndexed. This may well be a situation which has not existed in the test suite before this. Research is needed.

oleibman:
Unsurprisingly, using FlattenArray rather than FlattenArrayFixed fixes the problems reported in this test, but causes regression problems with other Subtotal tests.

Issue #2884, opened 2022-06-14
Attribute 'count' missing from /xl/sharedStrings.xml and there should be cleaner (with map) before write text to xl/sharedStrings.xml


ppodgorskicrido:
This is:
- [x] a bug report

I report this bug, because Excel (Office 365) shows that is need to repair xlsx file.

### What is the expected behavior?
In xlsx file, file xl/sharedString.xml:
- node <sst> should has attribute "count" like: count="10925"
- `&quot;` should be replaced by `"`
- `&#039;` should be replaced by `'`

### What is the current behavior?
In xlsx file, xl/sharedStrings.xml file:
- node <sst> has only 2 attributes: xmlns and uniqueCount
- `&quot;` and `&#039;`  is not converted to `"` and `'`

### What are the steps to reproduce?
Xlsx file was created from database (UTF-8). Text was extracted from HTML.
Text contain `&quot;` and `&#039;`.

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$writer = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet));
$writer->save("path.xlsx");
```

### What features do you think are causing the issue

- [ ] Reader
- [x] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calulations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

### Does an issue affect all spreadsheet file formats? If not, which formats are affected?
It could affect xls file.

### Which versions of PhpSpreadsheet and PHP are affected?
PhpSpreadsheet: 1.23
PHP: 8.1.6
Ubuntu: 20.04

ppodgorskicrido:
IMO htmlspecialchars should be removed:
[XMLWriter](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Shared/XMLWriter.php#L92)

Accord to [XML guidelines](https://www.w3.org/TR/xml/#syntax) `&` should be escaped by using `&amp;` (also `<`), but there is no need to escape `"` and `'` in text.

oleibman:
I do not understand the problem you are reporting. When you open the file in Excel, does the string in the cell not have the correct contents? Also, what harm is caused by the missing "count" attribute?

ppodgorskicrido:
The problem is that Excel won't open a xlsx file, which contains `&quot;` or `&#039;` in xl/sharedStrings.xml.

[...]

PR #2913, opened 2022-06-29
Apply chars limit in single cell in xlsx writer

ppodgorskicrido:
This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

#2884

### DIFF
--- a/src/PhpSpreadsheet/Shared/StringHelper.php
+++ b/src/PhpSpreadsheet/Shared/StringHelper.php
@@ -12,6 +12,11 @@ class StringHelper
     const STRING_REGEXP_FRACTION = '(-?)(\d+)\s+(\d+\/\d+)';
 
     /**
+     * Maximum limit of chars in single cell in excel.
+     */
+    private const CELL_CHARS_LIMIT = 32767;
+
+    /**
      * Control characters array.
      *
      * @var string[]
@@ -687,5 +692,18 @@ class StringHelper
         $v = (float) $textValue;
 
         return (is_numeric(substr($textValue, 0, strlen($v)))) ? $v : $textValue;
+    }
+
+    /**
+     * If text exceed chars limit in cell, it is truncated.
+     */
+    public static function applyCharsLimit(string $text): string
+    {
+        if (strlen($text) > self::CELL_CHARS_LIMIT) {
+            $text = substr($text, 0, self::CELL_CHARS_LIMIT);
+            trigger_error('Text exceeded max chars limit in single cell and was truncated: ' . self::CELL_CHARS_LIMIT, E_USER_WARNING);
+        }
+
+        return $text;
     }
 }
--- a/src/PhpSpreadsheet/Writer/Xlsx/StringTable.php
+++ b/src/PhpSpreadsheet/Writer/Xlsx/StringTable.php
@@ -97,7 +97,7 @@ class StringTable extends WriterPart
                 if ($textToWrite !== trim($textToWrite)) {
                     $objWriter->writeAttribute('xml:space', 'preserve');
                 }
-                $objWriter->writeRawData($textToWrite);
+                $objWriter->writeRawData(StringHelper::applyCharsLimit($textToWrite));
                 $objWriter->endElement();
             } elseif ($textElement instanceof RichText) {
                 $this->writeRichText($objWriter, $textElement);
@@ -185,7 +185,7 @@ class StringTable extends WriterPart
             // t
             $objWriter->startElement($prefix . 't');
             $objWriter->writeAttribute('xml:space', 'preserve');
-            $objWriter->writeRawData(StringHelper::controlCharacterPHP2OOXML($element->getText()));
+            $objWriter->writeRawData(StringHelper::applyCharsLimit(StringHelper::controlCharacterPHP2OOXML($element->getText())));
             $objWriter->endElement();
 
             $objWriter->endElement();
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
   - Treat rows/columns containing no cell records as empty (default)
   - Treat cells containing a null value as empty
   - Treat cells containing an empty string as empty
+- Apply cell limit (32,767 chars) in xlsx writer [Issue #2884](https://github.com/PHPOffice/PhpSpreadsheet/issues/2884)
 
 ### Changed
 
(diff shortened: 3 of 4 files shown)


src/PhpSpreadsheet/Shared/StringHelper.php
```diff
@@ -688,4 +693,17 @@ public static function testStringAsNumeric($textValue)
 
         return (is_numeric(substr($textValue, 0, strlen($v)))) ? $v : $textValue;
     }
+
+    /**
+     * If text exceed chars limit in cell, it is truncated.
+     */
+    public static function applyCharsLimit(string $text): string
+    {
+        if (strlen($text) > self::CELL_CHARS_LIMIT) {
```
MarkBaker, line 702: Is the limit characters, or is it bytes? `strlen()` measures the string in bytes; the PhpSpreadsheet `StringHelper::countCharacters()` method returns the character length

src/PhpSpreadsheet/Shared/StringHelper.php
```diff
@@ -688,4 +693,17 @@ public static function testStringAsNumeric($textValue)
...
+            $text = substr($text, 0, self::CELL_CHARS_LIMIT);
```
MarkBaker, line 703: Likewise with `substr()` which bases its count on bytes; `StringHelper::substring()` works with characters

src/PhpSpreadsheet/Writer/Xlsx/StringTable.php
```diff
@@ -97,7 +97,7 @@ public function writeStringTable(array $stringTable)
                 if ($textToWrite !== trim($textToWrite)) {
                     $objWriter->writeAttribute('xml:space', 'preserve');
                 }
-                $objWriter->writeRawData($textToWrite);
+                $objWriter->writeRawData(StringHelper::applyCharsLimit($textToWrite));
```
MarkBaker, line 100: Should this be applied on write? Or when the cell value is set? e.g. perhaps through the value binder

src/PhpSpreadsheet/Shared/StringHelper.php
```diff
@@ -11,6 +11,11 @@ class StringHelper
     //    Fraction
     const STRING_REGEXP_FRACTION = '(-?)(\d+)\s+(\d+\/\d+)';
 
+    /**
+     * Maximum limit of chars in single cell in excel.
+     */
+    private const CELL_CHARS_LIMIT = 32767;
```
MarkBaker, line 17: Is there a particular reason why you've made this constant private rather than public?

oleibman:
Aside from Mark's comments, there is a bigger issue here - your test case doesn't actually exercise the code in question. That is because the string is already truncated when the value is set in the cell - the code that does this is in method checkString in Cell/DataType, which is called by method setValueExplicit in Cell/Cell, which is called by both DefaultValueBinder and StringValueBinder. Are you using a custom binder? We would need code that shows how you manged to create this spreadsheet avoiding that check in the first place to figure out the proper solution.

ppodgorskicrido:
I'm using pure library (v 1.23.0) without any additions. This project is based on Symfony with Doctrine ORM. Code is simple:
```php
<?php
/* Add columns names on top of excel sheet */
$rowArray = [
    // array of strings as columns headers
];
$sheet->fromArray($rowArray);

//then assing values using:
$sheet->setCellValueByColumnAndRow(1, $row, $entity->getDescription()); // function returns string|null

$writer = new XlsxWriter($spreadsheet); // alias of ...\Writer\Xlsx
try {
    $writer->save($this->targetDirectory . '/' . $exportFilePath);
} catch (Exception $e) {
    $this->logger->error('Cannot save excel to path: ' . $exportFilePath);
    $this->logger->error('Error message: ' . $e);
}
unset($spreadsheet);
```

oleibman:
I am not able to duplicate your result:
```php
require __DIR__ . '/PhpSpreadsheet' . '/vendor/autoload.php';

$text = str_repeat('Ω', 35000);
$sheet->setCellValueByColumnAndRow(1, 1, $text);
$sheet->setCellValueByColumnAndRow(1, 2, '=LEN(A1)');
echo 'Length of data in cell A1 is ',
    $sheet->getCell('A2')->getCalculatedValue(),
    "\n";
$filename = 'issue.2884b.xlsx';
$writer->save($filename);
echo "saved $filename\n";
```
This should be approximately equivalent to your code. Note that I am using a string composed entirely of multiple byte UTF8 characters, just to make sure this isn't a bytes-vs-characters issue. When I run it, I see what I expected, which is not what you are seeing:
```
Length of data in cell A1 is 32767
saved issue.2884b.xlsx
```
And the output file opens with no problem. Can you try running this to see your results.

oleibman:
I hate to say it, but this looks like an Excel bug to me. I have read in the file where you originally reported a problem, and confirmed using mb_strlen that the length of the data in J2 is 32,767 characters. It is truncated, with the last characters being "a fryzury to dzieło W". Now, when I open a new spreadsheet and paste that text into a cell, and use LEN to see the length of that cell, it reports 32,751! I think this difference might be because there are 4 ampersand characters in the text, which have been translated from `&` to `&amp;`, resulting in an extra 4*4=16 characters. I don't understand why this should be so, but it is consistent with the facts. Then, maybe there is some merit in your original suggestion that `&quot;` is pushing it over the limit. However, if it is, that's an Excel bug. And, in any case, there's no sensible action we can take. If we arbitrarily truncate at 32,767 _after_ htmlspecialchars, we may wind up cutting off part of `&amp;` or `&quot;`, winding up with our xml corrupted in a different way.
BTW, I tried adding ENT_NOQUOTES to the htmlspecialchars call in XMLWriter. This did, in fact, leave the quotes unchanged,but it was not sufficient to allow opening the resulting file without error.

oleibman:
I think I can confirm a bug in Excel. Run the following program:
```php
$longstring1 = str_repeat('a', 32767);
$longstring2 = str_repeat('b', 32767);
echo mb_strlen($longstring1, 'UTF-8'), "\n";
echo mb_strlen($longstring2, 'UTF-8'), "\n";
$filename = 'issue.2884.toolong1d.csv';
$stream = fopen($filename, 'wb');
fputs($stream, "$longstring1\n$longstring2\n");
fclose($stream);
echo "write complete to $filename\n";
```
You can easily confirm that, as expected, this file shows two 32,767-character lines when viewed in a text editor. Not so when you open it in Excel - cell A1 contains the first 32,759 a's, and cell A2 contains the last 7. One appears to have been lost altogether. Likewise, cell A3 contains the first 32,759 b's, A4 contains the last 7, and one is missing. LibreOffice (which does not have a 32K restriction) shows the data as 32,767 characters in each of A1 and A2. Note that this demo does not depend on PhpSpreadsheet, uses only ASCII characters, and does not use XML replacement for the ampersands.

oleibman:
Closing this PR. The problem appears to be with Excel, not with PhpSpreadsheet, and the code in this PR won't fix it. I will leave the original issue open so that the problem remains documented.

PR #2921, opened 2022-07-02
Keep Calculated String Results Below 32K

oleibman:
This is the result of an investigation into issue #2884 (see also PR #2913). It is, unfortunately, not a fix for the original problem; see the discussion in that PR for why I don't think there is a practical fix for that specific problem at this time.

Excel limits strings to 32,767 characters. We already truncate strings to that length when added to the spreadsheet. However, we have been able to exceed that length as a result of the concatenation operator (Excel truncates); as a result of the CONCATENATE or TEXTJOIN functions (Excel returns #CALC!); or as a result of the REPLACE, REPT, SUBSTITUTE functions (Excel returns #VALUE!). This PR changes PhpSpreadsheet to return the same value as Excel in these cases. Note that Excel2003 truncates in all those cases; I don't think there is a way to differentiate that behavior in PhpSpreadsheet.

However, LibreOffice and Gnumeric do not have that limit; if they have a limit at all, it is much higher. It would be fairly easy to use existing settings to differentiate between Excel and LibreOffice/Gnumeric in this respect. I have not done so in this PR because I am not sure how useful that is, and I can easily see it leading to problems (read in a LibreOffice spreadsheet with a 33K cell and then output to an Excel spreadsheet). Perhaps it should be handled with an additional opt-in setting.

I changed the maximum size from a literal to a constant in the one place where it was already being enforced (Cell/DataType). I am not sure that is the best place for it to be defined; I am open to suggestions.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Calculation/Calculation.php
```diff
@@ -4711,11 +4712,19 @@ private function processTokenStack($tokens, $cellID = null, ?Cell $cell = null)
                                 //    Perform the required operation against the operand 1 matrix, passing in operand 2
                                 $matrixResult = $matrix->concat($operand2);
                                 $result = $matrixResult->getArray();
+                                if (isset($result[0][0])) {
+                                    $result[0][0] = mb_substr($result[0][0], 0, DataType::MAX_STRING_LENGTH, 'UTF-8');
```
MarkBaker, line 4716: Is it better to call the `SharedString::substring()` method for consistency?

src/PhpSpreadsheet/Calculation/TextData/Concatenate.php
```diff
@@ -23,7 +25,18 @@ public static function CONCATENATE(...$args): string
...
 
         foreach ($aArgs as $arg) {
+            $value = Helpers::extractString($arg);
+            if (ErrorValue::isError($value)) {
+                $returnValue = $value;
+
+                break;
+            }
             $returnValue .= Helpers::extractString($arg);
+            if (mb_strlen($returnValue, 'UTF-8') > DataType::MAX_STRING_LENGTH) {
```
MarkBaker, line 35: Is it better to call the `Shared\StringHelper::countCharacters()` method for consistency?

src/PhpSpreadsheet/Calculation/TextData/Concatenate.php
```diff
@@ -56,15 +69,27 @@ public static function TEXTJOIN($delimiter, $ignoreEmpty, ...$args)
...
             if ($ignoreEmpty === true && is_string($arg) && trim($arg) === '') {
                 unset($aArgs[$key]);
             } elseif (is_bool($arg)) {
                 $arg = Helpers::convertBooleanValue($arg);
             }
         }
 
-        return implode($delimiter, $aArgs);
+        $returnValue = ($returnValue !== '') ? $returnValue : implode($delimiter, $aArgs);
+        if (mb_strlen($returnValue, 'UTF-8') > DataType::MAX_STRING_LENGTH) {
```
MarkBaker, line 88: As above, `Shared\StringHelper`

oleibman: Done.

src/PhpSpreadsheet/Calculation/TextData/Concatenate.php
oleibman: Yes, it is. I still haven't figured out where *all* the bodies are buried :-). Change is made. Note that a small change was necessary for countCharacters, allowing the specifying of the length parameter as null.

Issue #1635, opened 2020-09-01
DateTime 4 years difference because Date::setExcelCalendar


ybelenko:
This is:

```
- [ ] a bug report
- [x] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

I've spent almost 5 hours to investigate weird bug. I've made a small parser application and covered it with unit tests. I didn't understand why my assert of `DateTime`s fails when I add new test to completely unrelated class. As it turned out PHPUnit runs all tests simultaneously, so if one test contains reading of file in `.numbers` format and second test handles `.xlsx` then the `PhpOffice\PhpSpreadsheet\Shared\Date::$excelCalendar` value will be `self::CALENDAR_WINDOWS_1900` for both tests which is wrong.

### What is the expected behavior?
I don't know how to fix this issue with current class architecture, it's obviously you cannot switch from static method to internal one.

### What is the current behavior?
When I load Mac file and then Win file I expect to calculate `DateTime` from Mac file correctly.

### What are the steps to reproduce?

1. Load spreadsheet `A` file created on Windows system.
2. Load spreadsheet `B` file created on Mac system.
3. Get value from datetime cell of the file `A` and convert it to `DateTime` with `Date::excelToDateTimeObject` method. Resulted `DateTime` instance will be different from original file(4 years different to be more specific).

### Which versions of PhpSpreadsheet and PHP are affected?
- `PHP 7.4.6 (cli) (built: May 28 2020 23:40:38) ( NTS )`
- `"phpoffice/phpspreadsheet": "^1.14"`

PR #2937, opened 2022-07-13
2.x calendar changes

MarkBaker:
This is:

```
- [ ] a bugfix
- [X] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [X] Code style is respected

Calendar setting (1900/1904) from a loaded spreadsheet is now saved against the spreadsheet, allowing handling of multiple spreadsheets with different settings, rather than maintained as a global setting. The static global setting is still used for calcuation/formatting, but is set from the spreadsheet when required, then reset afterwards.

This does mean it's dangerous to rely on the old setting, because it is no longer guaranteed to be consistent


oleibman:
@MarkBaker - this change is probably a fix for issue #1635

PR #2949, opened 2022-07-19
Fix Cell Coordinate Error, such as Undefined array key 1,2,3,4,5,6,7,…

hurricanetx:
…8,9,0

Reason: $columnLookup no exists string to number mapper ,eg: '1' => 1

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Error when operating on a cell containing numbers: very serious so fixed.


MarkBaker:
What is the "very serious bug" that this PR fixes? Where is `columnIndexFromString()` being called with a numeric value rather than an alpha `A` -> `XFD` string address? Why should `columnIndexFromString()` be changed so drastically? If you're creating a PR it should be linked to an Issue, what Issue does this PR resolve? Is there a test case for this issue?

`columnIndexFromString()` was written to convert an alpha column address to its numeric equivalent; if there is somewhere in the codebase where a numeric column index is being passed to this method, then it should really be resolved at source rather than by changing this method

MarkBaker:
And the implementation won't work correctly if the numeric value passed in is more than one digit, passing in '11' will return 27; or if it's an alphnumeric like 'A1' or '1A', (which will also return 27)

hurricanetx:
Sorry, I made a mistake and submitted the wrong code. This bug does not exist, and the solution to fix it is incorrect, thank you for your patience in explaining it.

PR #2959, opened 2022-07-22
Add throws tag to Writer::save method

VincentLanglet:
This is:

```
- [x] phpdoc
```

HI @MarkBaker, some implementation of this method are throwing an Exception, to help the user using try/catch when calling this method the phpdoc should mention about the thrown exception.
This is also useful for static analysis tools which checking if exception are caught.


VincentLanglet:
Maybe you have time to review this @oleibman ?

oleibman:
I am not really in a position to evaluate this change. The `@throws` comments were all removed 2.5 years ago (https://github.com/PHPOffice/PhpSpreadsheet/commit/7e79782dae60edeea90040bf9abec25c8befaea7). 89 files were changed; I don't know what would differentiate this one from any of the others.

VincentLanglet:
> I am not really in a position to evaluate this change. The `@throws` comments were all removed 2.5 years ago ([7e79782](https://github.com/PHPOffice/PhpSpreadsheet/commit/7e79782dae60edeea90040bf9abec25c8befaea7)). 89 files were changed; I don't know what would differentiate this one from any of the others.

Hi, this throws tag is not different from the others. I can add others too if wanted, it's just that I discovered this one was missing especially because I used this method and needed to catch the exception.

They were removed with the argument that "they had no value" and this is terribly wrong.

When I call 
```
$this->foo();
``
`
if it can throws exception I can end up with a 500 in production.
When the phpdoc is preventing me an exception can be thrown, I can do
```
try {
    $this->foo();
}.catch (\Exception $e) {
    $this->handle($e);
}
```

This is more important now than in the past because statis analysis tools are now developed and they are able to detect missing try/catch.

Issue #2953, opened 2022-07-20
Unsupported operand types in Matrix.php:746


u01jmg3:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

This following formula is supported.

```
=SUM(4 * INDEX(Lookup!$B$2, MATCH($A2, Lookup!$A$2, 0)))
```

### What is the current behavior?

```
Fatal error: Uncaught Error: Unsupported operand types in vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Shared\JAMA\Matrix.php:746
Stack trace:
#0 vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Calculation\Calculation.php(5152): PhpOffice\PhpSpreadsheet\Shared\JAMA\Matrix->arrayTimesEquals(Array)
#1 vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Calculation\Calculation.php(4675): PhpOffice\PhpSpreadsheet\Calculation\Calculation->executeNumericBinaryOperation(Array, Array, '*', 'arrayTimesEqual...', Object(PhpOffice\PhpSpreadsheet\Calculation\Token\Stack))
#2 vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Calculation\Calculation.php(3632): PhpOffice\PhpSpreadsheet\Calculation\Calculation->processTokenStack(Array, 'B2', Object(PhpOffice\PhpSpreadsheet\Cell\Cell))
#3 vendor\phpoffice\phpspreadshee in vendor\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Shared\JAMA\Matrix.php on line 746
```

Looks like multiplication is being attempted with an array.

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$reader       = new \PhpOffice\PhpSpreadsheet\Reader\Xlsx();
$spreadsheet  = $reader->load('test.xlsx');
$worksheet    = $spreadsheet->getActiveSheet();
$fileContents = $worksheet->toArray();
[...]

PR #2964, opened 2022-07-28
Ensure multiplication is performed on a non-array value

u01jmg3:
This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

Fixes https://github.com/PHPOffice/PhpSpreadsheet/issues/2953


src/PhpSpreadsheet/Shared/JAMA/Matrix.php
```diff
@@ -742,6 +743,9 @@ public function arrayTimesEquals(...$args)
                         $value = trim($value, '"');
                         $validValues &= StringHelper::convertToNumberIfFraction($value);
                     }
+                    if (is_array($value)) {
+                        $value = Functions::flattenArray($value)[0];
```
u01jmg3, line 747: This could be more robust and changed to:

```php
$flattened = Functions::flattenArray($value);
$firstKey = array_key_first($flattened);
$value = $flattened[$firstKey];
```

MarkBaker: The whole of the JAMA library should really be replaced completely; the code is awful and without tests, and it's no longer a supported library. Unfortunately, I've yet to find a matrix library that actually uses MS Excel matrix logic, so I've been building my own but that's still Work-in-Progress

u01jmg3: Would be curious to see your WIP when it's ready.

Can we still progress this fix? I've applied it here but that's not to say it couldn't be caught and rectified higher up in the code. This is just my assessment and what I've found to fix the problem.

u01jmg3:
@oleibman, @MarkBaker: anything more I can do to get this merged? Thanks

PR #2979, opened 2022-08-02
MtJpGraph support added

f1mishutka:
This is:
```
- [X] refactoring
```

[jpgraph/jpgraph](https://packagist.org/packages/jpgraph/jpgraph) package seems abandoned and outdated. There is a maintained fork [mitoteam/jpgraph](https://packagist.org/packages/mitoteam/jpgraph) with latest library sources version and php 8.1 support. This PR adds ability to use it instead jpgraph/jpgraph.

We do not want to break existing compatibility with jpgraph/jpgraph, so there is JpGraphRendererBase.php implementation of IRenderer interface added with two descendants JpGraph (old one) and MtJpGraphRenderer (new one).

So now it is possible to use latest jpgraph version with PhpSpreadsheet without additional dancing or manual sources pathing.
* `composer require mitoteam/jpgraph`
* `Settings::setChartRenderer(\PhpOffice\PhpSpreadsheet\Chart\Renderer\MtJpGraphRenderer::class);`

Changelog and readme updated as well.


oleibman:
I am not really knowledgeable enough about other packages to decide whether this should move forward or not. I will have to defer to @MarkBaker  or @PowerKiKi for that. However, I can suggest some changes which will be needed beforehand.

You have to be able to pass Phpstan. The way Jpgraph is implemented causes many Phpstan problems. In phpstan.neon.dist, we hide those problems with the following line:
```
    excludePaths:
        - src/PhpSpreadsheet/Chart/Renderer/JpGraph.php
```
I think it would be okay if you added your Renderer code similarly. I don't know if that's a good idea long-term, but we might be able to evaluate how to approach that after we decide that we want to move this forward.

The test Helper/SampleTest.php specifically excludes some Jpgraph tests.
```php
    public function providerSample(): array
    {
        $skipped = [
            'Chart/32_Chart_read_write_PDF.php', // Unfortunately JpGraph is not up to date for latest PHP and raise many warnings
            'Chart/32_Chart_read_write_HTML.php', // idem
            'Chart/35_Chart_render.php', // idem
        ];
```
We would definitely want to see that these samples work correctly with your code. You would probably have to make changes to them (and remove them from the skipped array), or clone them (without changing skipped), to use your code. These aren't really formal tests, but you can review their output after running the full test suite and confirm that everything looks okay.

f1mishutka:
To make things little bit clear:

mitoteam/jpgraph is not some JpGraph fork. This is just original JpGraph library sources maintained as composer package. So this is exactly the same what jpgraph/jpgraph package was. Two important differences:
- latest (and updated!) version of original jpgraph library sources - so we have php 8.1 support
- different approach to load library and modules in composer autoloader. That is why I implemented separate renderer for it (oh, and because another classes namespace as well).

Renderer code is not changed at all. That is why I moved it all to JpGraphRendererBase class, so there is only protected `init()` method is overridden in two different implementations. The goal was to keep old code working exactly same way as before allowing to use mitoteam/jpgraph as replacement if required.

_> You have to be able to pass Phpstan. The way Jpgraph is implemented causes many Phpstan problems. In phpstan.neon.dist, we hide those problems with the following line_

Done. Thank you for suggestion!

We also have confirmation that my code is working well from @rolinger here: https://github.com/mitoteam/jpgraph/issues/7

P.S.: we use both JpGraph and PhpSpreadsheet in our projects but we do not use charts functionality, so there was no intersection for them before now.

oleibman:
Reviewing your change, I see you have enabled the use of mitoteam without actually making it part of the project. That actually seems like a good staged approach - I can take care of integrating it into the test suite later. I have, in the meantime, done so in a test version, and it seems to produce reasonable results. So I am willing to move forward with it as is. However, you need to resolve the merge conflicts before I can do so. (Github won't show me the problem, but I suspect changelog is the culprit.)

f1mishutka:
> Reviewing your change, I see you have enabled the use of mitoteam without actually making it part of the project. That actually seems like a good staged approach - I can take care of integrating it into the test suite later. 

Yes, the idea was just to give it a try. If this will go well without major problems there will be a possibility to replace outdated jpgraph/jpgraph package completely.

> I have, in the meantime, done so in a test version, and it seems to produce reasonable results. So I am willing to move forward with it as is.

Great news, thank you! Now we have at least two confirmations that this approach works.

> However, you need to resolve the merge conflicts before I can do so. (Github won't show me the problem, but I suspect changelog is the culprit.)

There was just minor conflict in changelog. Resolved and merged.

f1mishutka:
Also rebased on latest upsteam and squashed all changes into one commit.

oleibman:
Thank you for your contribution. I look forward to re-enabling the abandoned tests.

f1mishutka:
Hi, @oleibman!

Added LICENSE.md file with license notes to mitoteam/jpgraph . Could you please take a look if this enough?

As for PHP8.2: waiting for php-redis package provided for php8.2 for us to be able to test and fix jpgraph under php 8.2 (we use redis a lot in our project). Anyway we plan to release jpgraph with php 8.2 support as soon as php8.2 will be released (we have strong intention to switch to php 8.2 for our project it as soon as possible).

Thank you for your efforts with tests and checks!

oleibman:
Thanks for adding license info. I reiterate that IANAL; I am, however, definitely happier having the license available than not. Since your project is not part of the "production" release for PhpSpreadsheet, I think what you specify is definitely sufficient.

When it comes time to test Php8.2, 2 of the samples which I have left enabled (Chart/32_Chart_read_write_HTML.php and the same for PDF) are good test candidates - they work but issue a few messages. Chart/35_Chart_render.php, which I have disabled, issues a ton of messages, so, once the other 2 are working, that would be the next step.

f1mishutka:
Hi, @oleibman!

Just added v10.2.0-beta of mitoteam/jpgraph with patches for PHP 8.2 compatibility. Could you please run tests and let me know if any warnings or deprecation notices left? Thank you.

oleibman:
I didn't notice any problems - good job!

Issue #2790, opened 2022-04-29
Install via composer error with psr/simple-cache latest version


Ahmed-Elrayes:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
install via composer

### What is the current behavior?
i'm using laravel 9 which comes with psr/simple-cache v3

 Problem 1
    - Root composer.json requires phpoffice/phpspreadsheet ^1.23 -> satisfiable by phpoffice/phpspreadsheet[1.23.0].
    - phpoffice/phpspreadsheet 1.23.0 requires psr/simple-cache ^1.0 || ^2.0 -> found psr/simple-cache[1.0.0, 1.0.1, 2.0.0, 2.x-dev] but the package is fixed to 3.0.0 (lock file version) by a partial update and that version does not match. Make sure you list it as an argument for the update command.

i've seen this issue [psr/simple-cache:3.0.0 #2703](https://github.com/PHPOffice/PhpSpreadsheet/issues/2703)
but unfortunately i need it to be used on the current project i'm working on.

### What are the steps to reproduce?
support [psr/simple-cache](https://packagist.org/packages/psr/simple-cache) version 3.0

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

If this is an issue with reading a specific spreadsheet file, then it may be appropriate to provide a sample file that demonstrates the problem; but please keep it as small as possible, and sanitize any confidential information before uploading.

### What features do you think are causing the issue

- [ ] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calulations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

### Which versions of PhpSpreadsheet and PHP are affected?
PHP8.0 and PhpSpreadsheet 1.23

MarkBaker:
As noted in the discussions related to that issue:
> Because of the use of mixed type in simple-cache 3.0, we can't match our implementation to the simple-cache 3.0 interface until we support only PHP8 and above. The only way we could work it would be to have two version-specific implementations, and a run-time identifier that chose which one to use.

[...]

PR #2989, opened 2022-08-04
Update psr/simple-cache 3.0

lovermans:
Update composer.json to support psr/simple-cache 3.0

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -75,7 +75,7 @@
         "markbaker/matrix": "^3.0",
         "psr/http-client": "^1.0",
         "psr/http-factory": "^1.0",
-        "psr/simple-cache": "^1.0 || ^2.0"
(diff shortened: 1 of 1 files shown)


MarkBaker:
No! No! No! If you insist on support for `psr/simple-cache 3.0` then you **MUST** provide backward compatibility for `psr/simple-cache 1.0` and `psr/simple-cache 2.0`, which means that you must provide implementations of Cell Collection that match both interfaces, so that PHP versions prior to 8.0 will not fatal error, and preferably without any performance degradation.

MarkBaker:
[Please read](https://github.com/PHPOffice/PhpSpreadsheet/issues/2790) to understand the problems with this

lovermans:
Thank's alot.

PR #2994, opened 2022-08-07
Use Only mb_convert_encoding in StringHelper sanitizeUTF8

### LINKED ISSUE #2982: UConverter
This is:

```
- [x] a bug report
- [ ] a feature request
```

Installing the library with composer and being able to use it without dependency errors.

After installing the library with composer and creating the spreadsheet I get an error on the UConverter class which is not found.

1 - Install with composer
2 - Example for create file with one cell value

```php
<?php

$writer->save('hello world.xlsx');

```
The error line is into Shared/StringHelper ad line 337.

- [ ] Reader
- [x] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

For Xlsx file

1.24.1 - 2022-07-18 Latest

oleibman:
This started out as a slightly different fix, but experimenting showed it could be simplified. SanitizeUTF8 uses all of UConverter, iconv, and mb_convert_encoding, but the last of those 3 is sufficient on its own. So eliminate the other 2, incidentally eliminating the need for php-intl in this function. The history of this change follows:

Fix #2982. That issue is actually closed, but it did expose a problem. Our test environments all enable php-intl, but that extension isn't a formal requirement for PhpSpreadsheet. Perhaps it ought to be. Nevertheless ...

Using UConverter for string translation solved some problems for us. However, it is only available when php-intl is enabled. The code tests if it exists before using it, so no big deal ... except it seems likely that the people reporting the issue not only did not have php-intl, but they do have their own autoloader which issues an exception when the class isn't found. The test for existence of UConverter defaulted to attempting to autoload it if not found. So, on a system without php-intl but with a custom autoloader, there is a problem. Code is changed to suppress autoload when testing UConverter existence.

Pending this fix, the workaround for this issue is to enable php-intl.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [ ] Code style is respected


MarkBaker (commented):
I'd probably also be inclined to add php-intl in the composer.json as a suggestion.

For one of my existing PRs, I'd been looking at duplicating the test suite (one set with php-intl, the other without), but hadn't done so because I couldn't see how to merge the code coverage

oleibman:
As you can see, I eliminated php-intl in this module after all. I agree that suggesting it is still a good move; I know that you've been on the verge of it with some of the translation and locale and other work that you've been up to. I won't be including that with this change. However, now that the mitoteam changes are merged, I'm about to start work on replacing jpgraph with it in the test suite; that will require changes to composer.json, and I will add the php-intl suggestion at the same time.

PR #3081, opened 2022-09-24
Calculation suppressFormulaErrors - Minor Breaks

### LINKED ISSUE #1531: suppressFormulaErrors is not configurable
This is:

```
- [ ] a bug report
- [x] a feature request
```

If a formula is invalid, the behaviour is controlled by Calculation::raiseFormulaError that either throws an exception or triggers an error

This is controlled by the suppressFormulaErrors public variable, but there is no easy way to configure that variable at the spreadsheet level. More over, having a public variable is a bit strange (why not a config option on the spreadsheet with proper getters/setters?). Also the doc blocks of raiseFormulaError are not in line with regular docblocks

```
// trigger an error, but nicely, if need be
protected function raiseFormulaError($errorMessage)
```

It would be nice to actually allow suppressing errors to allow excel generation and not block the process because of one error in a formula (currently, both errors and exception will block the generation of the file)

As an added bonus, it would be nice if the error message could include the actual formula, in large excel file in can be a real mess to find back which formula is causing the error and why it's invalid. having the cell reference is nice, but the actual formula would be super helpful in my opinion.

If a formula contains an error, it will trigger an error. I don't see how to throw an exception, although it wouldn't suppress anything (contrary to what the variable name would let me believe)

If you want to skip errors, you have to comment out everything in the raiseFormulaError method

    // trigger an error, but nicely, if need be
    protected function raiseFormulaError($errorMessage)
    {
        $this->formulaError = $errorMessage;
        $this->cyclicReferenceStack->clear();
        // if (!$this->suppressFormulaErrors) {
            // throw new Exception($errorMessage);
        // }
        // trigger_error($errorMessage, E_USER_ERROR);

        return false;
    }

Just make any sheet with a formula error in it, it will show a Formula Error

```php
<?php

$sheet->setCellValueByColumnAndRow(1, 1, '=SOMEERRORFUNCTION(A1:A10)');

$writer->save('php://output');
```

1.8.2 to 1.13

oleibman:
Fix #1531. Calculation has a property `suppressFormulaErrors`, which really doesn't work as one might expect. If a calculation throws an exception, the setting of this property might prevent the Exception from being thrown, but it will still trigger an Error. I do not think this makes sense, and will change it so the calculation will return `false`, which is part of the original design but which would essentially never happen. This allows the user to save a corrupt spreadsheet, but this was already possible through the use of `setPreCalculateFormulas(false)` on the Writer, so this doesn't really open any new exposures. It nevertheless might be considered a breaking change because of the difference in behavior.

Another break - the visibility of the property is changed from public to private, and a public setter/getter is added.

Although I am enabling this ability, I don't necessarily think it's a good idea to make use of it. See the original issue for a discussion of why. It is not mentioned in the official documentation, and I will not be adding documentation for it. The originator discovered it by reading the code, and I think that is sufficient for what will often be an ill-advised choice.

Many of the large number (unfortunately not quite all) of problems with Calculation.php in phpstan baseline are addressed. Of note are that private arrays `phpSpreadsheetFunctions`, `controlFunctions`, `comparisonOperators`, and `operatorPrecedence` are changed from static to const.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected


oleibman:
No interest in resolving Scrutinizer failure. One method needed to add some tests for null in order to satisfy Phpstan; Scrutinizer says these extra checks changed complexity from B to C, which is a failure condition. C'est la vie.

oleibman:
Phpstan seems to be timing out, for reasons which I do not understand. I may have to cancel this PR and submit another in its place.

oleibman:
Phpstan failure does not seem to be the result of a botched merge. I will resubmit a less ambitious PR.

oleibman:
Superseded by 3092. Closing.

PR #3094, opened 2022-09-30
Reduces extra memory usage on `__destruct()` calls

josecl:
This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

After reading a large worksheet of 1000 rows x 100 columns I call `$spreadsheet->__destruct()` to reduce the memory consumption and prevent to hit the PHP limit for the process.

It goes down from 56mb to 51mb. But, after calling `__destruct()` the `memory_get_peak_usage` increased from 56mb to 61mb. It eats some extra megabytes while freeing resources.

I made a little change to avoid call `array_keys()` on a big array.


oleibman:

PR #3108, opened 2022-10-07
Fix Remainder of Calculation vs. Phpstan Issues

oleibman:
I had tried to include these changes as part of an earlier effort, but something about them broke Phpstan. I removed them until I could determine the actual cause, which is ...

The array `$phpSpreadsheetFunctions` is a very large and very complicated array, typehinted as `array`. It is declared as private static; however, its content never changes (at least not now - there are some outstanding proposals that might change that). I have had some success with changing unmodifiable private static to private const. However, such a change here causes Phpstan to perform a lot more processing and eventually time out. So, leave it as static.

Having identified the cause of the problem, none of the other changes were problematic, so this PR applies the rest of them.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected


oleibman:
Not even remotely concerned by Scrutinizer failing due to complexity ratings, which Scrutinizer itself caused by requiring additional tests.

PR #3112, opened 2022-10-08
Added changes to the readme

yusuf-saif:
Added changes to the readme

This change is needed to remove ambiguity and make It more clear to people


MarkBaker:
There are more grammatical errors (what does `that we use other tools like MicroSoft Excel, LibreOffice Calc, WPS and others` mean) and factual errors (e.g. what is `xlv` format?) than before your change; and is there any value in capitalising heading titles that are already highlighted as titles in the markdown?

If this submission is for HacktoberFest credit, we aren't registered as a HactoberFest project.

PR #3113, opened 2022-10-09
Add other country currency IDR, MYR, ZAR, JPY, CNY, RUB, PHP

NajibCoding:
This is:

```
- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [ ] Code style is respected


MarkBaker:
Should we really just pack the library with constants for all these currencies? The format code is a string value that can be set as a string, and users that want constants for different currencies can always set them in userland code

Alternatively, some form of wizard that allows the easy creation of currency and accounting format masks would be preferable to all these constants

PR #3122, opened 2022-10-13
Eliminate Some Scrutinizer 'Major' Problems Part 2

oleibman:
Mostly Scrutinizer bizarre false positives. From Calculation/Engineering/Bessell (and several others):
```php
     private static function besselI2(float $x, int $ord): float
     {
        if ($x === 0.0) {
             return 0.0;
         }
```
Scrutinizer complains that `$x` can never equal 0.0 here. Huh???

Another example repeated several times, from Calculation/Engineering/BesselJ:
```php
$bool = false;
foreach (whatever) {
    ...
    $bool = !$bool;
    ...
}
```
Scrutinizer complains about the `!$bool` assignment because it says `$bool` is always false. Again, huh???

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Calculation/Engineering/BesselI.php
+++ b/src/PhpSpreadsheet/Calculation/Engineering/BesselI.php
@@ -111,9 +111,16 @@ class BesselI
         return ($x < 0.0) ? -$ans : $ans;
     }
 
+    /**
+     * Sop to Scrutinizer.
+     *
+     * @var float
+     */
+    private static $zeroPointZero = 0.0;
+
     private static function besselI2(float $x, int $ord): float
     {
-        if ($x === 0.0) {
+        if ($x === self::$zeroPointZero) {
             return 0.0;
         }
 
--- a/src/PhpSpreadsheet/Calculation/Engineering/BesselJ.php
+++ b/src/PhpSpreadsheet/Calculation/Engineering/BesselJ.php
@@ -167,7 +167,7 @@ class BesselJ
             if ($jsum === true) {
                 $sum += $bj;
             }
-            $jsum = !$jsum;
+            $jsum = self::logicalNot($jsum);
             if ($j === $ord) {
                 $ans = $bjp;
             }
@@ -176,5 +176,11 @@ class BesselJ
         $ans /= $sum;
 
         return ($x < 0.0 && ($ord % 2) === 1) ? -$ans : $ans;
+    }
+
+    // Sop to scrutinizer
+    private static function logicalNot(bool $arg): bool
+    {
+        return !$arg;
     }
 }
--- a/src/PhpSpreadsheet/Calculation/Financial/Coupons.php
+++ b/src/PhpSpreadsheet/Calculation/Financial/Coupons.php
@@ -261,6 +261,7 @@ class Coupons
             self::validateCouponPeriod($settlement, $maturity);
             $frequency = FinancialValidations::validateFrequency($frequency);
             $basis = FinancialValidations::validateBasis($basis);
+            self::doNothing($basis);
         } catch (Exception $e) {
             return $e->getMessage();
         }
@@ -315,6 +316,7 @@ class Coupons
             self::validateCouponPeriod($settlement, $maturity);
             $frequency = FinancialValidations::validateFrequency($frequency);
             $basis = FinancialValidations::validateBasis($basis);
+            self::doNothing($basis);
         } catch (Exception $e) {
             return $e->getMessage();
         }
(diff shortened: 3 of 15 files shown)


src/PhpSpreadsheet/Calculation/Engineering/BesselI.php
```diff
@@ -111,9 +111,16 @@ private static function besselI1(float $x): float
...
+     *
+     * @var float
+     */
+    private static $zeroPointZero = 0.0;
+
     private static function besselI2(float $x, int $ord): float
     {
-        if ($x === 0.0) {
+        if ($x === self::$zeroPointZero) {
             return 0.0;
```
MarkBaker, line 124: Does it work with a constant as well as with a static variable?

src/PhpSpreadsheet/Calculation/Engineering/BesselJ.php
```diff
@@ -167,7 +167,7 @@ private static function besselj2b(float $ax, int $ord, float $x)
             if ($jsum === true) {
                 $sum += $bj;
             }
-            $jsum = !$jsum;
```
MarkBaker, line 170: Why does scrutinizer force us to add the overhead of a method call?

A more efficient alternative might be:
```php
$jsum = $jsum !== true;
```
though Scrutinixer might complain about that as well

src/PhpSpreadsheet/Calculation/Financial/Coupons.php
```diff
@@ -261,6 +261,7 @@ public static function COUPNCD(
             self::validateCouponPeriod($settlement, $maturity);
             $frequency = FinancialValidations::validateFrequency($frequency);
             $basis = FinancialValidations::validateBasis($basis);
```
MarkBaker, line 263: Would
```php
$basis = (bool) FinancialValidations::validateBasis($basis);
```
be better here? No method call overhead to cast it to a boolean

src/PhpSpreadsheet/Calculation/Engineering/BesselJ.php
oleibman: I will try that later today as well. I suspect it also won't work, but agree that would be a nicer solution.

src/PhpSpreadsheet/Calculation/Financial/Coupons.php
oleibman: This was a response to cases where Scrutinizer was correct (in a nagging it-really-doesn't matter way). It says $basis is unused, so the assignment is dead code. Casting won't help. And, if you remove the statement altogether, it will complain that the parameter $basis is unused. We are in a bind; Excel accepts the parameter, so we have to accept it whether or not our code needs it. I tried annotating the parameter, but one of our scanners (can't remember if it was phpcs or php-cs-fixer) objected. This seemed like a cheap way to keep everyone happy. I intend to revisit all of these eventually. For now, I'm just looking for a way to eliminate problems like this so that it is easier for me to run through the Scrutinizer issues to find complaints which are legitimate (there are some of those) without having to continually wade through things I've already decided are harmless.

src/PhpSpreadsheet/Calculation/Engineering/BesselI.php
oleibman: I didn't try with a constant. I'll set up a test later today.

oleibman: See PR #3123, which I will delete in a day or two. Using a constant did not eliminate Scrutinizer's objection.

src/PhpSpreadsheet/Calculation/Engineering/BesselJ.php
oleibman: See PR #3123, which I will delete in a day or two. YES ... this did eliminate Scrutinizer's objection. I will update the appropriate places in this PR accordingly, tonight or tomorrow.

MarkBaker: It's also perhaps easier to read than `$x = !$x` where it's easy to miss the `!` when just glancing at the code, and doesn't have the overhead of a function call

oleibman: Changed in latest push.

src/PhpSpreadsheet/Calculation/Engineering/BesselI.php
oleibman: Type-hinting constant also didn't help. Sticking with static.

Issue #3128, opened 2022-10-19
TypeError : abs(): Argument #1 ($num) must be of type int|float, string given


ma-gnesium:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Formatting the number with no run-time error

### What is the current behavior?
TypeError : abs(): Argument #1 ($num) must be of type int|float, string given
in PhpSpreadsheet/Style/NumberFormat/NumberFormatter.php:74

### What are the steps to reproduce?

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

As a unit test :
```php
<?php

use PhpOffice\PhpSpreadsheet\Style\NumberFormat;
use PHPUnit\Framework\TestCase;

class NumberFormatTest extends TestCase
{
    public function testSmallValueWithComplexFormat(): void
    {
        $result = NumberFormat::toFormattedString(1E-17, '0 000.0');

        self::assertEquals('0 000.0', $result);
    }
}
```
If this is an issue with reading a specific spreadsheet file, then it may be appropriate to provide a sample file that demonstrates the problem; but please keep it as small as possible, and sanitize any confidential information before uploading.

### What features do you think are causing the issue

- [ ] Reader
- [x] Writer
- [x] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

### Does an issue affect all spreadsheet file formats? If not, which formats are affected?
This issue is independent of the spreadsheet file format used

### Which versions of PhpSpreadsheet and PHP are affected?
PHPSpreadsheet version 1.25.2 (or master)
PHP 8.1.9

ma-gnesium:
I'm reconsidering this bug report.
The original intention was to format numbers using a thousands separator.

However, I've found that the correct syntax for this is to use the comma (,) character as a thousands separator, whatever the locale you're dealing with.

So the fix is simply to use the correct syntax: the TypeError is then not encountered and the result is correct.
This test passes:
```
    public function testSmallValueWithComplexFormat(): void
    {
        $result = NumberFormat::toFormattedString(1E-17, '0,000.0');

[...]

PR #3129, opened 2022-10-19
Fix a TypeError on complex number format with small enough number

ma-gnesium:
TypeError occurs with numbers that convert to scientific notation (ie small enough ones) when using number formats that require the complex algorithm of NumberFormatter.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [ ] Code style is respected

This fixes https://github.com/PHPOffice/PhpSpreadsheet/issues/3128


oleibman:
Your change works well for your use case. It does not seem to work well for large numbers, e.g. 1E20 (which are also not handled well by existing code). Might it be possible to to solve that problem as well, and add test cases for that and intermediate numbers (e.g. 1E5)?

oleibman:
I've experimented a little with large numbers. `number_format` seems to return some flaky results. I think it would be fairly straightforward with the BCMath extension, but that is not a requirement at the moment.

oleibman:
A function which might be usable can be found at https://stackoverflow.com/questions/66135441/converting-float-to-string-without-exponential-notation
You probably want to do something about trailing zeros.

ma-gnesium:
Thanks you for the feedback @oleibman.
Reconsidering the original bug report, I'm closing this PR because the effective fix from my point of vue is to use a correct format syntax, ie using comma as a thousands separator.

PR #3144, opened 2022-10-26
Test against php 8.2 in not experimental mode

sergiy-petrov:


MarkBaker:
Normally we wait until GA to switch the new release from experimental to non-experimental; so I won't take any steps to merge this until then.
We also normally add the new daily build as experimental at that stage, because that reflects work already merged in PHP core for the 8.3 release

Issue #3145, opened 2022-10-27
SimpleXMLElement::addChild(): unterminated entity reference &quot;Col&quot;)),1)


zappzerapp:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Read the xlsx file with multiple sheets and dropdown-values, also with formulas including quotations

### What is the current behavior?
```php
ErrorException
SimpleXMLElement::addChild(): unterminated entity reference &quot;Col&quot;)),1)
vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xlsx.php:938
```

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$file = __DIR__ . DIRECTORY_SEPARATOR . 'reproduce.xlsx';
$reader = \PhpOffice\PhpSpreadsheet\IOFactory::createReaderForFile($file);
$spreadsheet = $reader->load($file);
```

### What features do you think are causing the issue

- [X] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

### Does an issue affect all spreadsheet file formats? If not, which formats are affected?
XSLX

### Which versions of PhpSpreadsheet and PHP are affected?
- 1.25.2
- master

oleibman:
Closing. Should be fixed by PR #3146.

PR #3146, opened 2022-10-27
Using htmlspecialchars to fix issue #3145

zappzerapp:
This is:

```
- [X] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [X] Code style is respected

This prevents the SimpleXMLElement error "unterminated entity reference" when HTML entities are present in a worksheet #3145


oleibman:
I am a bit concerned that the use of htmlspecialchars here might break other things for which we don't yet have unit tests. I need to run some more tests, but the following also fixes your test case and seems safer.
```php
$childNode = $node->addChild('formula1');
if ($childNode !== null) { // null should never happen
    $childNode[0] = (string) $item->formula1->children(Namespaces::DATA_VALIDATIONS2)->f;
}
```
See https://stackoverflow.com/questions/552957/rationale-behind-simplexmlelements-handling-of-text-values-in-addchild-and-adda
And https://stackoverflow.com/questions/2970575/php-simplexmlelement-set-text

oleibman:
I am not sure I agree with PhpStan's analysis. I will open an issue there. In the meantime, I think it is safe to eliminate the string cast. Please give that a try. I have another variant in mind if that doesn't work.

zappzerapp:
I think we need your other variant 😄

oleibman:
I am sorry this is causing such a problem. I will have something ready for you later today.

oleibman:
Please re-code the erring line as:
```php
$childNode[0] = (string) $item->formula1->children(Namespaces::DATA_VALIDATIONS2)->f; // @phpstan-ignore-line
```
I have verified that (a) it works, and (b) Phpstan doesn't complain. I apologize for not doing the (b) part for my earlier suggestions. We don't like using the annotations, but sometimes you just can't get around it.

To your test, please add an assertion that the formula is as expected:
```php
self::assertSame(expected formula, $sheet->getCell('B2')->getDataValidation()->getFormula1());
```
Also please add at the very end of your test:
```php
```

zappzerapp:
Now everything looks good 🙂

oleibman:
Thank you for your contribution, and your patience.

Issue #3155, opened 2022-11-03
[Bug] Percentages stored as strings are not converted during formula calculations as they are in Excel


fdjohnston:
This is:

```
- [x] a bug report
- [ ] a feature request
- [x] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
When converting a spreadsheet to an array using `toArray()`, formulas that reference a cell that contains a percentage stored as a string should still calculate, like they do in Excel.

### What is the current behavior?
PHPSpreadsheet returns a #VALUE! error instead of converting the string to a float.

### What are the steps to reproduce?
```php
<?php

require __DIR__ . '/vendor/autoload.php';

//Create reader
$xlsxReader = \PhpOffice\PhpSpreadsheet\IOFactory::createReader('Xlsx');
$xlsxReader->setReadDataOnly(true);

//Load worksheet
$workbook = $xlsxReader->load('Book1.xlsx');
$worksheet = $workbook->getSheetByName('Sheet1');

//Convert to array
$data = $worksheet->toArray(null, true, false, false);

var_dump($data);
die();

```

Output:
```php
array(2) {
  [0]=>
  array(2) {
    [0]=>
    string(7) "#VALUE!"
    [1]=>
    NULL
  }
  [1]=>
  array(2) {
    [0]=>
    int(100)
    [1]=>
    string(2) "2%"
  }
}
```

Expected output:

```php
array(2) {
  [0]=>
    [0]=>
    float(2)
    [1]=>
    NULL
  }
  [1]=>
  array(2) {
    [0]=>
    int(100)
    [1]=>
    string(2) "2%"
  }
}
```

I have attached an XLSX test file.
[Book1.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/9932417/Book1.xlsx)

### What features do you think are causing the issue

- [ ] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [x] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

The `validateBinaryOperand()` function in `src/PhpSpreadsheet/Calculation/Calculation.php` already supports converting fractions stored as strings to numbers.  I think adding a function here to convert percentages stored as strings to numbers could solve this problem.

### Does an issue affect all spreadsheet file formats? If not, which formats are affected?
[...]

PR #3156, opened 2022-11-04
Convert percentages stored as strings to numerics in formula calculations

fdjohnston:
This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

This is a pull request to address https://github.com/PHPOffice/PhpSpreadsheet/issues/3155 

A new function was added that checks it a string contains a percentage, and if so, converts it to a numeric which can be used in formula calculations.


src/PhpSpreadsheet/Shared/StringHelper.php
```diff
@@ -560,6 +560,22 @@ public static function convertToNumberIfFraction(string &$operand): bool
...
 
+    /**
+     * Identify whether a string contains a percentage, and if so,
+     * convert it to a numeric.
+     *
+     * @param string $operand string value to test
+     */
+    public static function convertToNumberIfPercent(string &$operand): bool
+    {
+        if (strpos($operand, '%', -1) !== false) {
```
MarkBaker, line 571: Perhaps this check should be tightened up, because a string that contains only `'%'` will return a true here, even with no actual numeric value to convert. MS Excel will treat this as a string, and it won't be treated as 0.0%, so it will give a `#VALUE!` error if used in an arithmetic formula.

MarkBaker: And MS Excel also treats a string containing a leading percentage sign (e.g. `'%2.5'`) as a numeric percentage value in numeric calculations

MarkBaker: There may also be whitespace before a leadingg percentage sign, or after a trailing percentage sign

fdjohnston: Great feedback, thanks.  Was only considering my particular case.
Just pushed a  commit that switches to a regular expression.  I've also attached an Excel sheet that tests a bunch of different permutations and combinations.
Let me know what you think.
[Excel Test Cases.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/9946417/Excel.Test.Cases.xlsx)

Also just realized that I'm working on `master` on my fork instead of a new branch; sorry about that.

src/PhpSpreadsheet/Shared/StringHelper.php
```diff
@@ -560,6 +562,22 @@ public static function convertToNumberIfFraction(string &$operand): bool
...
 
+    /**
+     * Identify whether a string contains a percentage, and if so,
+     * convert it to a numeric.
+     *
+     * @param string $operand string value to test
+     */
+    public static function convertToNumberIfPercent(string &$operand): bool
+    {
+        if (preg_match(self::STRING_REGEXP_PERCENT, $operand, $match)) {
```
MarkBaker, line 573: There's a couple of little tricks that we can do with the regexp to extract the numeric value directly, rather than doing the `str_replace()` call.
```
~^(?: *-? *\% *(?<PrefixedValue>[0-9]+\.?[0-9*]*) *| *\-? *(?<PostfixedValue>[0-9]+\.?[0-9]*) *\% *)$~
```
To avoid capture of the complete string, I've set that as a non-capture group, because we're only interested in whether the `%` character appears, we don't need to actually capture it or any padding around it; and then I've set capture groups for the actual numeric part of the string. I've also named these groups (`PrefixedValue` and `PostfixedValue`) so that they can be referenced by name in the `$matches` array.

Then, if a match is identified, we can extract the numeric part directly by name, whether it's prefixed by the `%` or postfixed.
```php
$poperand = (float) ($match['PostfixedValue'] ?? $match['PrefixedValue']) / 100.0;
```
The null coalescence identifies whether the number is prefixed or postfixed (using the named groups from `$matches`); and then does the division by 100 and casts the result to a float. There's no function call overhead; just PHP operations, so it should be more efficient if it's being called in a loop thatr checks a lot of cell values.

fdjohnston: Great trick with the regexp.  I agree that eliminating the `str_replace` would be a big improvement.
Unfortunately the changes to the regexp break the negative percentage cases.  I'll see if I can run with your idea and get it working for negative values.

MarkBaker: Ah, yes! I was being a bit to quick and casual with the regexp changes. If the capture group is expanded to encompass the sign as well, then we're potentially also capturing whitespace too, which would break other things.

One option would be to have additional pre/postfix capture groups for the sign, and a match there would trigger a `* -1` operation
```
~^(?: *(?<PrefixedNegative>-?) *\% *(?<PrefixedValue>[0-9]+\.?[0-9*]*) *| *(?<PrefixedNegative>\-?) *(?<PostfixedValue>[0-9]+\.?[0-9]*) *\% *)$~
```
and then
```php
$sign = $match['PostfixedNegative'] ?? $match['PrefixedNegative'];
$poperand *= ($sign === '-') ? -1.0 : 1.0;
```
So we're still using only operations, and no function calls.

Just to be bloody minded and awkward, MS Excel recognises a `%` prefix with the sign before or after the `%`, so `%-2.5` and `-%2.5` are both recognised as numeric values. Fortunately with a postfix `%`, the sign must be before the numeric part.
And, of course, the sign could be a `+` as well as a `-`; and the numeric part could be in scientific format, `-2.5E-4%`.

But if you get the basic negative working, I'll merge and then do some more tweaking myself to handle a few of those edge cases; and refactor it all into a separate helper.

I did say right from the start that this was a lot more complicated once you start to look at it in detail. ;-)

fdjohnston: You did warn me! I can't imagine what the code for Excel's formula parser must look like.
I'll take a crack at improving the regexp like you suggest.  

Should we be concerned about any of the other spreadsheet applications?  I'm focused entirely on Excel because it was the root of my issue, but I suppose it's worth considering how Open Office or Libre Office works?  Won't that be a nightmare if they handle these various combinations of `%`, `+`, and `-` differently....  I don't have either installed at the moment but I'll download a copy tonight and see how they behave.

fdjohnston: Okay sounds good.  I was able to get it working for combinations of `+`, `-`, and scientific notation by expanding on the approach you suggested.  
I broke out all the different pieces of the final calculation because it was getting a bit difficult to follow as single line, owing to all the null coalescing checks.  We could simplify it by moving all those checks inline.  
Note that, as I mention in the commit, I did have to add the `PREG_UNMATCHED_AS_NULL` flag to the `preg_match`, otherwise I was getting '' back when a capture group wasn't found.
Let me know what you think.

fdjohnston: Updated my Excel sheet with various permutations.
[Excel Test Cases.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/9963751/Excel.Test.Cases.xlsx)

MarkBaker:
Of course, you don't need to use the `pow()` function at all, PHP has a power operator that can be used instead

But I'm sure that casting to float will convert a string like `1.23e-4` to a float value without needing that additional complexity, and that the regexp can be simplified to extract just sign and numeric value rather than sign, mantissa, exponent sign and exponent.
Also, `E` and `e` can both be used for scientific format, so the regexp needs to be case-insensitive with the `i` switch

fdjohnston:
Yep, you're right.  Thanks for your patience on this PR.  Really appreciate your taking the time to point this stuff out.  
I think this most recent push should take care of everything.

MarkBaker:
Thank you for your work on this PR; it is appreciated.
I know that I migfht have been asking you to do more than you needed for your own use case, but that makes it more beneficial to other developers that may encounter the same issue

PR #3164, opened 2022-11-10
Refactoring for checks on strings containing formatted numeric values

MarkBaker:
This is:

```
- [ ] a bugfix
- [ ] a new feature
- [X] refactoring
- [ ] additional unit tests
```

Checklist:

- [X] Code style is respected

Refactoring for checks on strings containing formatted numeric values (e.g. percentages and fractions) when used in mathematical operations in the Calculation Engine

This is a refactoring exercise to make extract the logic into its own class, making it easier to manipulate formatted string values into their numeric value, and adding handling for new formatting methods in the future


oleibman:
@MarkBaker I see you've finally eliminated Php7.3. So why does Github still show a Php7.3 check and mark it required even though it will never run (and we therefore don't see "all required statuses have passed")? And, in a similar vein, should the Php8.2 test, which is no longer experimental, be marked as required?

MarkBaker:
> @MarkBaker I see you've finally eliminated Php7.3. So why does Github still show a Php7.3 check and mark it required even though it will never run (and we therefore don't see "all required statuses have passed")? And, in a similar vein, should the Php8.2 test, which is no longer experimental, be marked as required?

All the `7.3` references are removed from the github side: I think that `7.3` reference is from scrutinizer. I don't have access to scrutinizer itself, that would be @PowerKiKi to check on that side.

The `8.2` tests are set for no failure allowed. Currently, the `nightly` build (set for allow failure) is still against the `8.2-dev` branch of PHP, so the nightly is a duplication of 8.2; but that will change in two weeks when `8.2` goes to GA, and the `nightly` will switch to the planned `8.3` release.

PowerKiKi:
@MarkBaker I gave you access to Scrutinizer. 

@oleibman I cannot see the same thing you saw anymore, but I am pretty sure that GitHub was still waiting for a CI to pass for PHP 7.3 because it previously existed. That's how GitHub Actions works, it expect the previous check to still exist and pass. It's a bit cumbersome when we remove checks, but it should auto-correct itself once we merge something with the missing check. I'll see what happens with #3223 ...

oleibman:
@PowerKiKi Thank you for the explanation. I have confirmed with a new PR, and merging code with an existing PR, that both function correctly now.

Issue #3165, opened 2022-11-11
[FEATURE] Convert Currencies stored as text during formula calculations


fdjohnston:
This is:

```
- [ ] a bug report
- [x] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

Hi, it's me again.  Just ran into a situation similar to the percentage case we just worked through, but this time with currencies.  My specific case from the Formula Parser logs is:
```
Evaluating 1.07 + "$0.00"
Evaluation Result is a a #VALUE! error
```

My initial thoughts are we could do something similar to what we just did for percentages, but for currencies.  I did some playing around in Excel and it seems that it is much more restrictive about what is considered to be a currency than it is with percentages, though I wonder how localization settings are playing into this.  I expect one would have to handle cases with the currency symbol before or after the value, probably very similar to what we ran into with percentages.  I haven't thought this all the way through, but I wonder if we could even get away with turning the symbol into a capture group in the regexp and depending on if it's a currency symbol or a `%` have the `convertToNumberIfPercent()` function omit the division by 100.  There's no way it could be that simple, but I wonder if that is a direction that bears exploring.  

I going to assume that I'm underestimating the complexity of this like I did with percentages, but I thought it was worth pointing out.  

As before, I'd be happy to contribute a PR if this is something the maintainers think is worth fixing, or would this just be opening a can of worms?

### What features do you think are causing the issue

- [ ] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [x] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

MarkBaker:
We would certainly take a similar approach (have a look at [PR #3164](https://github.com/PHPOffice/PhpSpreadsheet/pull/3164) to see how I've extracted that logic into a Calculation Engine Helper).

Excel is a lot stricter about what does and doesn't qualify as a currency; and that is very much locale-based: I'm based in Europe, and not in the United States, so....
![image](https://user-images.githubusercontent.com/770298/201381976-4bd9198b-2a93-4bd3-b646-547ff93553e2.png)!![image](https://user-images.githubusercontent.com/770298/201382303-3cbf207d-1317-45c7-b137-d7b415e2f81c.png)
But it doesn't take locale decimal and thousands separators into consideration
[...]

PR #3189, opened 2022-11-18
Handle Currencies Stored as Strings in Formulas

fdjohnston:
Added a function to the new `FormattedNumber` helper that will use the currency code pulled from `localeconv` by `StringHelper::getCurrencyCode()`.  The currency code is `preg_quoted` and then dropped into a regexp that is modelled on the expression developed for `convertToNumberIfPercent`.  This will allow locale independent operation. Unfortunately `localeconv` only provides information about standard currency formats and not accounting formats.  This means that we can't say for sure whether or not a locale has an accounting format where the currency symbol shows up in a non-standard position (eg. left justified for some countries instead of appearing directly in front of the value).  The regexp should handle these cases.

The primary issue with this approach is that the regexp will incorrectly match invalid currency formats for some locals as it checks for the symbol before and after the value.

A possible improvement would be to pull `p_cs_precedes` and `n_cs_precedes` from `localeconv` and using them to determine if the currency symbol should appear before or after the value for the current locale.  Since white-space is ignored by the regexp, accounting formats should still work, as long as the accounting format doesn't involve moving the symbol to the opposite side of the value.

This is:

```
- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected

See issue #3165

### DIFF
--- a/src/PhpSpreadsheet/Calculation/Engine/FormattedNumber.php
+++ b/src/PhpSpreadsheet/Calculation/Engine/FormattedNumber.php
@@ -3,6 +3,7 @@
 namespace PhpOffice\PhpSpreadsheet\Calculation\Engine;
 
 use PhpOffice\PhpSpreadsheet\Calculation\Calculation;
+use PhpOffice\PhpSpreadsheet\Shared\StringHelper;
 
 class FormattedNumber
 {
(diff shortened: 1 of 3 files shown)


MarkBaker:
I've made a couple of small changes
 - Use of the unicode flag for the regexp, for currency codes that aren't a part of the standard ASCII character set
 - Some additional unit tests for positive sign
 - Handling of currency/accounting strings with a thousands separator (not locale-aware)

Thanks for your work on this

fdjohnston:
Great improvements; thank you.

Happy to contribute to such a great project, and may I say I've really enjoyed going back and forth with you on this.  I have definitely learned some new tricks with regular expressions.

If you ever decide to make the solution Locale aware or explore the PHP Intl option you mentioned over in #3165, and if you'd like a hand, I would be happy to try to help out.

PR #3262, opened 2022-12-26
Add Additional Method to XIRR if Newton-Raphson Does Not Converge

### LINKED ISSUE #689: XIRR returns Non Numeric value Encountered
XIRR returning Non Numeric Value Encountered thought the inputs passed are correct. Below is the input passed. 

array:2 [▼
  0 => -1000000.706
  1 => 947003.58
]
array:2 [▼
  0 => "2018-09-05"
  1 => "2018-09-26"
]

The Expected XIRR Return should be **-0.6118824173**

oleibman:
Fix #689. XIRR is calculated by making guesses which are hopefully better with each iteration. It is not guaranteed to succeed for Excel, PhpSpreadsheet, or any other implementation. PhpSpreadsheet uses the Newton-Raphson method for its guesses. So does Python package xirr (https://github.com/tarioch/xirr/), but, if Newton-Raphson fails to converge, Python tries Brent's method as an alternative. Two sets of non-converging data are noted in 689. For both, a solution does converge in Excel. For the first of the problems, a solution converges in Python with Newton-Raphson; but, for the second, a solution converges which requires Brent. For the Java package https://github.com/RayDeCampo/java-xirr on which Python was based, and which uses only Newton-Raphson, a solution converges for the first, and does not converge for the second.

To try to match the good results of the others, I added an alternate algorithm if Newton-Raphson fails. Brent's algorithm seems difficult to implement to me. I might have gone there regardless, but I first tried a slightly simpler alternative, bisection. This solved the problem for both of the cases in 689. Perhaps someone will one day report a problem that doesn't converge for Newton-Raphson or bisection, but does for Brent. We can review this decision then.

The new code causes 3 changes in the unit test. In all 3 tests, Excel and PhpSpreadsheet had not converged, but Python and/or Java had. I now believe that Python/Java is correct in those cases, and Excel is not. The new code aligns PhpSpreadsheet with Python/Java for those tests. It is, of course, impossible to know when Excel's implementation doesn't converge, so we aren't guaranteed to match its results in those hopefully rare situations.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [x] additional unit tests
```

Checklist:

- [x] Code style is respected


oleibman:
Not concerned about Scrutinizer "complexity" complaint.

PR #3287, opened 2023-01-09
Minor Fix for AND/OR/XOR

oleibman:
These 3 fall into the set of functions where Excel treats string literals differently depending on whether they are passed to the function directly or as a cell reference. PhpSpreadsheet is updated to try to duplicate that logic. New tests are added. Some existing test results had to change as a result of this code change.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected


oleibman:
I don't care about, and have no intention to act on, Scrutinizer's "complexity" message.

src/PhpSpreadsheet/Calculation/Logical/Operations.php
```diff
@@ -116,22 +85,7 @@ public static function logicalOr(...$args)
...
-            return $value !== null || (is_string($value) && trim($value) == '');
-        });
-
-        $returnValue = self::countTrueValues($args);
-        if (is_string($returnValue)) {
-            return $returnValue;
-        }
-
-        return $returnValue % 2 == 1;
+        return self::countTrueValues($args, 'xor');
```
MarkBaker, line 88: I might be tempted to have three wrapper methods, `countTrueValuesForAnd()`, `countTrueValuesForOr()`, `countTrueValuesForXor()` that call the basic `countTrueValues()` method, then handle the specific 0 count and comparison logic, rather than passing a string to the one method and then using if tests

oleibman: I changed to eliminate the ifs, not precisely the way you wanted, but addressing your concern, I hope.

PR #3306, opened 2023-01-20
Xls Reader Vertical Break and Writer Page Order

### LINKED ISSUE #3305: Row and column integer error when calling IOFactory::load
This is a **bug** report.

The file should load to be processed further.

An error is thrown when load is called.

> Error 0 on line 49 in \PhpSpreadsheet\Cell\CellAddress.php -> Row and Column Ids must be positive integer values

This is not happening with every file but files that used to load without error are now causing the error shown above. So far the files that are failing are all XLS. There has been limited use with XLSX and, so far, no failures with XLSX. I believe it is related to the file structure or something within the file itself as I can copy and paste the rows into a newly created file and it will load and process correctly.
```php
<?php

use PhpOffice\PhpSpreadsheet\Reader\Xlsx;
$testAgainstFormats = [\PhpOffice\PhpSpreadsheet\IOFactory::READER_XLS,
                                     \PhpOffice\PhpSpreadsheet\IOFactory::READER_XLSX];
$spreadsheet = \PhpOffice\PhpSpreadsheet\IOFactory::load('FailTestShort.xls',0,$testAgainstFormats);  // <<-- file attached - has only 2 rows

- [X] Reader (and/or possibly IOFactory::load

Right now it only seems to be affecting XLS format files.

PHP 8.2.1 and a fresh copy of PhpSpreadsheet that was pulled today (19 Jan 2023).

[FailTestShort.xls](https://github.com/PHPOffice/PhpSpreadsheet/files/10460781/FailTestShort.xls)

### PR DESCRIPTION
**Pull request by oleibman**
Fix #3305. Xls Reader can set vertical break specifying row 0, causing an exception. It is doubtful that Excel needs a row for a vertical break; code is changed to use row 1 if the input file specifies row 0 (or lower). Code in question has not been exercised in unit test suite. Similarly, code to set horizontal break (which probably does not have a bug) is not exercised in test suite. Finally, page order in Writer incorrectly uses value in opposite way that Reader does. A new sample is added to illustrate that these are all handled correctly; it is easier to verify this by visually comparing the source spreadsheet and the copy made from it. A unit test is also added for the same spreadsheet to formally assert that the 3 properties in question are both read and written correctly.

This is:

```
- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
```

Checklist:

- [x] Code style is respected


oleibman:
@MarkBaker Since other comments indicate you are ready for a new release, this probably ought to be part of it (corrects a problem introduced in 1.26 and one that may always have been wrong). I opened it so recently that I just want to make sure you have the opportunity to review before I merge it (or before you do).

MarkBaker:
> @MarkBaker Since other comments indicate you are ready for a new release, this probably ought to be part of it (corrects a problem introduced in 1.26 and one that may always have been wrong). I opened it so recently that I just want to make sure you have the opportunity to review before I merge it (or before you do).

I'd like to get a new release out this weekend if I can, although there's a couple of issues that I'm currently trying to resolve so that they can also be included in that release, Hopefully I'll get them fixed over the weekend.

PR #3329, opened 2023-01-30
Include previous exception in \PhpOffice\PhpSpreadsheet\Cell\Cell::getCalculatedValue()

mpdude:
This is a bugfix.

Checklist:

- [ ] Code style is respected

In my production logs, I regularly see exceptions that originate here:

https://github.com/PHPOffice/PhpSpreadsheet/blob/c2f806bae7f77507e1a6492d35b969387f547661/src/PhpSpreadsheet/Cell/Cell.php#L382-L391

To really follow up on this, I'd need to know more about the underlying exception that caused the problem. This exception is lost in the exception-handling code.

By including the exception as the "previous" one, it is possible to include also the initial problem spot.


MarkBaker:
Before you start adding new ways to retrieve details of formula evaluation errors in the Calculation Engine, you might want to take a look at the evaluation debugging tools that are already built-in 

https://gist.github.com/MarkBaker/3ea8664c73bfe6af5bc6f437803589f2

mpdude:
Will do. 

But keeping the original exception when re-throwing a new one is never wrong, is it?

mpdude:
Here is code that reproduces the exception I am seeing:

```php
<?php

use PhpOffice\PhpSpreadsheet\Cell\DataType;
use PhpOffice\PhpSpreadsheet\Shared\Date;

require_once 'vendor/autoload.php';

$reader = IOFactory::createReaderForFile('test.xlsx');
$spreadsheet = $reader->load('test.xlsx');
$sheet = $spreadsheet->getSheet(0);
$cell = $sheet->getCell('A1');
$cell->setDataType(DataType::TYPE_FORMULA);
$cell->setValue('= 44 7510557347');
var_dump(Date::isDateTime($cell));
```

In practice, I do not put this value into the cell myself, but it is in a cell of a spreadsheet uploaded by users, which fails when I try to read data from it.

MarkBaker:
Interesting! I'm surprised the MS Excel itself doesn't complain about that.

But it probably means that I should put some exception handling in the `isDateTime()` method to handle an exception in the calculation of the cell value.

MarkBaker (approved):

MarkBaker:
Bedankt

mpdude:
Graag gedaan

PR #3341, opened 2023-02-03
Fix PHP8.2 str_split function returns empty arrays for empty strings

HuongNV13:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

In PHP 8.2, the str_split function will returns empty arrays for empty strings.
See: https://php.watch/versions/8.2/str_split-empty-string-empty-array

We can use mb_str_split() instead


src/PhpSpreadsheet/Reader/Csv/Delimiter.php
```diff
@@ -60,7 +60,7 @@ protected function countPotentialDelimiters(): void
 
     protected function countDelimiterValues(string $line, array $delimiterKeys): void
     {
-        $splitString = str_split($line, 1);
+        $splitString = mb_str_split($line, 1);
         if (is_array($splitString)) {
```
oleibman, line 64: Unlike str_split, mb_str_split is documented to always return an array. So the following is_array check is now unneeded. Scrutinizer complains about its continued presence, and I am frankly surprised that Phpstan doesn't do likewise. Can you please delete the if statement and its closing bracket, leaving the rest as-is.

oleibman (dismissed):
I don't believe we are actually subject to any failures as a result of the Php change, but I think mb_substr_split is probably what should be using anyhow (in addition to other mb_ functions in lieu of the straight string equivalents). Can you please add some null string tests to `tests\data\Calculation\Engineering\Hex2Dec` (`[0, '""'],`), `tests\data\Calculation\Engineering\Oct2Dec` (`['0', '""'],`) and `tests\data\Calculation\MathTrig\Arabic` (`[0, ''],`), just to demonstrate that they work as in Excel.

ConnectGrid:
The Mbstring extension is not bundled with every distribution of PHP by default.

If the issue is splitting on an empty string, would it not be better to check for that case prior to calling `str_split()` (or `mb_str_split()`), or always ensure that an array is evaluated?

```
foreach (mb_str_split((array)$value) as $char) { ... }

// or

foreach (mb_str_split($value ?: []) as $char) { ... }
```

Would there ever be a false-y `$value` that would need to be acted upon by this function?

MarkBaker:
> The Mbstring extension is not bundled with every distribution of PHP by default.

It's tagged as a required dependency in `composer.json`; but we can't control if devs do a composer install with `--ignore-platform-reqs`; or pulls a phar from one of those 3rd party sites that bundles dependencies, but doesn't do any PHP requirements checking

I'd also like to see some tests to identify any input $value where the change to `str_split()` could actually cause an issue in those methods

oleibman:
I don't think it would be practical to use PhpSpreadsheet if mbstring is not enabled. Mbstring functions are directly called 45 times in 38 different modules. In addition, StringHelper methods, many of which use mbstring functions, are invoked in 186 different modules.

MarkBaker:
> I don't think it would be practical to use PhpSpreadsheet if mbstring is not enabled. Mbstring functions are directly called 45 times in 38 different modules. In addition, StringHelper methods, many of which use mbstring functions, are invoked in 186 different modules.

It would be a very limited subset of functionality, but not impossible: for character conversion, we try to use iconv where possible, with mb_ functions as a fallback

We also have our own `mbStrSplit` function implemented in the StringHelper class

src/PhpSpreadsheet/Reader/Csv/Delimiter.php
HuongNV13: Thanks, @oleibman 
I have added extra tests and removed the needed condition.

MarkBaker:
I see that you've added some unit tests; but I don't see any tests that demonstrate there is a bug unless this change to use `mb_str_split` instead of `str_split` is made. Perhaps there are cases where it might be a problem in identifying the csv separator; but I can't find any situation where using `str_split` is a problem in the Excel function implementations.

oleibman:
I am going to take another look at this. PR #4344 is a case where problems were caused due to parsing by bytes (like str_split) rather than parsing by characters (like mb_str_split). I will see if resolving the merge conflicts here results in a usable PR, and then decide how to proceed.

oleibman:

PR #3345, opened 2023-02-04
Permit Max Column for Row Breaks

### LINKED ISSUE #3143: Generated Excel file loses page breaks when prininting area does not start in the first column
This is:

```
- [X] a bug report
- [ ] a feature request
```

The Printing configuration should be kept after saving the document.

When the printing area does not start on the first column, the page breaks disappears or ar return to automatic ones.

This simple code allows to test the case, it just opens the file and save it without modification. 
The [area_ko.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/9870198/area_ko.xlsx) gives the following result (left is the original file, right is after opening and saving it with PhpSpreadsheet.)
image

The [area.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/9870241/area.xlsx) file is the same case just with the printing area set to start on the first column, and works as expected 
image

```php
<?php

// Load the file
$xlsxReader = \PhpOffice\PhpSpreadsheet\IOFactory::createReader('Xlsx');
$workbook = $xlsxReader->load('area.xlsx');    
//$workbook = $xlsxReader->load('area_ko.xlsx');    

// Save the file
$writer = new  \PhpOffice\PhpSpreadsheet\Writer\Xlsx($workbook);
$writer->save('result.xlsx');
```

- [ ] Reader
- [X] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

oleibman:
Fix #3143. Page break was dropped. Difference between good and bad was the use of attribute `max` in `brk` tag in the good spreadsheet. However, `max` was *not* required in a similar spreadsheet. So the reason for the problem isn't completely explained, although the likeliest explanation seems to be an Excel bug. Nevertheless, it can't really hurt to capture the `max` value on read (if present) and generate it on write. This resolves the issue. User is also permitted to specify max column when setting a row break programatically. Documentation is updated.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
I have tested further, using PhpSpreadsheet code to emulate the problem spreadsheet, and I am quite certain that the *only* difference between a spreadsheet which appears as expected and one which doesn't is the `brk` `max` attribute. This convinces me that Excel is acting strangely when faced with a row page break inside a defined print area. Bug? Feature? Who knows? At any rate, this result tells me that we should proceed with this PR. It does not convince me that it is always safe to add the `max` attribute, nor that the value should always be 16,383, but I can at least document when this should be done and suggest that as a reasonable value to use. I will update this PR with a documentation change and additional tests.

oleibman:
BTW, LibreOffice handles the printing correctly whether or not `max` is specified on `brk`.

PR #3347, opened 2023-02-04
Check $strictNullComparison outside of loops

ConnectGrid:
Might be considered a micro-optimization, but for large sheets, it could make a difference in performance. Only check the `$strictNullComparison` condition once, not for every single cell loop.

This is:

- [ ] a bugfix
- [ ] a new feature
- [X] refactoring
- [ ] additional unit tests

Checklist:

- [X] Code style is respected

Instantiating Worksheets from an array are currently less than optimal due to checking a boolean condition on every single cell loop.


src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -2922,23 +2922,30 @@ public function fromArray(array $source, $nullValue = null, $startCell = 'A1', $
...
-            $currentColumn = $startColumn;
-            foreach ($rowData as $cellValue) {
-                if ($strictNullComparison) {
+        if ($strictNullComparison) {
+            foreach ($source as $rowData) {
+                $currentColumn = $startColumn;
+                foreach ($rowData as $cellValue) {
                     if ($cellValue !== $nullValue) {
                         // Set cell value
                         $this->getCell($currentColumn . $startRow)->setValue($cellValue);
```
MarkBaker, line 2931: I've not run any timing tests on your code yet; but when it comes to micro-optimisations, then interpolation is more efficient than the concatenation that we use here:
```php
$this->getCell("{$currentColumn}{$startRow}")->setValue($cellValue);
```

Try running
```php
$callStartTime = microtime(true);

for ($row = 1; $row <= 65535; ++$row) {
    for ($column = 'A'; $column !== 'IV'; ++$column) {
        $cellAddress = $column . $row;
    }
}

$callEndTime = microtime(true);
$concatenationCallTime = $callEndTime - $callStartTime;

echo PHP_EOL;
echo 'Call time for concatenation was ' , sprintf('%.4f', $concatenationCallTime) , ' seconds' , PHP_EOL;


$callStartTime = microtime(true);

for ($row = 1; $row <= 65535; ++$row) {
    for ($column = 'A'; $column !== 'IV'; ++$column) {
        $cellAddress = "{$column}{$row}";
    }
}

$callEndTime = microtime(true);
$interpolationCallTime = $callEndTime - $callStartTime;

echo PHP_EOL;
echo 'Call time for interpolation was ' , sprintf('%.4f', $interpolationCallTime) , ' seconds' , PHP_EOL;
```
to see the difference

And I'm of the growing opinion that interpolation is actually more readable

oleibman: Ticket amazingly has no merge conflicts after all this time. I have run Mark's test program, and it now seems to favor concatenation over interpolation. So I will let this proceed as-is, assuming it still passes all tests, of course.

oleibman:
Nothing more to discuss.

oleibman:

PR #3391, opened 2023-02-22
Xls Writer Parser Handle Boolean Literals as Function Arguments

### LINKED ISSUE #3369: Xls Writer Cannot Parse TRUE When Supplied as a Function Argument
This is:

```
- [x] a bug report
- [ ] a feature request
```

A function argument of TRUE (and presumably FALSE) can be parsed by the Xls Writer so that the formula is reproduced faithfully when the spreadsheet is saved as Xls.

Parser throws an Exception at the end of function `fact`. The writer still writes the cell, but with its calculated value rather than its formula. I can see that another regexp or two needs to be tested for this, but I have not yet figured out what should happen if the new regexes are matched. Note that TRUE is parsed correctly as just a cell value (cell A1), and that the `TRUE()` function is parsed correctly when supplied as an argument (cell A4). However, both cells A2 (TRUE is first argument) and A3 (TRUE is last argument) are written out as a value rather than a formula.

```php
<?php

use PhpOffice\PhpSpreadsheet\Writer\Xls;

$sheet->getCell('A1')->setValue(true);
$sheet->getCell('A2')->setValue('=AND(TRUE)');
$sheet->getCell('A3')->setValue('=AND(TRUE, 0)');
$sheet->getCell('A4')->setValue('=AND(TRUE())');
$writer = new Xls($spreadsheet);
$outfile = 'formulaerr2.xls';
$writer->save($outfile);
echo "Saved $outfile\n";
```

- [ ] Reader
- [x] Writer, in particular Writer/Xls/Parser
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

Only Xls is affected.

All.

oleibman:
Fix #3369. The parser had failed to account for `TRUE` and `FALSE` when supplied as arguments to a function.

I had hoped to be able to do something about its inability to handle defined names as well. I failed. I think another section might need to be added to the Writer output which specifies the defined names. I haven't yet located any suitable documentation.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Not concerned with Scrutinizer "complexity" complaint.

PR #3410, opened 2023-02-28
Zipstream 3.0

oleibman:
Zipstream's newest production release requires Php8.1. This is not necessarily a problem for us as we are locked to an earlier release. However, psr/simple-cache became a real mess for us as people wanted to install PhpSpreadsheet alongside other products which required a version of simple-cache that could not run in some of the versions of Php which we support, and I would like to avoid having to go through that again. This ticket is a contingency to put us ahead of the curve in case such a problem should arise with Zipstream. There is reason to believe that Zipstream is not so widely used as simple-cache, so we do not need to rush this change in. However, should that belief prove incorrect, the change should be ready to go when needed.

This is:

- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
I have rebased several times without any hint of a problem. I now think it is safe to merge this, so am taking it out of draft status. I would like to delay merging it till after the next official release, which I believe will be soon.

MarkBaker:
> I would like to delay merging it till after the next official release, which I believe will be soon.

Early in the coming week is my plan

oleibman:
I've been expecting Scrutinizer to complain about this for a while. We execute only Zipstream2 or Zipstream3, leaving the other as entirely uncovered code. Scrutinizer therefore reports that coverage has decreased. Since the coverage test is locked at Zipstream 2, we could tell phpunit to ignore coverage for Zipstream3. That will leave us a problem some time in the future when the locked version switches. Or we could tell it to ignore coverage for both, future-proofing us, but making the coverage report slightly less accurate. Or we can just accept the fact that this change reduces coverage (by something like 1 statement out of 30,000+); that's the option I'm going for. (BTW, I believe the reason Scrutinizer hasn't complained till now is that the coverage report hasn't finished early enough for Scrutinizer to process it, but there was a backlog for Scrutinizer tonight, allowing the coverage report to finish first.)

oleibman:
I plan to merge this next Monday or Tuesday unless there is an official release between now and then, in which case I'll wait another week.

PR #3416, opened 2023-03-01
Xlsx Column Autosize Approximate for CJK

### LINKED ISSUE #3405: 中文自动列宽设置后无效
This is:

```
- [ ] a bug report
- [ ] a feature request
```

```php
<?php
           
 $dataSheet->getColumnDimension("G")->setAutoSize(true); 

如果某一列是CJK 其中的一种，这样的设置方式无效

 version     "phpoffice/phpspreadsheet": "^1.19",





- [ ] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

oleibman:
Fix #3405. Autosize is definitely not working well with CJK characters (column is not wide enough). User reports a workaround using `mb_strwidth` to calculate and set the column width. PhpSpreadsheet uses `mb_strlen` for width calculations. Change it to use mb_strwidth instead. For non-CJK strings, the results will be identical (and there are already unit tests on such strings which assert the expected results, and these tests did not need to change). For CJK strings, the results will be wider. The string I'm using to test comes from the issue. It currently results in a column width of 30.564. When I open the resulting sheet in Excel and auto-fit the column width, the width winds up as 43.00. So, as long as the computed width exceeds 43.00, the spreadsheet will show the full cell. With the new calculation, the computed width is 55.2722, satisfying our condition. This is wider than expected, but that is generally true for this type of computation. For example, for 'abcdefghijklmnopqrstuvwxyz', the computed width (before and after this change) is 31.7065, but Excel auto-fit actually uses 24.73.

"Exact width calculation" works well for fonts that support CJK, even without this PR. Since none of the fonts included in Shared/Font support CJK, this requires the use of the Shared/Font::setTrueTypeFontPath and Shard/Font::setExtraFontArray methods.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
I was wrong to suggest a problem with  exact width calculation" - I thought the font I was using supported the CJK characters, but it doesn't - testing with a font with verified support for the characters worked just fine. Description has been updated.

PR #3443, opened 2023-03-09
Convert cell value to numbeic (float) in processDomElementDataFormat

PouriaSeyfi:
If we want to use data-format attribute in html, we have to cast cell value to float (number). because format not applied to string values. We should know excel put Leading apostrophes " ' " to numeric string values. Leading apostrophes force excel to treat the cell's contents as a text value.

### DIFF
--- a/src/PhpSpreadsheet/Reader/Html.php
+++ b/src/PhpSpreadsheet/Reader/Html.php
@@ -560,6 +560,8 @@ class Html extends BaseReader
     private function processDomElementDataFormat(Worksheet $sheet, int $row, string $column, array $attributeArray): void
     {
         if (isset($attributeArray['data-format'])) {
+            $value = $sheet->getCell($column . $row)->getValue();
+            $sheet->setCellValue($column . $row, (float) $value);
             $sheet->getStyle($column . $row)->getNumberFormat()->setFormatCode($attributeArray['data-format']);
         }
     }


oleibman:
I think you have identified a problem, but not a proper solution. Here's a basic program to parse html.
```php
$spreadsheet = $reader->loadFromString($html);
var_dump($sheet->getCell('A1')->getValue());
var_dump($sheet->getCell('A1')->getFormattedValue());
var_dump($sheet->getStyle('A1')->getNumberFormat()->getFormatCode());
```
Now, let's look at some possible input:
```html
<table>
    <tr>
         <td data-format="#.000">3</td>
    </tr>
</table>
```
This works correctly with or without your change - `getFormattedValue` returns `3.000`.
Here's different html:
```html
<table>
    <tr>
         <td data-format="#.000">x</td>
    </tr>
</table>
```
This works correctly, or at least the same way as Excel, before your change - `getFormattedValue` returns `x`. I believe that your change breaks this behavior.
Here is still different html, taken from HtmlTest testCanApplyInlineDataFormat.
```html
<table>
    <tr>
         <td data-format="mmm-yy">2019-02-02 12:34:00</td>
    </tr>
</table>
```
Here, the existing test does not go far enough. It verifies that the format is set correctly, but doesn't check what `getFormattedValue` returns, which, in this case, is `2019-02-02 12:34:00`. The Html load has not treated the value as a date, and possibly should (I have no idea yet how difficult that would be). Your change doesn't do much better; casting to float converts the string to 2019.0, losing the month, day, and time.

Of course, date/time strings may not be the only problem. I think you have a specific problem in mind. What is the specific problem that prompted you to submit this PR?

PouriaSeyfi:
Yes thank you for your good explanation. you are right I think I had this problem in older versions.
now it works without problem. I identified a problem and I will create new PR

Issue #3444, opened 2023-03-09
How to specify data type in html using attribute html 


PouriaSeyfi:
During HTML to Excel conversion, it is sometimes necessary to know what date type should be used :  

```html
                <table>
                    <tr>
                        <!-- boolean data type -->
                        <td data-type="b">1</td>
                        <!-- string data type -->
                        <td data-type="s">12345678987654321</td>
                        <!-- formula data type -->
                        <td data-type="f">=CONCAT("TEXT A ","TEXT B")</td>
                    </tr>
                </table>
````

PouriaSeyfi:
Please review this PR 
@MarkBaker

oleibman:
Resolved by PR #3445.

PR #3445, opened 2023-03-09
Specify data type in html tags using attributes

PouriaSeyfi:
Due to #3444 issue
some times we need specify data type in html using attribute html.
I add new processDomElementDataType function in HTML.php and handle this.


tests/PhpSpreadsheetTests/Reader/Html/HtmlTest.php
```diff
@@ -363,4 +364,20 @@ public function testBorderWithColspan(): void
...
+                    <tr>
+                        <td data-type="b">1</td>
+                        <td data-type="s">12345678987654321</td>
+                    </tr>
+                </table>';
+        $reader = new Html();
+        $spreadsheet = $reader->loadFromString($html);
+        $firstSheet = $spreadsheet->getSheet(0);
+
+        self::assertEquals(DataType::TYPE_BOOL, $firstSheet->getCell('A1')->getDataType());
```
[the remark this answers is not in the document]
MarkBaker: If the data value doesn't match the specified datatype, then you'll need to ensure that it is cast correctly to match the datatype; and make sure that it's set explicitly in the cell to bypass the value binder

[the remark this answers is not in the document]
PouriaSeyfi: Thanks for your guide.
This is considered in the new commit.

PouriaSeyfi:
There is no any changes. please review my PR.

PouriaSeyfi:
I had some unwanted changes caused by phpstorm formatter. I fixed them and fixup my commits to new one . 
Also I update my branch with new changes in master. there is no conflict.
Please Review and merge my changes

oleibman:
I think there is probably some merit in this suggestion. I am trying to think through some of its ramifications. Before I start, there is one minor code change that the PR needs - you do not need to call setDataType after setCellValueExplicit.

Calling setCellValueExplicit can result in an Exception, e.g. if you supply an invalid data type. Is that what we want to happen here, or are we better off catching the Exception and ignoring the format? Regardless of that decision, we need some unit tests for this.

Your example for setting the type to formula is intriguing. However, your test case proves that formulas are already handled without the need to explicitly set the data type. What is not handled is treating a string beginning with an equal sign as a string (by setting quotePrefix to true) rather than a formula. You could accommodate this in your change by checking for this condition if data type is supplied as s.

I think that dates are an especially good use case for this change. You should add some examples for type d (ISO date, time, and date/time elements). Note that those will be converted from string to the Excel numeric representation, and that the conversion may fail.

PouriaSeyfi:
thanks for your great explanation
I think it is better to check datatypes in flushCell function for better performance. 
As you mentioned it is better to catching the Exception and ignoring the invalid data types.
Also I considered the equal sign in beginning content of cell in case of string datatypes, and I added some new unit test for this scenario and iso dates

src/PhpSpreadsheet/Reader/Html.php
```diff
@@ -283,15 +284,34 @@ protected function releaseTableStartColumn(): string
...
                 //    Only actually write it if there's content in the string
                 //    Write to worksheet to be done here...
-                //    ... we return the cell so we can mess about with styles more easily
-                $sheet->setCellValue($column . $row, $cellContent);
+                //    ... we return the cell, so we can mess about with styles more easily
+
+                // Set cell value explicitly if there is data-type attribute
+                if (isset($attributeArray['data-type'])) {
+                    $datatype = $attributeArray['data-type'];
+                    if ($datatype == DataType::TYPE_STRING || $datatype == DataType::TYPE_STRING2) {
```
oleibman, line 299: Do we also need to test for TYPE_INLINE here?

src/PhpSpreadsheet/Reader/Html.php
```diff
@@ -283,15 +284,34 @@ protected function releaseTableStartColumn(): string
...
+                        if (substr($cellContent, 0, 1) === '=' || is_numeric($cellContent)) {
```
oleibman, line 300: I do not understand the purpose of the is_numeric test. Explicitly setting a numeric value to type string should not require quotePrefix. If your testing indicated that it did require it, please add a test to show why.

tests/PhpSpreadsheetTests/Reader/Html/HtmlTest.php
```diff
@@ -363,4 +364,46 @@ public function testBorderWithColspan(): void
...
+        self::assertEquals('TEXT A TEXT B', $firstSheet->getCell('C1')->getFormattedValue());
+
+        // check string with beginning equal sign is string and not formula
+        self::assertEquals(DataType::TYPE_STRING, $firstSheet->getCell('D1')->getDataType());
+        self::assertEquals('=B1', $firstSheet->getCell('D1')->getValue());
+        self::assertIsString($firstSheet->getCell('D1')->getValue());
+        self::assertTrue($firstSheet->getCell('D1')->getStyle()->getQuotePrefix());
+
+        //check iso date
+        self::assertEquals($firstSheet->getCell('E1')->getValue(), 44613.43090277778);
```
oleibman, line 407: We sometimes see rounding problems. Please use assertEqualsWithDelta here; 1E-8 is probably a good enough delta.

tests/PhpSpreadsheetTests/Reader/Html/HtmlTest.php
```diff
@@ -363,4 +364,46 @@ public function testBorderWithColspan(): void
...
         }
     }
+
+    public function testDataType(): void
+    {
+        $html = '<table>
+                    <tr>
+                        <td data-type="b">1</td>
+                        <td data-type="s">1234567</td>
+                        <td data-type="f">=CONCAT("TEXT A ","TEXT B")</td>
```
oleibman, line 374: I would prefer to eliminate data-type in the formula statement. It should not be necessary. I would still like to retain the formula and the assertions based on it in order to confirm that existing behavior hasn't changed.

oleibman (changes requested):

oleibman:

PR #3448, opened 2023-03-10
Refactor unit tests to ensure that assertions and setup/teardown are in the actual test for ease of review

MarkBaker:
This is:

- [ ] a bugfix
- [ ] a new feature
- [X] refactoring
- [X] additional unit tests

Checklist:

- [X] Code style is respected

Refactor unit tests to ensure that assertions are in the actual test and not in an abstract class; and that setup/teardown are in the test and not an abstract. This means that assertions and setup/teardown are always in the file when reviewing PRs.

Also enforce more rigorous Excel Function implementation by testing the underlying implementation, call via the Calc Engine, and execution from in a worksheet. Separate out unhappy path (exception) checks into a separate test, so that a single test isn't made overcomplex checking for every potentiality.

Ensure the use of ExcelError functions wherever possible rather than string literals for Excel errors, ensures consistency and reduced typos, but also paves the way for being able to return a ExcelError object in the future, so that error testing in the implementations and Calc Engine can look for an ExcelError object rather than testing for a string beginning with `#`.

Scrutinizer may dislike variadics, for variable number of arguments of mixed type; but tough. It's 100% valid PHP, accepted by phpstan, and makes life a lot easier.

Initial work here covers all the database and datetime unit tests for Excel function implementations.


MarkBaker:
OK! It's official. Scrutinizer is taking the piss.
image

PR #3467, opened 2023-03-18
Handle REF Error as Part of Range

### LINKED ISSUE #3453: Calculations containing `#REF!` no longer throw exceptions from 1.26.0
This is:

```
- [x] a bug report
- [ ] a feature request
```

Prior to version `1.26.0` calculating the value of a formula that contains a reference error (eg. `=SUM(B1:#REF!)`) would result in a 

```
[PhpOffice\PhpSpreadsheet\Calculation\Exception]               
  Worksheet!B4 -> Cell coordinate can not be zero-length string
```

From `1.26.0` until now (`1.28.0`) the result is `0` instead of an exception.

```php
<?php

$worksheet->setCellValue('B4', '=SUM(B1:#REF!)');

// Throws Calculation\Exception when < 1.26.0
$value = $worksheet->getCell('B4')->getCalculatedValue();

// Returns 0 in 1.26.0+
\var_dump($value);
```

- [ ] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [x] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

I have tested with `.xlsx`, don't see why it wouldn't affect all other formats.

PHP 8.1.16
PHPSpreadsheet 1.26.0+ (until at least 1.28.0)

oleibman:
Fix #3453. User sets a valid formula (e.g. `=SUM(Sheet2!B1:Sheet2!B3)`), and then does something to invalidate the formula (e.g. delete Sheet2). Excel changes the formula to `SUM(#REF!:#REF!)` when the spreadsheet is saved; apparently someone thought this was a good idea. But PhpSpreadsheet (a) used to throw an Exception when it evaluated the formula, and (b) now gives a result of `0` when evaluating the formula. Neither is ideal. It would be better to propagate the `#REF!` error.

It is likely that more tests are needed, which is why I will keep this in draft status for a bit.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


ndench:
Nice work! This seems to correctly propagate the ref error back to the caller. I wonder if this counts as a BC break since it is different behavior to both pre-1.26.0 and post-1.26.0?

What sort of other tests do you think are necessary? Your current tests seem to cover everything way of getting a `#REF!` error that I know of. Maybe some other ExcelErrors should be tested as well?

oleibman:
I don't think any other Excel errors need to be tested. The only reason this change is needed is because Excel, for whatever reason, changes the formulas. I am not aware of any cases where, for example, it replaces the function argument with `#VALUE!`. But, of course, I also wasn't aware of the `#REF!` case. That's why I wanted more eyes on this change before I move it forward.

I don't really consider this a BC break. It was returning the wrong result before; it seems like any other bug fix in that respect.

Issue #1797, opened 2021-01-15
Chart styling is lost on simple load / save process


lawless-m:
This is:

```
- [ x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

Output styling should be the same format as the input

### What is the current behavior?

Styling is not preserved

### What are the steps to reproduce?

[ChartInput.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/5819988/ChartInput.xlsx)

[ChartOutput.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/5819991/ChartOutput.xlsx)

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$reader = IOFactory::createReader('Xlsx');
$reader->setIncludeCharts(true);
$spreadsheet = $reader->load('ChartInput.xlsx');

$writer = IOFactory::createWriter($spreadsheet, 'Xlsx');    

$writer->setIncludeCharts(true);
$writer->save('ChartOutput.xlsx');

?>

### Which versions of PhpSpreadsheet and PHP are affected?

## 1.16.0 - 2020-12-31

 php --version
PHP 7.4.3 (cli) (built: Oct  6 2020 15:47:56) ( NTS )

pierre-H:
Same problem 😢

kozhevnikovpe:
not fixed
try to run code from 1st message
saved file looks different

oleibman:
@kozhevnikovpe It makes things a lot easier if you describe the differences you see rather than just "saved file looks different". It also helps if you open new issues (possibly cross-referencing the closed issue in your description) rather than keep adding to an issue that was closed 8 months ago.

The sample file 32readwriteLineChart5 was added for the PR which addressed this issue. They are very close, but I agree that there are still some discrepancies.
- The input chart has a border.
- In the input chart, there is no vertical line next to the vertical axis.
- The font sizes on the axis labels and the data points differ.

Can you please open separate issues for each of these, and any other discrepancies that you see. I think I am pretty close to a solution for the first two. The third may require a lot more effort.

PR #3476, opened 2023-03-22
Font/Effects/Theme Support for Chart Data Labels and Axis

### LINKED ISSUE #3477: Xlsx Better Support Needed for Scheme/Theme Colors
This is:

```
- [x] a bug report
- [ ] a feature request
```

Theme/scheme colors should be read as part of Xlsx Reader, and reproduced by Xlsx Writer.

Reader appears to ignore scheme colors. Writer appears to write theme file with hard-coded colors. As a result, the input and output files may differ in their coloring when scheme colors are used. This is especially noticeable in some charts (see PR #3476); I suspect, but am not certain, that other situations might also be affected.

After the PR identified above is merged, this behavior is demonstrated by  sample 32readwriteLineChart6.

- [x] Reader
- [x] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

This is probably just a problem for Xlsx.

All.

oleibman:
Addresses some remaining issues with 32readwriteLineChart5 (see issue #1797). Font size is covered. So are effects, although the results are a bit odd. For the new spreadsheet 32readwriteLineChart6, the Axis labels have a yellow-ish glow, but reading and writing the spreadsheet in PhpSpreadsheet gives them a purple-ish glow. Nevertheless, the new test shows that the output file uses schemeClr accent4, as does the input file. So the effect is handled correctly, but it seems there is likely to be a difference between theme colors (Writer/Xlsx/Theme appears to write hard-coded color schemes). Fixing that will be a great deal more difficult, with a large chance of regression, and will need to happen in a separate PR (one that I am not currently investigating, but I will open a new issue). Effects using srgbClr (and probably sysClr) should be okay.

Fix #3477. It turns out the color schemes are not so difficult to support as I had feared, and it probably applies only or mostly to charts. When reading Xlsx, the theme colors will now also be used for writing. This means that a file can be loaded and saved and its chart colors will now be preserved. If the spreadsheet is created new, Excel 2007-2010 colors are used. The writer is currently hard-coded to use them, so this avoids making this a breaking change. The theme colors can be explicitly changed if desired, and Excel 2013+ colors can be introduced very easily.
```php
$spreadsheet->getTheme()
    ->setThemeColorName(Theme::COLOR_SCHEME_2013_PLUS_NAME);
```
Likewise, if the old behavior of changing to the 2007-2010 scheme rather than using the input values is desired, that is easy to achieve after the load has taken place.
```php
$spreadsheet->getTheme()
    ->setThemeColorName(Theme::COLOR_SCHEME_2007_2010_NAME);
```
The new Theme class introduced by this change can easily be extended to include Fonts and Effects. Unlike Colors, I am unsure what the practical effects of changing those to, say, the 2013+ defaults would be.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
As usual, no concern with Scrutinizer's "complexity" warning.

PR #3508, opened 2023-04-03
Add Ability to Ignore Cell Errors in Excel

### LINKED ISSUE #1141: Ignore cell errors in Excel (e.g. "Number Stored as Text")
This is:

```
- [ ] a bug report
- [x] a feature request
```

When a cell contains a value that looks like a number but is intentionally formatted as text (sometimes this is desirable because the values might not always be purely numeric and you don't want Excel to reformat them), and written to an XLSX file, Excel displays a warning beside the cell when viewing the worksheet.

Excel has an option to "Ignore" the error, which appears to be a flag that is then written to the cell. It would be great if this functionality could be exposed as part of the PhpSpreadsheet API.

image

For reference, there is a JavaScript project for writing XLSX files called *SheetJS* which has this feature, [here is how they do it](https://github.com/SheetJS/js-xlsx/blob/36fcb00045ad6837383debab4b31caaed4e3b5f7/bits/67_wsxml.js#L573).

The warning is visible to users next to Text formatted cells that Excel thinks are numbers, and there's nothing I can do about it. Users might choose to "Convert to Number" and create formatting problems (including data corruption if exported to CSV).

N/A

oleibman:
Fix #1141, which had been closed as stale, but which I have reopened. Excel will show cells with certain "errors" with a green triangle in the upper left. The suggestion in the issue to use quotePrefix to suppress the numberStoredAsText error is ineffective. In Excel, the user can turn this indicator off for individual cells. Cells where this is turned off can be detected at read time, and PhpSpreadsheet will now process those. In addition, the user can explicitly set the ignored error as in Excel.
```php
$cell->getIgnoredErrors()->setNumberStoredAsText(true);
```

There are a number of different errors that can be ignored in this fashion. This PR implements `numberStoredAsText` (which is likely to be by far the most useful one), `formula`, `twoDigitTextYear`, and `evalError`, all of which are demonstrated in the new test spreadsheet. There are several others for which I am not able to create good examples; I have not implemented those, but they can be easily added if needed (`calculatedColumn`, `emptyCellReference`, `formulaRange`, `listDataValidation`, and `unlockedFormula`).

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Cell/Cell.php
```diff
@@ -71,6 +71,18 @@ class Cell
      */
     private $formulaAttributes;
 
+    /** @var bool */
```
MarkBaker, line 74: I might move these (and the corresponding individual getters/setters) into a separate class, and just have a single getter/setter in Cell for `get/setIgnoreErrors()`

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -2222,4 +2228,48 @@ private static function extractPalette(?SimpleXMLElement $sxml): array
...
+    {
+        $attributes = self::getAttributes($xml);
+        $sqref = (string) ($attributes['sqref'] ?? '');
+        $numberStoredAsText = (string) ($attributes['numberStoredAsText'] ?? '');
+        $formula = (string) ($attributes['formula'] ?? '');
+        $twoDigitTextYear = (string) ($attributes['twoDigitTextYear'] ?? '');
+        $evalError = (string) ($attributes['evalError'] ?? '');
+        if (!empty($sqref)) {
+            $explodedSqref = explode(' ', $sqref);
+            $pattern1 = '/^([A-Z]{1,3})([0-9]{1,7})(:([A-Z]{1,3})([0-9]{1,7}))?$/';
```
MarkBaker, line 2242: Can we use any of the existing pattern constants?

oleibman: The patterns in Calculation are too complicated - they allow for worksheet names and dollar signs and case-insensitivity, none of which can occur here (and probably other stuff - I haven't analyzed the entire pattern). Are other more simple patterns defined elsewhere?

src/PhpSpreadsheet/Cell/Cell.php
oleibman: I'm not sure what you have in mind here. Do you want the user to do something like:
```php
$cell->setIgnoreErrors('numberStoredAsText'); // obviously provide constants
```
Where the code in Cell does something like:
```php
private $ignoreErrors = new IgnoreErrors();

public function setIgnoreErrors(string $type, bool $flag)
{
    $this->ignoreErrors->apply($type, $flag);
}
```

MarkBaker: I was thinking more like
```
$cell->getIgnoreErrors()->setignoredErrorFormula(true);
```
doesn't even need a setter in the Cell, just one new method there for the getter; then in the Cell\IgnoreErrors class, a getter/setter for each of the error types that can be ignored, defaulting to false unless set by the Reader, or by a user.
It keeps the increase in the number of new methods/properties in Cell down to a minimum; but still gives explicit methods for each error type that can be ignored

oleibman: Got it - thanks. Latest push implements this suggestion.

PR #3527, opened 2023-04-20
Avoid allocating RowDimension when the row is visible by default

goetas:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

When there is an auto filter range defined as A1:A:100000, it forces the filter to iterate over all the rows in the range and see if needs to hide them ($this->workSheet->getRowDimension((int) $row)->setVisible($result);)

that allocates a RowDimension object for each row for each sheet, and boom, a lot of memory wasted.

This change allocates the `RowDimension` object only if the row visibility needs to be changed.

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/AutoFilter.php
+++ b/src/PhpSpreadsheet/Worksheet/AutoFilter.php
@@ -1049,7 +1049,11 @@ class AutoFilter
                 }
             }
             //    Set show/hide for the row based on the result of the autoFilter result
-            $this->workSheet->getRowDimension((int) $row)->setVisible($result);
+            //    If the RowDimension object has not been allocated yet and the row should be visible,
+            //    then we can avoid any operation since the rows are visible by default (saves a lot of memory)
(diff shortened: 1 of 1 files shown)


oleibman:
You need to add a unit test to demonstrate the effectiveness of your change, perhaps something which demonstrates that the an auto-filtered cell is visible even though the row dimension is not allocated. (That's what I'm assuming your change does; please correct me if I've misunderstood.)

goetas:
My guess is that i was pushing to the limits this library. We have ~5 tabs with ~10-90k rows each with formulas, filters and styling... all of it was really CPU and memory intensive (even by dropping some features or monkey patching some parts of the lib, 20G memory and 8 cores were not able to keep a satisfactory generation time for us). We decided to use the less supported https://en.wikipedia.org/wiki/SpreadsheetML , that has its own problems but at least CPU and memory problems are gone.

I'm closing this as I'm not actively working on it anymore.

PR #2096, opened 2021-05-13
Fix for #1449 / Removing rows or columns that include range edges

jabouillei:
- [X] a bugfix
- [ ] a new feature

Checklist:

- [X] Changes are covered by unit tests
- [X] Code style is respected
- [X] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ ] CHANGELOG.md contains a short summary of the change
- [ ] Documentation is updated as necessary

### Why this change is needed?
Removing rows or columns that overlap the edge of a range should adjust the edges of the range.
Also, formulas in column A should be adjusted even if only 5 columns are being removed starting with column G. (A formula in column A may refer to a range that includes column G.)

Issue #1449, opened 2020-04-23
Removing rows or columns that include range edges


jabouillei:
This is a bug report. I have a fix for PHPExcel and confirmed that the fix is not present in PhpSpreadsheet.
https://github.com/PHPOffice/PHPExcel/compare/1.8...jabouillei:patch-1

### What is the expected behavior?

Removing rows or columns that overlap the edge of a range should adjust the edges of the range.
Also, formulas in column A should be adjusted even if only 5 columns are being removed starting with column G.  (A formula in column A may refer to a range that includes column G.)

### What is the current behavior?

The critical function for adjusting references only considers the case of adding rows or columns, not removing them.  It also works if the removed rows or columns do not overlap the edge of the range.
There is also an uncommented "if" that "continue"s the loop that updates formulas in the initial columns of the sheet, so for example, formulas in column A do not get updated.

### What are the steps to reproduce?

I provided a fix and have already put more time into this that I was supposed to, so I'm not providing a _complete_ example, but here is what I've been using...

<?php
require_once('PHPExcel/PHPExcel/IOFactory.php');
$objReader = PHPExcel_IOFactory::createReader('Excel2007');
$objReader->setIncludeCharts(true);
$content = $objReader->load('/tmp/testremovecolumn.xlsx');
foreach ($content->getWorksheetIterator() as $worksheet) {
    //$worksheet->removeColumn('F',2);
    $worksheet->removeRow(3,3);
}
$objWriter = PHPExcel_IOFactory::createWriter($content, "Excel2007");
$objWriter->setPreCalculateFormulas(false);
$objWriter->setIncludeCharts(true);
$objWriter->save('/tmp/testremovecolumn2.xlsx');
?>
That just requires a test file that has, for example, =SUM(C4:F7) in cell A1 and a bunch of numbers in the range C4:F7.

### Which versions of PhpSpreadsheet and PHP are affected?

I'm using PHPExcel 1.8 on php 7.2.12 and php 5.6.16.  I looked at the source code for PhpSpreadsheet master and see that neither problem has been addressed yet.
Someday, we'll get all our servers on php 7 and we'll probably move to PhpSpreadsheet.  If the issue hasn't been addressed by then, I'll post a fix for PhpSpreadsheet master instead of just PHPExcel, but that might be a couple of years.

jabouillei:
[...]

PR #3528, opened 2023-04-20
Partial Solution for Removing Rows or Columns that Include Edge Ranges

oleibman:
This PR will probably remain in draft status for some time. It needs updates, and I need more time to understand exactly what it is doing. It is a replacement for PR #2096, intended as a fix for issue #1449. That PR has been stuck, with a good deal of acrimony, for over 2 years. In the interim, all the parameters for the functions involved were renamed, the functions were refactored to have fewer parameters with new class properties replacing the missing parameters, special code was added for absolute references, and code was refactored putting some of the methods involved in a new class under a new method name. In addition, the original PR was not Php8-compliant (one of the sticking points regarding implementing it), and there were a great many formatting problems (another sticking point). This has made it very difficult to figure out exactly what was intended in the original PR.

This PR (a) addresses some of the issues associated with the original issue (deleting columns/rows at the end of a range passed to a formula *on the same sheet*), and (b) does not cause any existing tests to fail. Many more unit tests are needed (another sticking point); in particular, there are changes involving data validations, comments, defined names, and other elements which are not adequately tested.

Aside from those, PhpSpreadsheet does not adjust formulas on a different sheet which involve deleted rows/columns, and it should, as Excel does.

Similarly, it appears that Excel doesn't do anything extraordinary when the right/bottom part of the range is deleted, but when the left/top part is deleted, it converts the range to #REF! at least some of the time. This needs more research; if confirmed, it needs to be added to PhpSpreadsheet.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
I'm tired of looking at this ticket. This is a clear example of "perfect is the enemy of goo". Waiting for perfection is going to prevent its useful aspects from seeing the light of day. I am taking it out of draft status, and intend to install it soon. I will open a new issue to document what remains to be done.

Issue #3529, opened 2023-04-21
Why is there not a Offset for MemoryDrawing but Drawing when Html::writeImageInCell？


dubox:
This is a part of source code of Html::writeImageInCell.
Class MemoryDrawing has offsets attributtes also, and I dont understand why not sets 'top','left' for the 'img' Tag.

```php
 if ($drawing instanceof Drawing) {
                $filename = $drawing->getPath();

                // Strip off eventual '.'
                $filename = (string) preg_replace('/^[.]/', '', $filename);

                // Prepend images root
                $filename = $this->getImagesRoot() . $filename;

                // Strip off eventual '.' if followed by non-/
                $filename = (string) preg_replace('@^[.]([^/])@', '$1', $filename);

                // Convert UTF8 data to PCDATA
                $filename = htmlspecialchars($filename, Settings::htmlEntityFlags());

                $html .= PHP_EOL;
                $imageData = self::winFileToUrl($filename, $this->isMPdf);

                if ($this->embedImages || substr($imageData, 0, 6) === 'zip://') {
                    $picture = @file_get_contents($filename);
                    if ($picture !== false) {
                        $imageDetails = getimagesize($filename) ?: [];
                        // base64 encode the binary data
                        $base64 = base64_encode($picture);
                        $imageData = 'data:' . $imageDetails['mime'] . ';base64,' . $base64;
                    }
                }

                $html .= '<img style="position: absolute; z-index: 1; left: ' .


               /**** give Offset ****/     
$drawing->getOffsetX() . 'px; top: ' . $drawing->getOffsetY() . 'px; width: ' .
                    $drawing->getWidth() . 'px; height: ' . $drawing->getHeight() . 'px;" src="' .
                    $imageData . '" alt="' . $filedesc . '" />';
            } elseif ($drawing instanceof MemoryDrawing) {
                $imageResource = $drawing->getImageResource();
                if ($imageResource) {
                    ob_start(); //  Let's start output buffering.
                    imagepng($imageResource); //  This will normally output the image, but because of ob_start(), it won't.
                    $contents = (string) ob_get_contents(); //  Instead, output above is saved to $contents
                    ob_end_clean(); //  End the output buffer.

                    $dataUri = 'data:image/png;base64,' . base64_encode($contents);

                    //  Because of the nature of tables, width is more important than height.
                    //  max-width: 100% ensures that image doesnt overflow containing cell
                    //  width: X sets width of supplied image.
[...]

PR #3535, opened 2023-04-23
Update Html.php

dubox:
Add position attr for image whitch driving by MemoryDrawing

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [ ] Code style is respected

[https://github.com/PHPOffice/PhpSpreadsheet/pull/3529](https://github.com/PHPOffice/PhpSpreadsheet/pull/3529)


oleibman:
I have updated your description to point to issue 3529 rather than 3530. If you feel it should be 3530, please explain why.

You need to include a unit test to demonstrate that your code is working as desired. It might also be helpful to include a sample so that we can see the practical effect of your change on the output file.

dubox:
> I have updated your description to point to issue 3529 rather than 3530. If you feel it should be 3530, please explain why.
> 
> You need to include a unit test to demonstrate that your code is working as desired. It might also be helpful to include a sample so that we can see the practical effect of your change on the output file.

I have added a unittest at tests/PhpSpreadsheetTests/Writer/Html/MemoryDrawingOffsetTest.php

oleibman:
Thank you for supplying the test. I have a few minor changes that I would like you to make to it, but won't be able to document those until tonight.

tests/PhpSpreadsheetTests/Writer/Html/MemoryDrawingOffsetTest.php
```diff
@@ -0,0 +1,50 @@
...
+    /**
+     * @dataProvider dataProvider
+     */
+    public function testMemoryDrawingOffset(int $w, int $h, int $x, int $y): void
+    {
+        $spreadsheet = new Spreadsheet();
+        $sheet = $spreadsheet->getActiveSheet();
+        $drawing = new MemoryDrawing();
+        if (($image = file_get_contents('https://avatars.githubusercontent.com/u/1836015')) === false) {
+            throw new Exception('image err1');
```
oleibman, line 22: Having to fetch the drawing from the web adds an extra dependency to this test, one that is certainly not needed for it. Please save the drawing as a local file, somewhere like `tests/data/Reader/HTML/memoryDrawingTest.jpg`. After that, it is not necessary to throw an exception when file_get_contents or imagecreatefromstring fails - you should just assert that the result of the function call is not false. Your test code should look something like:
```php
        $image = file_get_contents('tests/data/Reader/HTML/memoryDrawingTest.jpg');
        self::assertNotFalse($image, 'unable to read file');
        $image = imagecreatefromstring($image);
        self::assertNotFalse($image, 'unable to create image from string');
```

dubox: I have done those, and thank u for point out the problems in my code.

PR #3605, opened 2023-06-03
Improvements for Data Validation

### LINKED ISSUE #3592: Data Validation in XML reader
This is:

```
- [ ] a bug report
- [x] a feature request
```

Respect datavalidation in the XML reader

Validation is not yet parsed.

- [X] Reader

Only XML format is affected

All versions are affected

I attach a simple XML file which should cover the most relevant cases - if you need anything more just ask.
As I am still stuck with an ancient version I implemented this myself (quite rude, but it still might be used as a starting point - actually it seems easier than I expected. $children_x are the Excel-namespaced children of the worksheet).

Care must be taken with the "hide"-Options: XML uses "show" as a default and can only hide; in PhpSpreadsheet the defaults are set the other way round.

One big issue seems to be performance if ranges are used. Column A in the sample file uses a simple, single validation rule, column B uses one with exception of B1 - the XML reflects how Excel thinks of this. I have not found a way to replicate this into PhpSpreadsheet, but using Coordinate::extractAllCellReferencesInRange() to split ranges into single cells is really expensive.
[validation.txt](https://github.com/PHPOffice/PhpSpreadsheet/files/11611320/validation.txt)
[validate.txt](https://github.com/PHPOffice/PhpSpreadsheet/files/11611323/validate.txt)

### LINKED ISSUE #3594: FreezePanes in XML reader
> I will not be adding SplitHorizontal and its 4 associated items to my PR;
> it will need a lot more research.

I have been going through the trunk code and either am I missing something or PhpSpreadsheet does not (yet) understand *split* mode, even though it is still supported in modern Excel.

*Freeze* is supported, however, so this could be imported from XML as well. A pane is frozen if  (and only if) the /FreezePanes/ tag is present. In this case /SplitHorizontal/ and /SplitVertical/ determine the number of rows to be seen in the top-left pane (if any, otherwise the respective tag can be left out), /TopRowBottomPane/ and /LeftColumnRightPane/ determine the first row/column visible in the other panes.
Worksheet::freezePane() seems for me the corresponding method for all of those tags.

I could not find any method matching /ActivePane/, so perhaps this is not yet supported by PhpSpreadsheet.

A minimum example implementing this is here:
[freeze.txt](https://github.com/PHPOffice/PhpSpreadsheet/files/11616454/freeze.txt)
The top-left pane contains 2 rows (SplitHorizontal) and 6 columns (SplitVertical). The top-left cell of the bottom-right pane is set to row 11 and column 21 (U), as XML-values are 0-based.

A completely different thing is *splitting* which occurs if /FreezePanes/ is not used. /SplitHorizontal/ and /SplitVertical/ then specify offsets in *points* from the upper left corner to mark where the windows is split (with separate scroll bars for the left/right and top/bottom part). If PhpSpreadsheet does not support this, there is no use to care for this mode at the moment.

oleibman:
Fix #3592. Fix #3594. The original issue asked that Data Validation be added for Xml spreadsheets. However, doing so exposed some other Data Validation problems, which are corrected along with the Xml portion.

The main other problem is that the code for `Cell\DataValidator::isValid` covered almost no possibilities. It is expanded to include all possibilities, except for Custom Types, which I will continue to think about. Many tests are added.

An obscure problem is that Xlsx Reader does not quite work properly when the Data Validation cells are beyond the high-used column or row. We use the high-used values because it is completely impractical to set all the cells in a range when an entire column or row is selected. This change causes the leftmost top cell in a range to be explicitly allocated, affecting the high-used values. It is not 100% effective, but it will be much more difficult to get into the problem situation, which I will continue to think about.

Although our expectation is that we will not see any Xml Spreadsheets using non-standard namespacing, much of the Xml Reader code is changed to use proper namespace techniques. A few areas which just had nothing to do with this change continue to use less robust techniques; I plan to address those soon after implementing this change.

One of the Data Validation types is for time of day. Testing revealed that `Shared\Date::convertIsoDate` was insufficiently robust when determining whether its input consisted of a time of day without date. It is now more robust.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Big surprise - added a lot of functionality, Scrutinizer thinks it's too complex. Oh well.

PR #3608, opened 2023-06-07
Added a trait to help asserting cell values in PHPUnit tests

stollr:
This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

When writing functional tests and one would like to check if the generated spreadsheet has the correct values it is possible to compare every cell by its coordinates. But this is very effortful and hard to read.

With this PR I want to add a trait that can be used in PHPUnit tests to compare all cell values at once.

Look at the following example where the underlying spreadsheet may look like that:

| Firstname | Lastname | Age |
|----------------|----------------|-------|
| Christian | Miller | 35 |
| Tim | Lemon | 22 |

The test file may look like that:

```php
class SpreadsheetGeneratorTest extends TestCase
{
    use \PhpOffice\PhpSpreadsheetTests\Custom\SpreadsheetAssertionsTrait;

    public function testGenerate(): void
    {
        // prepare the test
        $spreadsheet = IOFactory::load($generatedFile->getPathname());
        $this->assertSpreadsheetCellValuesMap([
             ['Firstname', 'Lastname', 'Age'],
             ['Christian', 'Miller', '35'],
             ['Tim', 'Lemon', '22'],
         ], $spreadsheet->getActiveSheet());
    }
}
```

If there are rows which are not needed to check, just add an empty array

```php
        $this->assertSpreadsheetCellValuesMap([
             [], // the first row is only informational text - no need to make any assertions
             ['Christian', 'Miller', '35'],
             ['Tim', 'Lemon', '22'],
         ], $spreadsheet->getActiveSheet());
```

If you like this. I can also add some documentation, if you can tell me where to add it.

### DIFF
--- a/tests/PhpSpreadsheetTests/Custom/SpreadsheetAssertionsTrait.php
+++ b/tests/PhpSpreadsheetTests/Custom/SpreadsheetAssertionsTrait.php
@@ -1,0 +1,43 @@
+<?php
+
+namespace PhpOffice\PhpSpreadsheetTests\Custom;
+
+use PhpOffice\PhpSpreadsheet\Worksheet\Worksheet;
+use PHPUnit\Framework\Assert;
+
+trait SpreadsheetAssertionsTrait
+{
+    /**
+     * Assert that the items of the passed value map are identical to the cell values of the worksheet.
+     *
+     * @example Basic usage: ```
+     *     $this->assertSpreadsheetCellValuesMap([
+     *         ['Firstname', 'Lastname', 'Age'],
+     *         ['Christian', 'Miller', '35'],
+     *         ['Tim', 'Lemon', '22'],
+     *     ], $spreadsheet->getActiveSheet());
+     * ```
+     * @example You can ommit the comparison of a row by adding an empty array: ```
+     *     $this->assertSpreadsheetCellValuesMap([
+     *         [], // the first row won't be compared
+     *         ['Christian', 'Miller', '35'],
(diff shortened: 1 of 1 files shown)


oleibman:
What advantage would this trait have over  (used in over a dozen existing tests):
```php
        $this->assertSame([
             ['Firstname', 'Lastname', 'Age'],
             ['Christian', 'Miller', '35'],
             ['Tim', 'Lemon', '22'],
         ], $spreadsheet->getActiveSheet()->toArray());
```

stollr:
You're right that's possible. The trait has the advantage that some rows may be ignored. But i don't know if this is enough of an advantage to include it. Feel free to close this PR if you do not need it.

oleibman:
I think `rangeToArray` would serve your "ignore rows" cases. In the example where you want to skip the first row:
```php
        $this->assertSame([
             ['Christian', 'Miller', '35'],
             ['Tim', 'Lemon', '22'],
         ], $spreadsheet->getActiveSheet()->rangeToArray('A2:C3'));
```
Admittedly, if you want to ignore rows other than just at the beginning (or at the end), you would require multiple assertions of this type. I am not convinced that there is a strong enough use case for that, so as you suggested, I will close this PR. Nevertheless, thank you for the suggestion and the work you put into it.

stollr:
You're welcome ☺️

PR #3623, opened 2023-06-25
Permit Meta Viewport in Html

### LINKED ISSUE #3565: Viewport Implementation
This is:

```
- [ ] a bug report
- [X] a feature request
- [X] **not** a usage question
```

Solution given in: https://github.com/PHPOffice/PhpSpreadsheet/commit/7f77c0ad940851b06cde7715d837e4d29205b6bc

Can I open a pull request?

If an attribute is set to a html writer, then the given html should fit the width of the browser.

Nothing.

oleibman:
Fix #3565. Requester suggested adding viewport as a property of Html Writer. I think it would fit better as a Document Property, so that it can be both read and written, just like other meta properties, making it easier to load an Html spreadsheet and save it as a new one preserving this property. It is an open question whether we should try to preserve this property when writing to other spreadsheet types, e.g. as a custom property for Xlsx. For now, I am not attempting that. Requester also suggested this as a boolean property, but there are other parameters and different values that might be used for viewport. So I am defining it as a string property, with a constant SUGGESTED_VIEWPORT that can be used for requester's use case.

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
No concern at all with Scrutinizer "complexity" failure.

PR #3653, opened 2023-07-27
Fixed inconsistent string handling in SUM() implementations, issue #3652

### LINKED ISSUE #3652: Inconsistent string handling between Sum::sumErroringStrings() and Sum::sumIgnoringStrings()
### This is:

- [X] a bug report
- [ ] a feature request

### What is the expected behaviour?
Numeric string handling in `Sum::sumErroringStrings()` and `Sum::sumIgnoringStrings()` should be consistent.

### What is the current behaviour?
`sumErroringStrings` casts all string params to integers whereas `sumIgnoringStrings` just sums them. The expected behaviour would be for both functions to cast string inputs to floats.

```php
<?php

use PhpOffice\PhpSpreadsheet\Calculation\MathTrig\Sum;

// This outputs int(4)
var_dump(Sum::sumErroringStrings("1.2", "3.4"));

// This outputs float(4.6)
var_dump(Sum::sumIgnoringStrings("1.2", "3.4"));
```

- [ ] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [X] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

Yes, affects all formats

All PHP versions and 1.28 of phpspreadsheet

betterphp:
Fix #3652.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [X] Code style is respected

To resolve inconsistent results from the two SUM() implementations. Issue #3652


oleibman:
Thank you for pointing out the problem. You are correct that the code should be more consistent between sumIgnoringStrings and sumErroringStrings. However, as they are currently coded, sumIgnoringStrings is more correct - you should be changing sumErroringStrings to look more like it rather than vice versa. The logic in sumErroringStrings to cast the value to int is really intended only for null and null-string arguments. I think it needs to be changed to:
```php
                if (empty($arg)) {
                    $arg = 0;
                }
```
All other string arguments will automatically be cast to float or int as appropriate.

Once you have made that change, there is no longer any reason to change RefErrorTest, so you should restore it to its original form. However, your observation makes it clear that we have not included a test for a string which represents a floating point value in the SUM tests (tests/data/Calculation/MathTrig/SUM and SUMLITERAL). Please add such a test to each of those.

oleibman:
The existing tests for float supply the literal as a float, not as a string (which is the bug you uncovered). So, we have:
```
[39.1, 5.7, true, 30, 2.4],
```
But we need to add a test for
```
[39.1, '5.7', true, 30, 2.4],
```
As to your question, can you give me an example where SUM yields a different result than calling sumerroringstrings directly?

S-code652 (approved):

src/PhpSpreadsheet/Calculation/MathTrig/Sum.php
```diff
@@ -47,7 +47,7 @@ public static function sumIgnoringStrings(...$args)
      *
      * @param mixed ...$args Data values
      *
-     * @return float|string
+     * @return float|integer|string
```
oleibman, line 50: To avoid php-cs-fixer error, use "int" rather than "integer".

oleibman (approved):

betterphp:
Hey, sorry I've been away for a few weeks - thanks for the extra commits @oleibman  and for merging :)

PR #3695, opened 2023-08-30
replace ezyang/htmlpurifier with voku/anti-xss

dmytromikhieiev1985:
This is:

- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

ezyang/htmlpurifier has LGPL licence. In this PR htmlpurifier replaced with package which has MIT licence

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## Unreleased - TBD
 
+### MINOR BREAKING CHANGE
+
+- Typing was strengthened by leveraging native typing. While this should not change any behavior, it might need minor
+  adaption of your code if you use static analysis tools such as PHPStan or
+  Psalm. [PR #3718](https://github.com/PHPOffice/PhpSpreadsheet/pull/3718)
+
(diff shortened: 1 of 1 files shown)


oleibman:
I don't have a problem with this change insofar as it doesn't break any tests. And I know you're not introducing this problem, but I really dislike `testXssInComment`. It tests to confirm that the string after being sanitized doesn't match the string going in, but that's a weak test. A poor sanitizer, for example, could just add a space after the first space it sees, and the test would pass. Could you rewrite the test so that it takes an expected result as a parameter and then compares the result to the expected result? Also, looking at the github page for voku/anti-xss, it has at least 2 examples which aren't in our test set (example 2 hexadecimal html and example 7 iframe (except that unlike the example we don't want to allow iframe)); can you add tests for those, and any other examples that you don't see represented?

oleibman:
Here's what I think XssVulnerabilityTest should be:
[XssVulnerabilityTest.php.txt](https://github.com/PHPOffice/PhpSpreadsheet/files/12488485/XssVulnerabilityTest.php.txt)

PR #3696, opened 2023-08-30
Writer Xls Handle Characters Outside Unicode BMP

### LINKED ISSUE #642: PhpSpreadsheet generates a faulty Xls file when cells contain emojis
This is an old story and it only happens when generating Xls files, it doesn't with Xlsx files. The same issue in the deprecated PHPExcel https://github.com/PHPOffice/PHPExcel/issues/1320
No one said why this is happening.
The reader works fine with emojis.
Isn't this issue fixable?

oleibman:
Fix #642. Opened over 5 years ago, probably the oldest problem I've worked on. And https://github.com/PHPOffice/PHPExcel/issues/1320, opened a year before that. And https://github.com/SpartnerNL/Laravel-Excel/issues/1521.

Shared/StringHelper::UTF8toBIFF8UnicodeLong calculates incorrect length for strings when they contain characters outside Unicode BMP. Xls uses UTF-16 to encode its strings, and characters outside BMP require a surrogate pair to encode. PhpSpreadsheet (and PhpExcel before it) have been counting these as a single character, but Excel counts them as 2. Change to compute the length as half the number of bytes in the UTF-16 string, as Excel does.

A formal test is added, but it's a bit difficult to follow. So I aso added a non-BMP emoji to sample 27template.xls, which will cause it to be both read by Xls reader and written by Xls writer. This would previously have created a corrupt worksheet. The emoji is now handled correctly.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Scrutinizer error is weird. It attributes added complexity to a source member which isn't being changed by this PR. I think this must have been a timing problem between the merging of a different PR (which did change the source member in question) and the pushing of this one. The error will probably disappear when I update this branch. Regardless, I have no concerns about it.

PR #3707, opened 2023-09-02
Table Destructor

oleibman:
There is a circular reference between Worksheet and Table, which can lead to a memory leak. I tried to break this connection in the Worksheet destructor earlier, but this seemed to cause some sort of error with Phpunit 10 on Github, an error that I am unable to duplicate in my local environments. This PR will allow me to explore the problem to see if I can come up with any useful diagnostic data. If I get to a combination that works, I will consider merging it.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
This seems to have worked, so I am moving it out of draft status. Differences between this and prior attempt:
- Don't know if this is a different version of Phpunit 10.
- Added destructor to Table. I can't see why this would matter, but other descendants of Worksheet with potential circular references (e.g. Row and Column) do so.
- Changed Worksheet destructor to empty Table collection directly rather than calling a subroutine which returns self. I can sort of convince myself that this is the change that mattered if I don't think about it too much.

PR #3709, opened 2023-09-06
ListWorksheetInfo/Names for Html/Csv/Slk

### LINKED ISSUE #3706: Create interface for support of multiple worksheets per file
This is a feature request.

When using `IOFactory::identify()` and `IOFactory::createReader()`, the resulting reader may or not support multiple worksheets. It would be helpful to have an interface, perhaps called `multipleWorksheetInterface`, which contains methods like `::listWorksheetNames()` and `::setLoadSheetsOnly()`. Any reader class with these methods would implement the interface. 

This would allow using code like `if ($reader instanceof multipleWorksheetInterface) ...` to identify readers that have this support.

One can currently do `if (method_exists($reader, 'listWorksheetNames')) ...`. An interface would be a more convenient way of detecting this support.

oleibman:
Fix #3706. ListWorksheetInfo is implemented for all Readers except Html. For most (not all), ListWorksheetInfo is more efficient than reading the spreadsheet. I can't think of a way to make that so for Html, but that shouldn't be a reason to leave it unimplemented.

ListWorksheetNames is not implemented for Html, Csv, or Slk. It isn't terribly useful for those formats, but that isn't a reason to omit it. The requester's use case consists of using IOFactory to create a reader for a file of unknown format and determining the first sheet name. That seems legitimate, but it is currently not possible without extra user code if the file is Html, Csv, or Slk; this PR will make it possible.

When Excel opens a Slk or Csv file, the sheet name is based on the file name. PhpSpreadsheet does this for Slk, but it uses a default name for Csv. I am not interested in creating a break for that behavior, but I have added a new boolean property `sheetNameIsFileName` with a setter to Csv Reader. The requester actually mentioned that possibility in our discussion, although it is not essential to the request.

As an adjunct to the issue, the requester wishes to use the worksheet name in `setLoadSheetsOnly`. That is already possible for Html, Csv, and Slk, but that particular property is ignored for those formats. I do not see a reason to change that behavior. This treatment is now explicitly noted in the documentation for property `loadSheetsOnly`.

There had been no tests for what happens when `loadSheetsOnly` is specified but no sheets match the criteria for the formats for which this makes sense (Xlsx, Xls, Ods, Gnumeric, Xml). The behavior was not consistent - some formats threw an Exception while others continued with a single empty worksheet. All cases attempt to set the active sheet, and they will now all throw identical Exceptions when they attempt to do so in this situation. Tests are added for each.

There also had been no tests for `loadSheetsOnly` returning more than one sheet. One is added.

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:

src/PhpSpreadsheet/Reader/Xls.php
```diff
@@ -417,6 +417,9 @@ class Xls extends BaseReader
      */
     private $baseCell;
 
+    /** @var bool */
+    private $activeSheetSet = false;
```
PowerKiKi, line 421: Please refrain from introducing non-native typing. Now that we require PHP 8.0 we should be able to natively type almost everything, especially properties.

oleibman: Will do.

PR #3710, opened 2023-09-06
Drop Php7.4 Support

oleibman:
Fix #3634. Its time has come. Based on our experience dropping 7.3, I anticipate problems with Github Actions. No code changes.

This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [ ] Changes are covered by unit tests
  - [ ] Changes are covered by existing unit tests
  - [ ] New unit tests have been added
- [ ] Code style is respected
- [x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ ] CHANGELOG.md contains a short summary of the change and a link to the pull request if applicable
- [ ] Documentation is updated as necessary

### Why this change is needed?

Provide an explanation of why this change is needed, with links to any Issues (if appropriate).
If this is a bugfix or a new feature, and there are no existing Issues, then please also create an issue that will make it easier to track progress with this PR.

PR #3713, opened 2023-09-06
Drop PHP 7.4 support

PowerKiKi:
This is:

- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

This is according to our formal, published, policy to only support EOL PHP after 6 months.

See https://phpspreadsheet.readthedocs.io/en/latest/#php-version-support

Also share the exact same dev deps across all PHP version for GitHub Actions so runs are faster, much most importantly they are stable and predictable. And we decide manually when we want to migrate to PHPUnit 10.

Supersedes #3710


PowerKiKi:
@oleibman, ever slightly out of topic, but I strongly suggest to not try supporting multiple PHPUnit versions in the same codebase. Because of PHP 7.4 we had no choice to use PHPUnit 8 and 9. But before this PR we supported PHPUnit 8, 9 and 10. This is _a lot_ of work and juggling, as you saw yourself. And unfortunately it brings no value to us. It is much easier to keep our dev dependencies as stable and predictable as possible.

So now that PHP 7.4 is out the picture, all PHP versions can share the exact same version of PHPUnit 9 (not 10). I did that by replacing `composer update` by `composer install` in GitHub Actions (amongst a few other things). I strongly encourage you to keep this way of working, shared dev-deps, as long as possible. And whenever we feel confident all PHP versions can move to PHPUnit 10, then we can migrate once and for all.

oleibman:
@PowerKiKi Noted, thanks. It was an intellectually satisfying task to get Phpunit10 working and continuously tested (practical because coverage test used locked version 9 while 8.1 test used latest version 10 so that both were exercised), but, obviously, no problem with reverting to the single version. (I've also gone through the same exercise with PhpWord, though not installed of course.) I'll keep my notes handy for when Php8.0 is dropped next year, allowing use of Phpunit10 throughout.

Your change was a lot bigger than I'd envisioned. Nice work finding all those opportunities to simplify code right away!

PowerKiKi:
I'd like to follow up with a few PR adding types, since we can leverage union types since PHP 8.0 and thus get rid of some PHPdocs. So all-in-all dropping PHP 7.4 is even bigger than that :P

PR #3738, opened 2023-09-15
Added Conditional Formatting: ColorScale for Xlsx

redtailmatt:
This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

ColorScale was not previously supported. Now it is. Currently, only the Xlsx Writer has been updated.


oleibman:
Thank you for starting this; it seems like it could be a valuable addition. However, there are a lot of changes that will need to be made before it is ready.

You will need to eliminate the phpcs problems. I believe these are due to the use of tabs rather than spaces as whitespace in some of your statements.

You will need to eliminate the phpstan problems. I think many of those will be solved by adding
```php
use PhpOffice\PhpSpreadsheet\Style\Color;
```
in ConditionalColorScale.php. Then, you need to adjust the types of your parameters and return values, e.g. `setMinimumColor` probably requires a Color parameter, not a string.

Once we have satisfied phpcs and phpstan, your code is still only half there. While you can make a case for adding it only to the Xlsx Writer, you can make a much stronger case if you add corresponding support to the Xlsx Reader as well.

Then, once that is taken care of, we will need to add some unit tests to demonstrate that your code is working correctly.

oleibman:
No concerns with Scrutinizer "complexity" warning.

oleibman:

PR #3740, opened 2023-09-19
Recent Scrutinizer Messages

oleibman:
Get rid of them where possible.

This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -3442,7 +3442,7 @@ public function getHashCode()
      * @param string $range Range to extract title from
      * @param bool $returnRange Return range? (see example)
      *
-     * @return ($range is non-empty-string ? ($returnRange is true ? array{0: string, 1: string} : string) : ($returnRange is true ? array{0: null, 1: null} : null))
+     * @return ($range is non-empty-string ? ($returnRange is true ? array{0: string, 1: string} : string) : ($returnRange is true ? array{0: null, 1: null} : null)) Scrutinizer does not understand this but it helps Phpstan
```
PowerKiKi, line 3445:
```suggestion
     * @return ($range is non-empty-string ? ($returnRange is true ? array{0: string, 1: string} : string) : ($returnRange is true ? array{0: null, 1: null} : null))
```

Don't think we should have those kind of comment that end up in end-user documentation.

PowerKiKi (commented):
On a more general note, I feel nowaday PHPStan is _way_ better than Scrutinizer static analyzer and we probably should not spend too much time on accommodating Scrutinizer false positives.

While Scrutinizer is still able to catch things that PHPStan cannot, so it makes sense to keep it, I don't think we should riddle our code with `@scrutinizer ignore-type` or similar. Instead we can ignore whole range of errors directly from Scrutinizer UI. Typically `$drawing->setImageResource(/** @scrutinizer ignore-type */ $gdImage);` is a false positive that clutter our code for nothing.

PowerKiKi:
I gave you access to Scrutinizer:

image

oleibman:
@PowerKiKi converting to draft so that you can carry on with your work in #3743. As for the comments in the doc-block, I couldn't figure out how else to indicate that Scrutinizer cares but we consciously don't care (as opposed to we've never even looked at the message).

I agree that Phpstan is way better than Scrutinizer in its analysis.

PowerKiKi:
> couldn't figure out how else to indicate that Scrutinizer cares but we consciously don't care

Now you should be able to do that on https://scrutinizer-ci.com/g/PHPOffice/PhpSpreadsheet/issues/master

oleibman:
Superseded by #3743.

PR #3743, opened 2023-09-20
Drop all Scrutinizer annotations

PowerKiKi:
This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [ ] Code style is respected

Because they tend to clutter our code a lot and unfortunately, there are lots of false positives. Instead, it would probably be better to deal with false positives out of band, via the Scrutinizer web UI.


oleibman:
@PowerKiKi I don't have a problem with what you're doing. When I started looking at the Scrutinizer messages, there were just too many to cope with, and I needed a way to distinguish the wheat from the chaff. So I added the annotations to get things to a manageable number. Sometimes annotations weren't possible, and I added a comment so that anyone looking at the Scrutinizer reports would know that this is something that someone has looked at and isn't concerned about. The drop7.4 changes added back a lot of messages, so I was planning to repeat the process with #3740. However, I am glad to pause that in favor of something better. I will add more notes in the PR.

PowerKiKi:
@oleibman I'll let you review and comment or fix it. If you feel this is OK, then you can merge it yourself.

You probably want to double-check especially my [second commit](https://github.com/PHPOffice/PhpSpreadsheet/pull/3743/commits/cbffcbcbbce8a1e9c0c930377842a6ce732d8c7c).

oleibman:
@PowerKiKi There seem to be a whole bunch of new false positives. Just to look at one, what do we do with a problem like (in Calculation/Engine/Operands/StructuredReference):
```php
$table = $cell->getWorksheet()->getTableByName($this->tableName);
```
Scrutinizer is absolutely wrong in its analysis, which claims that you've introduced 4 new errors to that program:
```
Are you sure the assignment to $table is correct as
$cell->getWorksheet()->getTableByName($this->tableName)
targeting PhpOffice\PhpSpreadsheet\Worksheet\Worksheet::getTableByName()
seems to always return null.
```
You didn't touch this code, nor the signature for Worksheet/getTableByName (returns `?Table`). I'm not sure what has changed to make Scrutinizer come to this obviously incorrect conclusion. Is there anything you can do with your approach to make it shut up about these?

oleibman:
I think maybe I see it now. If I merge this change, I can go to the master report in Scrutinizer and create filters for problems like the one I mentioned above. If that is so, it would definitely be satisfactory. I probably can't get to it till midweek, but I will look forward to trying it out then.

PowerKiKi:
AFAIK filters on Scrutinizer are not scoped by branch. A filter is active for all branches at once. But yes we can and should filter even more rules...

oleibman:
@PowerKiKi Thank you. This is wonderful! I filtered out some of the new issues, and confirmed that they show up as "filtered" rather than "all issues" on a PR that I have yet to merge. So, the "all issues" report can continue to report issues that haven't been analyzed/addressed, and we are nevertheless able to see what we consider to be false positives on the "filtered" report  without polluting the "all issues" report. Not that I begrudge the work on the "tricky" workarounds, but this is a lot better.

PR #3759, opened 2023-10-05
Add stream support to IReader

apreiml:
When using backends like minio or s3 to store files, those files are usually retrieved into memory. This pull request tries to add stream support for IReaders. This will allow to load files in memory using a php://memory stream.

I've only started with changing the interface and adding stream support to csv. I'd like to know, if this is a desirable feature which has the chance of being merged, before I continue with changing the other readers. 

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [ ] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Reader/BaseReader.php
+++ b/src/PhpSpreadsheet/Reader/BaseReader.php
@@ -194,7 +194,6 @@ abstract class BaseReader implements IReader
      */
     protected function openFile($file): void
     {
-        $fileHandle = false;
         if (is_string($file)) {
             $filename = $file;
             File::assertFile($file);
@@ -204,6 +203,8 @@ abstract class BaseReader implements IReader
         } elseif (is_resource($file)) {
             $filename = 'stream';
             $fileHandle = $file;
+        } else {
+            throw new ReaderException('invalid type for file. Only file path or a stream resource is allowed');
         }
         if ($fileHandle === false) {
             throw new ReaderException('Could not open file ' . $filename . ' for reading.');
--- a/src/PhpSpreadsheet/Reader/BaseReader.php
+++ b/src/PhpSpreadsheet/Reader/BaseReader.php
@@ -160,7 +160,10 @@ abstract class BaseReader implements IReader
         }
     }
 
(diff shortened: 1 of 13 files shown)


oleibman:
This has arisen in the past (see issue #1931), but was rejected because it wasn't a good fit for Xlsx (or Ods) files. Which doesn't mean we can't think about it again. But ... your test case does something like:
```php
        $stream = fopen('data://text/plain;base64,' . base64_encode($data), 'rb');
        self::assertNotFalse($stream);
        $reader = new Csv();
        $spreadsheet = $reader->load($stream);
```
But we already have a means to do this without any changes to PhpSpreadsheet:
```php
        $stream = fopen('data://text/plain;base64,' . base64_encode($data), 'rb');
        self::assertNotFalse($stream);
        $reader = new Csv();
        $spreadsheet = $reader->loadSpreadsheetFromString(stream_get_contents($stream));
```
Do you feel that your change adds functionality or ease of use beyond that approach?

apreiml:
@oleibman my plan is trying to replace ZipArchive with ZipStream in the readers, which would support working with streams.

As for the Csv example: A stream may be something else than a string stream. It's just the most convenient one for testing stream support. I'd like to introduce stream support at the IReader interface level, so that it works for all readers with the same api. For CSV and XML it's easier to work from strings, so they basically get the support for little effort.

oleibman:
It sounds like maybe you have an idea on how to solve the streaming problem for Xlsx. That would certainly be useful. Do you think you could attack that first, and leave xml/etc. till later (no need to undo what you've already done for csv)?

One important thing to remember - ZipStream3 supports Php8.1+ 64-bit, but we will support Php8.0 and 32-bit for some time to come, so whatever you come up with should support ZipStream2 if possible (might be willing to throw an exception if not possible, but would have to think about it).

Also, we no longer support Php7, so prefer to define parameter types in code rather than doc-blocks.

oleibman:
Useful start, but no update in 3 years, and too many merge conflicts. Closing.

PR #3776, opened 2023-10-25
Slk Shared Formulas

### LINKED ISSUE #2267: SYLK shared formulae
This is:

```
- [X] a bug report
- [ ] a feature request
```

This SYLK file uses the shared formula optimization that is present in XLS and XLSX:

```sylk
ID;PWXL;N;E
P;PGeneral
F;P0;DG0G10;M320
B;Y3;X1;D0 0 9 0
C;Y1;X1;K1
C;Y2;K2;ER[-1]C+1
C;Y3;K3;S;R2;C1
E
```

Excel understands this as a worksheet with A1 set to 1, A2 set to A1+1 (cached value 2), A3 set to A2+1 (cached value 3)

PHPSpreadsheet incorrectly handles the A3 line, setting A2 to 3, [because of an error in Slk.php](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Reader/Slk.php#L280): 

```php
            switch ($rowDatum[0]) {
                case 'C': // <-- should be removed
                case 'X':
                    $column = substr($rowDatum, 1);

                    break;
                case 'R': // <-- should be removed
                case 'Y':
                    $row = substr($rowDatum, 1);
```

The code appears to assume that `C` and `R` refer to the cell but it in fact refers to the referenced cell.  It will attempt to set A2 to the value 3.

Unfortunately there's no official spec for any of this, but it appears that the `C` record supports the following fields:

- `S`: signals that the cell is using a shared formula 

- `R` and `C`: row/column of the cell with the formula (1-indexed numeric values)

`print_r($sheet->getCell("A2")->getValue());` should print 2

`print_r($sheet->getCell("A2")->getValue());` prints 3

Save the SYLK codeblock to `test.slk` and run

```php
<?php

$spreadsheet = \PhpOffice\PhpSpreadsheet\IOFactory::load("test.slk");
print_r($sheet->getCell("A2")->getValue());
```

All versions of PHPSpreadsheet (error inherited from PHPExcel) and every commit of PHPExcel on github (blamelog points to https://github.com/PHPOffice/PHPExcel/commit/509f27e5c601c002090af18a5dd545b06b7183ca )

oleibman:
Fix #2267. The Slk format has a way to express a "shared formula", but the Slk reader does not yet understand it. Thanks to @SheetJSDev for documenting the problem and pointing the way towards a solution. It has taken a long time to get there. Part of the problem is that I have not been successful in getting Excel to use this type of construction when saving a Slk file. So I have resorted to saving a Slk file where shared formulas *could* be used, and then editing it by hand to actually use them. It would not surprise me in the least to have neglected one or more possible ways to specify a shared formula; but, at least the issue as documented is resolved, and if new issues arise, we'll probably be in better shape to deal with them.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
No concern with Scrutinizer "complexity" message. Scrutinizer also flags one statement; as is often the case, this is a false positive, and will be filtered out from now on.

PR #3791, opened 2023-11-13
Use case insentive comparison to get sheet name

kpn13:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

For now, the method which check if a sheetname already exists, uses a case-sensitive comparison which can cause issue as Excel doesn't allow to have 2 sheets with the same name and uses a case-insensitive comparison. So, I propose to respect the same rule.

### DIFF
--- a/src/PhpSpreadsheet/Spreadsheet.php
+++ b/src/PhpSpreadsheet/Spreadsheet.php
@@ -701,7 +701,7 @@ class Spreadsheet implements JsonSerializable
     {
         $worksheetCount = count($this->workSheetCollection);
         for ($i = 0; $i < $worksheetCount; ++$i) {
-            if ($this->workSheetCollection[$i]->getTitle() === trim($worksheetName, "'")) {
+            if (strcasecmp($this->workSheetCollection[$i]->getTitle(), trim($worksheetName, "'")) === 0) {
                 return $this->workSheetCollection[$i];
             }
         }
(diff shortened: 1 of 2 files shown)


oleibman:
Please add `expectExceptionMessage` to your test (see `testNotSerializable` in the member you're changing for an example). This helps ensure that the test fails for the "right" reason. I won't insist on it, but it would be helpful if you could do the same for the other tests in this member that call `expectException` without calling `expectExceptionMessage`.

kpn13:
Thank you for the feedback, it's done!

oleibman:
I have found a situation which your change does not cover. I need to think about it a little more.
```php
$sheet1 = $spreadsheet->getActiveSheet();
$sheet1->setTitle('firstsheet');
$sheet2 = $spreadsheet->createSheet();
$sheet2->setTitle('Firstsheet', true, false);
```
If the third parameter to setTitle is true (default), PhpSpreadsheet will append a number to the sheet name until it finds a unique name. However, if it is false, it won't do that, and so this code will create a corrupt spreadsheet. I do not understand the use case for that third parameter. In the test suite, it is used only in WorksheetTest; I would be surprised if it was ever used "in the wild", and I imagine "caveat emptor" should apply to anyone who uses it. So, I don't think you should do anything about it. But, the first statement in Worksheet::setTitle:
```php
        // Is this a 'rename' or not?
        if ($this->getTitle() == $title) {
            return $this;
        }
```
For consistency with your change, that should probably be a case-insensitive compare as well. Can you make that change and add an appropriate test in WorksheetTest::testSetTitleDuplicate?

oleibman:
No, it turns out that Excel allows you to change the sheet name in this manner (and adjusts defined names etc. accordingly). So the change I suggested to setTitle should not be made after all. Sorry if you wasted any time on it.

oleibman:

kpn13:
Thank you for your work!

PR #3800, opened 2023-11-28
Chart Dynamic Title and Special Font Properties

### LINKED ISSUE #3797: Chart title loses reference to cell
This is:

```
- [x] a bug report
```

If I have a graph title that relates to a cell, this should stay as is. I.e. the graph title's value is "=A1", deriving the value from Cell A1.

When the graph has gone through the reader and writer, the output does have the correct data, i.e. the value of A1, but it is not a reference to A1 anymore.

```php
<?php

$template_file = 'test-rotated_xlabel.xlsx';
$filename = 'test_' . date( 'Y-m-d_H-i-s' ) . '.xlsx';

// Load the template file
$reader = IOFactory::createReader('Xlsx');
$reader->setIncludeCharts(TRUE);

$spreadsheet = $reader->load($template_file);

// Set the headers to force a download of the Excel file
header( 'Content-Type: application/vnd.ms-excel' );
header( 'Content-Disposition: attachment; filename="' . $filename . '"' );
header( 'Cache-Control: max-age=0' );

// Create a writer and send the Excel file to the browser

ob_end_clean();
$writer->setIncludeCharts(TRUE);
$writer->save( 'php://output' );
exit;
```

Here is the sample Excel file that I use: 
[test-rotated_xlabel.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/13452598/test-rotated_xlabel.xlsx)

- [x] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [x] Charts
- [ ] AutoFilter
- [ ] Form Elements

PHP 8.1

oleibman:
Fix #3797. Excel allows a Chart Title to be a formula, albeit a very rigidly limited one. It can only be a reference to a single cell, and the worksheet name must be specified, and the column and row must be absolute. Methods are added to Chart/Title to accommodate this (and styling for it). This will be handled for input/output for Xlsx, and for output for Html.

The sample file which was submitted with this issue demonstrated that something else was missing. When setting the font for a chart title in Excel, you can specify all-caps or small-caps, options not available for most cell formatting. These are now added.

The sample file also fell into the category of spreadsheets which lose one or more charts when converted to Html. I have redone the "extend rows and charts" logic in Html Writer. It is now clearer (I hope) and more efficient, and hopefully this problem will not arise again.

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

  - [x] New unit tests and samples have been added
- [x] Code style is respected


oleibman:
Not concerned about Scrutinizer complexity complaint.

PR #3808, opened 2023-12-02
Let Phpstan Run on Samples

oleibman:
Phpstan currently analyzes all source and test members. We already run phpcs and php-cs-fixer on samples as well. I would expect that samples are often used as templates for code in userland; it behooves us to be at least as careful with those members as for the others which are already being analyzed.  Aside from 1300+ messages `Variable $helper might not be defined.`, which will be suppressed in phpstan.neon.dist, there are really only a few changes needed for sample members, so that part of the code base was already in good shape, and is now even better. No annotations were needed.

This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
- [x] code quality

Checklist:

- [x] Code style is respected

### DIFF
--- a/phpstan.neon.dist
+++ b/phpstan.neon.dist
@@ -8,6 +8,7 @@ parameters:
     paths:
         - src/
         - tests/
+        - samples/
     excludePaths:
         - src/PhpSpreadsheet/Chart/Renderer/JpGraph.php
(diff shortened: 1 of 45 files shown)


oleibman:
No concern about Scrutinizer complexity message.

PR #3822, opened 2023-12-12
Configure Read the Docs

PowerKiKi:
In response to https://github.com/PHPOffice/PhpSpreadsheet/pull/3793#issuecomment-1850882630

readthedocs.org build every commit pushed on `master`. However builds have been failing for the past two months, because they now require a config file. I assumed that config file was added, but I was wrong. So I added it.


PowerKiKi:
@oleibman this should answer your question.

I also enabled builds on pull-requests, so we can preview doc changes, like so:

image

PR #3834, opened 2023-12-19
Reduce memory consumption of Worksheet::rangeToArray() when many empty rows are read

### LINKED ISSUE #3814: Allowed memory size of X bytes exhausted in PhpSpreadsheet\Worksheheet\Worksheet.php on line 3108
This is:

```
- [X] a bug report
- [ ] a feature request
```

The spreadsheet array containing the textual data from the spreadsheet is initialized.

Fatal error: Allowed memory size of 536870912 bytes exhausted (tried to allocate 36864 bytes) in PhpSpreadsheet\Worksheheet\Worksheet.php on line 3108

```php
<?php

require_once './vendor/autoload.php';

$spreadsheet = \PhpOffice\PhpSpreadsheet\IOFactory::load('./example.xlsx');

$names = $spreadsheet->getSheetNames();
$result = [];
for($i=0;$i<$spreadsheet->getSheetCount();$i++)
{
    $result[$names[$i]] = $spreadsheet->getSheet($i)->toArray();
}

print_r($result);
```
If this is an issue with reading a specific spreadsheet file, then it may be appropriate to provide a sample file that demonstrates the problem; but please keep it as small as possible, and sanitize any confidential information before uploading. File attached below.

- [X] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

This happens for a specific file, I am attaching it here: [example.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/13600790/example.xlsx)

I have tested with 1.29 on PHP 7.4 (BTW, I thought 7.4 is no longer supported?) and PHP 8.2 - happens on both.

I've looked into `rangeToArray` function a little bit, this file has range of `A1:AMJ1048576`, and when we are looping 
`for ($col = $minCol; $col !== $maxCol; ++$col) {` we never reach `$maxCol` value of `AMK`

I understand the file is crazy, yet I believe Reader should not go into infinite loop even for a crazy file like this.

sj-i:
This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
- [x] performance improvement

Checklist:

- [x] Code style is respected

There may be cases where some data is located "far away" in a sheet, like `AMJ1048576`, and the rest of the rows are mostly empty. This change dramatically reduces the memory usage in such cases. For example, currently using `Worksheet::toArray()` with `example.xlsx` in #3814 eats up over 20GB of memory in PHP 8.2. After this change, the memory consumption becomes less than 32MB.

So, this fixes #3814

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 - COUNTIFS Does Not Require xlfn. [Issue #3819](https://github.com/PHPOffice/PhpSpreadsheet/issues/3819) [PR #3827](https://github.com/PHPOffice/PhpSpreadsheet/pull/3827)
 - Strip `xlfn.` and `xlws.` from Formula Translations. [Issue #3819](https://github.com/PHPOffice/PhpSpreadsheet/issues/3819) [PR #3828](https://github.com/PHPOffice/PhpSpreadsheet/pull/3828)
 - Recurse directories searching for font file. [Issue #2809](https://github.com/PHPOffice/PhpSpreadsheet/issues/2809) [PR #3830](https://github.com/PHPOffice/PhpSpreadsheet/pull/3830)
+- Reduce memory consumption of Worksheet::rangeToArray() when many empty rows are read. [Issue #3814](https://github.com/PHPOffice/PhpSpreadsheet/pull/3814) [PR #3834](https://github.com/PHPOffice/PhpSpreadsheet/pull/3834)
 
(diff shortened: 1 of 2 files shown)


oleibman:
@sj-i This is very impressive analysis and an ingenious solution. I need to study it a little more to understand it better, but my intention is to merge it within the next day or two.

oleibman:
Thank you for your contribution, and for the lesson about CoW.

PR #3838, opened 2023-12-20
handle the case where the input in not an int or a float

kheiroguenoune:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Shared/Date.php
+++ b/src/PhpSpreadsheet/Shared/Date.php
@@ -220,17 +220,18 @@ class Date
         } else {
             $baseDate = new DateTime('1899-12-30', $timeZone);
         }
-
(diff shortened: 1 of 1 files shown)


oleibman:
Aside from fixing the phpstan, phpcs, and php-cs-fixer problems:

If there is a problem which this PR solves, please open an issue so that we can track it. Also, please add a test case which fails without your PR but succeeds with it.

oleibman:
Closing. Over a year without further documentation or test case. In addition, typing has been added to the function in question - the parameter must be int or float, so no further test is needed; and the result must always be a DateTime object, so the suggested solution which returns null is not permissible.

PR #3842, opened 2023-12-26
fix: IReader returning type

k0mar12:
This is:

- [Y] a bugfix
- [N] a new feature
- [N] refactoring
- [N] additional unit tests

Checklist:

- [ ] Code style is respected

*BaseReader* obviously returning *Spreadsheet* in *load* method, but IReader doesn't specify that.

### DIFF
--- a/src/PhpSpreadsheet/Reader/IReader.php
+++ b/src/PhpSpreadsheet/Reader/IReader.php
@@ -141,5 +141,5 @@ interface IReader
      *
      * @return Spreadsheet
      */
-    public function load(string $filename, int $flags = 0);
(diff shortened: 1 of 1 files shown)


oleibman:
I don't really have a problem with moving this change forward, but I'd like a little more information. I agree that it looks a little fishy that the BaseReader and IReader signatures don't match. But I don't see this causing a problem anywhere, even when we run the unit tests against the bleeding-edge nightly Php 8.4 build. What is the problem you are seeing that makes this change necessary, or, at least, desirable?

PowerKiKi:
Was fixed in c303df24081430dccdcdd294a50a4330a345b512

PR #3906, opened 2024-02-17
Improve Performance of Csv Writer

### LINKED ISSUE #3904: Writing a CSV file with many lines noticeable slower than before
This is:

```
- [x] a bug report
- [ ] a feature request
```

Writing a CSV file with many lines was generally fast in 1.29.0, as expected.

Writing a file with many lines is noticeable slower in 2.0.0 than 1.29.0. I have pinpointed the issue to commit 096e193495754784cee5d9212728a0f366dc4f75 because the earlier commit 9f87d7c8e6139fb3117b04372e7fad409f5e647e before that is fast. The culprit seems to be \PhpOffice\PhpSpreadsheet\Worksheet\namedRangeToArray(). The writing time seems to increase more and more when more lines are added. This can be tested by writing 10k lines vs. 20k lines where the difference is already big.

```php
<?php

$rows = [];

// Add many lines to one column only
for ($i = 0; $i < 50000; $i++) {
    $rows[] = ['A' . $i + 1];
}

$writer = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($spreadsheet, \PhpOffice\PhpSpreadsheet\IOFactory::WRITER_CSV);

$worksheet->fromArray($rows);

$start = \microtime(true);

$writer->save('test.csv');

echo \microtime(true) - $start . "\n";
// 0.71157717704773 in 1.29.0
// 350.71549415588 in 2.0.0
```

- [ ] Reader
- [x] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

I think this only affects CSV. Writing an XLSX file with the same code is still fast.

PhpSpreadsheet 2.0.0, tested with PHP 8.2.13

oleibman:
Fix #3904. PR #3839 provided a huge performance boost for sparsely populated spreadsheets. Unfortunately, it degraded performance for more densely populated spreadsheets when writing Csvs. The reason is that Csv Writer calls toArray for each row, meaning that a lot of the intermediate data used to speed things up needs to be recalculated for every row. It would be better off calling toArray just once for the entire spreadsheet; however this gives back some of the memory improvements of PR #3834. However, the memory effects can be substantially alleviated by supplying a Generator function to do the work. This PR does that; the result is that Csv Writer is now quite a bit faster, and with only a small memory uptick, vs. its performance in PhpSpreadsheet 1.29.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,9 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 ### Fixed
 
 - IF Empty Arguments. [Issue #3875](https://github.com/PHPOffice/PhpSpreadsheet/issues/3875) [Issue #2146](https://github.com/PHPOffice/PhpSpreadsheet/issues/2146) [PR #3879](https://github.com/PHPOffice/PhpSpreadsheet/pull/3879)
+- Changes to floating point in Php8.4. [Issue #3896](https://github.com/PHPOffice/PhpSpreadsheet/issues/3896) [PR #3897](https://github.com/PHPOffice/PhpSpreadsheet/pull/3897)
+- Handling User-supplied Decimal and Thousands Separators. [Issue #3900](https://github.com/PHPOffice/PhpSpreadsheet/issues/3900) [PR #3903](https://github.com/PHPOffice/PhpSpreadsheet/pull/3903)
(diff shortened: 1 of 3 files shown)


oleibman:
No concerns about Scrutinizer "complexity" complaint.

yakobe:
Hi @oleibman!

Thank you for this fix. I will try it out by pinning to the commit. Any ideas when the next release will be?

PR #3924, opened 2024-02-29
Default Style Alignment

### LINKED ISSUE #3918: Default style alignment does not work for Xlsx Writer
This is:

```
- [x] a bug report
- [ ] a feature request
```

Default alignment style should be applied with `Xlsx` writer.

Default alignment style is not applied.

```php
<?php

$spreadsheetStyle = $spreadsheet->getDefaultStyle();
$spreadsheetStyle->getAlignment()
    ->setWrapText(true)
    ->setHorizontal(\PhpOffice\PhpSpreadsheet\Style\Alignment::HORIZONTAL_CENTER)
    ->setVertical(\PhpOffice\PhpSpreadsheet\Style\Alignment::VERTICAL_CENTER);

$cell = $worksheet->getCell('A1');
$cell->setValue('aaaaaaaaaaaaaa');

// $xlsx = new \PhpOffice\PhpSpreadsheet\Writer\Xls($spreadsheet); // Working
$xlsx = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet); // Not Working

$xlsx->save('alignment.xls'); // Optionally update with `.xls`
```

- [ ] Reader
- [x] Writer
- [x] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

It affects `Xlsx` writer but not `Xls` writer. I did not test for other formats

At least v2.0.0 (the latest at this time): https://github.com/PHPOffice/PhpSpreadsheet/releases/tag/2.0.0

### References

- Potentially the same issue: https://github.com/PHPOffice/PhpSpreadsheet/issues/3456
- The fix used in #3456: https://github.com/PHPOffice/PhpSpreadsheet/pull/3459 (Tit does not look like this can be straightforward ported to v2)
- Working file: [alignment.xls](https://github.com/PHPOffice/PhpSpreadsheet/files/14396739/alignment.xls)
- Not Working file: [alignment.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/14396742/alignment.xlsx)

oleibman:
Fix #3918, sort of. Xlsx cells use the default style when they omit the `s` tag, or when `s="0"` is specified. LibreOffice does not honor Alignment in the default Style unless the cell explicitly uses the second form, even though it honors other default Styles (e.g. bold font) when `s` is omitted. Gnumeric seems to have the same problem. A bug report has been filed with LibreOffice.

In the meantime, this PR adds to Xlsx Writer an optional boolean property `explicitStyle0` with setter and getter. Default is false, which will continue the current behavior by adding an `s` tag only when the cell uses a non-default style (this is how Excel itself behaves). When set to true, Xlsx Writer will explicity write `s="0"` for all cells using default style. This will allow users to create an Xlsx spreadsheet with default alignment of cells that will show up correctly when the spreadsheet is viewed with LibreOffice. Technically speaking, it is *probably* safe to always use `true`, except that the spreadsheet size will be a bit larger. However, my hope is that this is a temporary measure which can go away when the vendors have had a chance to fix their problems, hence the `false` default.

This is:

- [x] a bugfix (workaround for vendor problem)
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
A win for Scrutinizer! I had miscoded some tests, which resulted in a Scrutinizer warning, and which should have resulted in a test failure but didn't. We have always cautioned that it is dangerous to store a Cell in a variable. It appears that the same is true for Style. The problem shows up in the following code adapted from my test:
```php
        $styleA1 = $sheet2->getStyle('A1');
        var_dump($styleA1->getFont()->getItalic());  // false, because A1 is not italic
        $styleA2 = $sheet2->getCell('A2')->getStyle();
        var_dump($styleA1->getFont()->getItalic()); // true, because A2 is talic
```
I will correct the test, and attempt to update the documentation.

oleibman:
No concerns with Scrutinizer "complexity" complaint.

PR #3946, opened 2024-03-12
Fix default value for Conditional::$text

Phenix789:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

[`Worksheet:640`](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Writer/Xlsx/Worksheet.php#L640) call `Conditional::getText()`, but if the `Conditional::$text` field is not initialized with a value by calling `getText()` PHP throw a typed error.  
The fix is just to set a default value to empty string to avoid error by calling `getText()` uninitialized.  
In addition `Worksheet` call `getText()` to check if the value is not empty, so empty string is already handled.

### DIFF
--- a/src/PhpSpreadsheet/Style/Conditional.php
+++ b/src/PhpSpreadsheet/Style/Conditional.php
@@ -84,7 +84,7 @@ class Conditional implements IComparable
     /**
      * Text.
      */
-    private string $text;
+    private string $text = '';
 
     /**
      * Stop on this condition, if it matches.
--- a/tests/PhpSpreadsheetTests/Style/ConditionalFormatting/PR3946Test.php
+++ b/tests/PhpSpreadsheetTests/Style/ConditionalFormatting/PR3946Test.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpOffice\PhpSpreadsheetTests\Style\ConditionalFormatting;
+
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Style\Conditional;
+use PhpOffice\PhpSpreadsheet\Style\Fill;
+use PhpOffice\PhpSpreadsheet\Style\Style;
+use PhpOffice\PhpSpreadsheet\Writer\Xlsx as XlsxWriter;
+use PHPUnit\Framework\TestCase;
+
+class PR3946Test extends TestCase
+{
+    public function testConditionalTextInitialized(): void
+    {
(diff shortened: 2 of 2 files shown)


oleibman:
The commentary to your PR says that you are making 2 changes, but I see only 1. Should there be another?

Please add a unit test case which would have failed before your change but which will succeed with it.

Phenix789:
I update PR description and yes it's only 1 fix.  
But it's hard to write a test for this case, it just adding a default value for a property to avoid uninitialized error.

---

Furthermore, after several investigations, it appears that `Conditional::$text` is required for the "text" types, and using a `Wizard` seems more appropriate than directly using `new Conditional()`.  
My fix seems only to hide the error.  
So, I think a will close this PR.  
However, having a PHP error might not be the most understandable error either, right ?  
What could we do ? Any suggestion ?  
Thanks

oleibman:
Your test should be something like the following:
```php
use PhpOffice\PhpSpreadsheet\Style\Conditional;
use PhpOffice\PhpSpreadsheet\Style\Fill;
use PhpOffice\PhpSpreadsheet\Style\Style;
use PhpOffice\PhpSpreadsheet\Writer\Xlsx as XlsxWriter;

        $cellRange = 'C3:E5';
        $style = new Style();
        $style->applyFromArray([
            'fill' => [
                'color' => ['argb' => 'FFFFC000'],
                'fillType' => Fill::FILL_SOLID,
            ],
        ]);

        $condition = new Conditional();
        $condition->setConditionType(Conditional::CONDITION_CONTAINSTEXT);
        $condition->setOperatorType(Conditional::OPERATOR_CONTAINSTEXT);
        $condition->setStyle($style);
        $sheet->setConditionalStyles($cellRange, [$condition]);

        $writer = new XlsxWriter($spreadsheet);

        $writerWorksheet = new XlsxWriter\Worksheet($writer);
        $data = $writerWorksheet->writeWorksheet($sheet, []);
        self::assertStringContainsString('<conditionalFormatting sqref="C3:E5">', $data);
```
This code will error out with the current version of PhpSpreadsheet because of the error you're reporting, but will succeed after your change is made, which is exactly what we are looking for in a test. Let me know if you need help making the code above into a formal unit test.

oleibman:

Phenix789:
Thanks for the merge 👍.  
And sorry for neglecting this PR.  
I've been quite busy these past few weeks, but I was planning to finish it soon.

PR #3957, opened 2024-03-20
Html Writer Comments - Breaking Change

### LINKED ISSUE #3954: Make voku packages optional
This is:

```
- [ ] a bug report
- [ x ] a feature request
```

PhpSpreadsheet is about handling spreadsheets like Excel and LibreCalc files. We don't use the HTML support part of the package at all. While it is great that there is a HTML support, it should be optional.

Move the voku package to the "suggested" part of the composer.json.

The reason for this being a problem is that it in turn gets "voku/portable-utf8" which has a bootstrap.php running a ini_set(). We don't want a open source package to run ini_set on our application without our knowledge. voku/portable-utf8 did take down our system some time ago, so we have a company restriction against using that package. It is too bad if we cannot use phpspreadsheet because of a HTML support that we don't really use.

Yes, it affect all formats, but it should only affect the HTML format.

2.*

oleibman:
Fix #3954. In response to a Security Incident, a package was added to the project to sanitize the Html for comments attached to a cell. We have tried at least 2 different packages for this purpose, and users have raised legitimate concerns about both.

I believe that adding a sanitizer package, although it addresses the problem, was overkill. Cell comments are a RichText object, and Html Writer already handled RichText cell *values*. Values did not figure in the Incident, and they surely would have done so if they were a problem because it is a lot easier to set up test cases for values than for comments. RichText values were not a problem because they were sanitized with `htmlspecialchars`; if comments were to use the same code that RichText was already using, it would likewise be safely sanitized. As an added bonus, the existing code for comments only uses the plaintext value, but the values code would allow the comments to be styled, just as they are for Xlsx.

This is a breaking change - I don't think it will affect a lot of users, but there may be some. It is worth noting that the comment block before function `writeComment` has a link explaining what is being done. That link mentions only styling elements, not other possibilities like hyperlinks. At any rate, if people are putting styling tags (e.g. `<b>`) in their comments for this purpose, those will no longer work; the styling needs to be applied to the RichText elements. That will keep the user code the same regardless of the format of the intended output, which is certainly a good thing, but it is a break. If people are trying to insert non-styling tage (e.g. `<a>`), those will no longer work, just as there is no way to do it for Xlsx (although Excel itself may convert the raw text to a hyperlink, but that's out of scope, at least for now). I also note that, even with the current code, I haven't yet found a way to keep the Html comment in place long enough to actually click on any hyperlinks.

The main package being dropped brings several other packages along with it, so the project as a whole will have a slightly lighter footprint than before.

The existing `XssVulnerability` test cases are all preserved. Although the final result of the sanitizing changes, it can easily be seen that the results are all harmless.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## Unreleased - TBD
 
+### BREAKING CHANGE
+
+- Writing of cell comments to Html will now sanitize all Html tags within the comment, so the tags will be rendered as plaintext and have no other effects when rendered. Styling can be achieved by using the Font property of of the TextRuns which make up the comment, as is already the cases for Xlsx. [PR #3957](https://github.com/PHPOffice/PhpSpreadsheet/pull/3957)
+
 ### Added
 
 - Default Style Alignment Property (workaround for bug in non-Excel spreadsheet apps) [Issue #3918](https://github.com/PHPOffice/PhpSpreadsheet/issues/3918) [PR #3924](https://github.com/PHPOffice/PhpSpreadsheet/pull/3924)
--- a/composer.json
+++ b/composer.json
@@ -84,8 +84,7 @@
         "markbaker/matrix": "^3.0",
         "psr/http-client": "^1.0",
         "psr/http-factory": "^1.0",
-        "psr/simple-cache": "^1.0 || ^2.0 || ^3.0",
-        "voku/anti-xss": "^4.1"
+        "psr/simple-cache": "^1.0 || ^2.0 || ^3.0"
     },
     "require-dev": {
         "dealerdirect/phpcodesniffer-composer-installer": "dev-main",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "84f741787fe1d67df41a23c583e92220",
+    "content-hash": "c7a7ba2e1cc9ac479e06ab28916c6b80",
     "packages": [
         {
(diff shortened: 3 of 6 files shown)


oleibman:
@MarkBaker @PowerKiKi Due to the security-related nature of the original problem (and because it's a breaking change), I do not want to move ahead with this without giving you the opportunity to review it. So I will keep it in draft status for a while - please let me know what you think. Also, any idea why the Doc Build failed? This is the first problem I've seen with that since it was introduced, and there aren't even any doc changes in the PR. I assume it's transient, but, just in case you see somthing ...

oleibman:
Doc Build succeeded this time - as suspected, whatever the error was was transient.

oleibman:
I have taken the request out of draft status. I will be away for a while. I will not merge it before April 18.

PowerKiKi:
Thanks for that, and sorry I didn't gave a feedback earlier, even so I did had a look and it seems like a reasonable things to do.

cmanley:
@oleibman thanks for your commit. I'm glad you removed those voku dependencies. It's been wreaking havoc for me. I hope your commit ends up in a release soon. For now I must decide between falling back to the latest 1.* version or using a commit hash in composer.json.

PowerKiKi:
Don't downgrade, I'll release a new version shortly.

PowerKiKi:
https://github.com/PHPOffice/PhpSpreadsheet/releases/tag/2.1.0

oleibman:
@cmanley I think Csv Reader does clean up after itself. It can change a setting in `setAutoDetect`. But that method is called only at the beginning of `loadStringOrFile`, and the old value is restored at the end of that method, and there are no premature returns in between. If you are still concerned about it, you can call `setTestAutoDetect(false)` before attempting to load a Csv; this will eliminate the ini_set altogether.

PR #3976, opened 2024-04-06
fix: Handle `null` value for `$plotGroupingType` in `writePlotGroup`

### LINKED ISSUE #3975: Uncaught TypeError: PhpOffice\PhpSpreadsheet\Writer\Xlsx\Chart::writePlotGroup(): Argument #6 ($plotGroupingType) must be of type string, null given
This is:

- [x] a bug report

PhpSpreadsheet sould not fail loading and writing this file

PhpSpreasheet throws a TypeError when writing the file

(for now I do not have a reproduction without an excel file)

```php
<?php

use PhpOffice\PhpSpreadsheet\Reader\IReader;
use PhpOffice\PhpSpreadsheet\Reader\Xlsx as XlsxReader;
use PhpOffice\PhpSpreadsheet\Writer\IWriter;
use PhpOffice\PhpSpreadsheet\Writer\Xlsx as XlsxWriter;

require_once "vendor/autoload.php";

$reader = new XlsxReader();
$spreadsheet = $reader->load("example_plot_grouping_type_null.xlsx", IReader::LOAD_WITH_CHARTS);

$writer = new XlsxWriter($spreadsheet);
$writer->save("example_updated.xlsx", IWriter::SAVE_WITH_CHARTS);
```

Here is the file: 
[example_plot_grouping_type_null.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/14893580/example_plot_grouping_type_null.xlsx)

- [ ] Reader
- [x] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [x] Charts
- [ ] AutoFilter
- [ ] Form Elements

I'm not sure

Latest version of PhpSpreadsheet (2.0.0) with PHP 8.1.2-1ubuntu2.14 (but should be any supported php version.

homersimpsons:
This is:

- [x] a bugfix
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Fixes #3975


homersimpsons:
Note that I am not aware enough about the internals to create a concise enough unit test (without relying on an external file), but I would happily use some guidance.

oleibman:
Sometimes our tests do need to rely on an external file. Do you have an input file which demonstrates the problem when you load and then try to save it? If so, we can use that to construct a unit test.

homersimpsons:
> Sometimes our tests do need to rely on an external file. Do you have an input file which demonstrates the problem when you load and then try to save it? If so, we can use that to construct a unit test.

Yes I have such file in #3975: https://github.com/PHPOffice/PhpSpreadsheet/files/14893580/example_plot_grouping_type_null.xlsx

oleibman:
Thank you for the sample file. You are correct that your fix eliminates the exception; however, the file that is produced is corrupt (error message when Excel opens it). So your change needs something more. I will take a look, but it may take several days before I can get to it.

homersimpsons:
> however, the file that is produced is corrupt (error message when Excel opens it).

Okay, so it looks like google sheet has an issue then.

> So your change needs something more. I will take a look, but it may take several days before I can get to it.

No worries, I found this bug while investigating another one so I'm not facing it. I put some details in https://github.com/PHPOffice/PhpSpreadsheet/issues/3975 about my analysis of the issue.

PR #3986, opened 2024-04-16
Update StringValueBinder.php

marlocorridor:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Fix check for Formula data type because expression is intended to check if `convertFormula` is set to `true`.


marlocorridor:
for issue #3987 3987

oleibman:
StringValueBinder is working as expected. See https://phpspreadsheet.readthedocs.io/en/latest/topics/accessing-cells/#using-value-binders-to-facilitate-data-entry. In particular, "By default, the StringValueBinder will cast any datatype passed to it into a string. However, there are a number of settings which allow you to specify that certain datatypes shouldn't be cast to strings, but left "as is":". If you want formulas to *not* be converted to strings, you must specify
```php
$stringValueBinder->setFormulaConversion(false)
```

marlocorridor:
I thought that default of `StringValueBinder` is to format string values to their type.
Now I learned that this is to show their "String" value by default.

Thanks!

PR #4006, opened 2024-05-01
RTL Text Alignment in Xlsx Comment

oleibman:
Fix #4004. RTL text can be included in a comment, and will display correctly, but the comment as a whole will be left-aligned. There already exists an `alignment` property for Comment, but it is unused. This PR will allow that property to be set, and to be read from and written to an Xlsx spreadsheet when possible.

The Xml tags that govern this property are found, unusually, in a drawing Vml file. The important property is `x:TextHAlign`, which can be set to Left, Right, Center, Justified, or Distributed. There are other tags which seemed like they were relevant to this problem, but I don't think they actually matter:
```xml
  <v:textbox style='mso-direction-alt:auto'>
   <div style='text-align:right;direction:rtl'></div>
  </v:textbox>
```

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


berberon:
Hi

Thanks for taking care of this.
When will i be able to use compuser to DL this change?

PR #4016, opened 2024-05-05
Better Definitions for Mixed Parameters and Values Part 1 of Many

oleibman:
Improve documentation within program by making explicit what types of values are allowed for variables described as "mixed". In order to avoid broken functionality, this is done mainly through doc-blocks. This will get us closer to Phpstan Level 9, but many changes will be needed before we can consider that.

Executable code change is to add 2 new methods to Cell - `getValueString` and `getCalculatedValueString`, which are the same as the methods without the String suffix, except that they always return a string result. These are invoked at several places in the code where it makes sense, in particular by Writer code.

This is:

- [ ] a bugfix
- [x] improved API documentation
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Scrutinizer issue is because it can't parse one complex array definition. PhpDocumentor and PhpStan have no problem with it. Marked issue as 'ignored' in Scrutinizer.

PR #4019, opened 2024-05-06
Html Reader Non-UTF8 Charsets

### LINKED ISSUE #866: Loading UTF8 content is not possible with HTML reader class
This is:

```
- [x] a bug report
- [ ] a feature request
```

UTF8 HTML content loaded into the spreadsheet.

Failure to load because the validation of tags cannot be completed.
startsWithTag needs to check for BOM when looking for first character in document.

Use Simple example 46. Open template file and add BOM. Retry example.

"phpoffice/phpspreadsheet": "1.6.0"
 "php": ">=7.1.20"

Suggested simple solution in startsWithTag :
Make function not static and add 
```php
if($this->inputEncoding == 'UTF-8') {
    if(substr($data, 0, 3) === chr(0xEF) . chr(0xBB) . chr(0xBF)) {
      $data = substr($data, 3);
     }
}
```

### LINKED ISSUE #1681: Bug in function XmlScanner::toUtf8 -> Suspicious Double-encoded XML
This is:

```
- [ x ] a bug report
- [ ] a feature request
```

A file encoded with something different than UTF-8 can be imported

When we have an XML file that has the "encoding=..." the function in XmlScanner::toUtf8 will always throw an exception.

The call the mb_convert_encoding() do not modify the XML to change the "encoding=...". 
As of this, when you check again if the charset is now UTF-8 the second time, it's still false as it's based on the preg_match.

\>= 1.12

### LINKED ISSUE #3995: Checking and converting loaded content in HTML Reader based on charset
This is:

```
- [x] a bug report
- [x] a feature request
```

That an HTML with a non "UTF-8" charset is converted into UTF-8 before being loaded.

Inside `PhpOffice\PhpSpreadsheet\Reader\Html::loadIntoExisting` it fails to load as the `preg_replace_callback` step throws an error due to the non "UTF-8" encoding of `$convert`.

Create a HTML file with `iso-8859-1` encoding. And try to read it using the HTML Reader.

Reader.

This only affects the HTML Reader.

PHP 8.2 and 8.3 and PhpSpreadshee 1.29 and 2.0.

### Proposed change to code to resolve this issue
I have knocked up a fix for this but I'm not sure if it should be more complex or in it's own function. I based this on the code that is in `PhpOffice\PhpSpreadsheet\Reader\Security\XmlScanner::toUtf8` function. I don't think it belongs in there as it is to do with HTML methods of declaring encodings but I'm happy to be corrected:

```html
<meta http-equiv="Content-Type" content="text/html; charset=iso-8859-1" />

<meta charset="iso-8859-1">
```

```php
    /**
     * Loads PhpSpreadsheet from file into PhpSpreadsheet instance.
     */
    public function loadIntoExisting(string $filename, Spreadsheet $spreadsheet): Spreadsheet
    {
        // Validate
        if (!$this->canRead($filename)) {
            throw new Exception($filename . ' is an Invalid HTML file.');
        }

        // Create a new DOM object
        $dom = new DOMDocument();

        // Reload the HTML file into the DOM object
        try {
            $convert = $this->getSecurityScannerOrThrow()->scanFile($filename);

            // START OF NEW CODE

            // Check for non-"UTF-8" charset
            $pattern = '/charset="?(.*?)("|;)/';
            $result = preg_match($pattern, $convert, $matches);
            if ($result) {
                $charset = strtoupper($matches[1]);
                $convert = mb_convert_encoding($convert, 'UTF-8', $charset);
                $convert = is_string($convert) ? $convert : '';
            }

            // END OF NEW CODE

            $lowend = "\u{80}";
            $highend = "\u{10ffff}";
            $regexp = "/[$lowend-$highend]/u";
            /** @var callable $callback */
            $callback = [self::class, 'replaceNonAscii'];
            $convert = preg_replace_callback($regexp, $callback, $convert);
            $loaded = ($convert === null) ? false : $dom->loadHTML($convert);
        } catch (Throwable $e) {
            $loaded = false;
        }
        if ($loaded === false) {
            throw new Exception('Failed to load ' . $filename . ' as a DOM Document', 0, $e ?? null);
        }
        self::loadProperties($dom, $spreadsheet);

        return $this->loadDocument($dom, $spreadsheet);
    }
```

oleibman:
Fix #3995. Fix #866. Fix #1681. Php DOM loadhtml defaults to character set ISO-8859-1, but our data is UTF-8. So Html Reader alters its html so that loadhtml will not misinterpret characters outside the ASCII range. This works for UTF-8, but breaks other charsets. However, loadhtml uses the correct non-default charset when charset is specified in a meta tag, or when the html starts with a BOM. So, it is sufficient for us to alter the non-ASCII characters only when (a) the data does not start with a BOM, and (b) there is no charset tag.

This will allow us to use:
- UTF-8 files or snippets without BOM, with or without charset
- UTF-8 files with BOM (charset should not be specified and will be ignored if it is)
- UTF-16 files with BOM (charset should not be specified and will be ignored if it is)
- all charsets which are ASCII-compatible for 0x00-0x7f when the charset is declared. This applies to ASCII itself, many Windows and Mac charsets, all of ISO-8859, and most CJK and other-language-specific charsets.

We cannot use:
- UTF-16BE or UTF-16LE declared in a meta tag
- UTF-32, with or without a BOM (browser recommendation is to not support UTF-32, and most browsers do not support it)
- unknown (to loadhtml) or non-ASCII-compatible charsets (EBCDIC?)

I will note that the way I detect the `charset` attribute is imperfect (e.g. might find it in text rather than a meta tag). I think we'd need to write a browser to get it perfect. Anyhow, it is about the same as XmlScanner's attempt to find the `encoding` attribute, and, if it's good enough there, it ought to be good enough here.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Scrutinizer new issues are both false positives, and are now suppressed.

Issue #811, opened 2018-12-12
mime_content_type for CSV can be return 'text/html'


nevigen:
This is:

```
- [x] a bug report
```

When CSV file contains fields with some html tags (for example "&lt;a href=""...""&gt;link&lt;/a&gt;"), mime_content_type return 'text/html', not 'text/plain'. It need to add 'text/html' to array $supportedTypes in \PhpOffice\PhpSpreadsheet\Reader\Csv.php

```
        $type = mime_content_type($pFilename);
        $supportedTypes = [
            'text/html',
            'text/csv',
            'text/plain',
            'inode/x-empty',
        ];

        return in_array($type, $supportedTypes, true);
```

### Which versions of PhpSpreadsheet and PHP are affected?
php 7 and last PhpSpreadsheet

PowerKiKi:
Seems to duplicate #564.

If your file has a `csv` extension it will be read, otherwise we try our best to guess, but it's only that a guess. If you know it's a CSV, then create the reader yourself, instead of relying on `IOFactory` guesswork.

dmitrystas:
File has CSV extension, but mime_content_type return 'text/html'

Issue #564, opened 2018-06-23
Could not open CSV files with "<a href" inside


ltressens:
This is:

```
- [√] a bug report
```

### What is the expected behavior?

CSV file is opened by PhpSpreadsheet

### What is the current behavior?

An Exception is raised

### What are the steps to reproduce?

Try to open a CSV file with "<a href" inside.

Please provide a [Minimal, Complete, and Verifiable example](http://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

require __DIR__ . '/vendor/autoload.php';

$spreadsheet = new \PhpOffice\PhpSpreadsheet\Reader\Csv();
$spreadsheet->setDelimiter(';');
$spreadsheet->setEnclosure('"');
$spreadsheet->load('test.csv');

// test.csv contains :
// "Column A";"<a href=''>Link</a>";"Column C"
```

### Which versions of PhpSpreadsheet and PHP are affected?

PHP 5.6.33 / PhpSpreadsheet 1.2.1
PHP 7.1.16 / PhpSpreadsheet 1.2.1

PowerKiKi:
I just realized you use and outdated version. Upgrade to at least 1.4.0 to get the fix.

PR #4040, opened 2024-05-22
Csv Reader Allow Use of mimetype=text/html Files Without Extension

### LINKED ISSUE #4036: Cannot read csv files that contain '<a href=' and do not have the extension ‘.csv’ ‘.tsv’.
This is:
```
- [x] a bug report
```

I think the problem is similar to #564, but it happens under certain conditions, even in recent versions.
This can be a problem if getting a file from $_FILES['upfile']['tmp_name'] in php, since there is no extension.

* The file extension is not .csv and .tsv.
* It contains “<a href=”.

Can read csv files that satisfy all of the above.

An error occurs when trying to read a csv file that satisfies all of the above.
```
Fatal error: Uncaught PhpOffice\PhpSpreadsheet\Reader\Exception: test is an Invalid Spreadsheet file. in /project/vendor/phpoffice/ phpspreadsheet/src/PhpSpreadsheet/Reader/Csv.php:288
```

index.php
```php
<?php

$spreadsheet = new \PhpOffice\PhpSpreadsheet\Reader\Csv();
$spreadsheet->load('test');

// test contains: 
// aaa,bbb,"<a href="
```

- [x] Reader

csv

PhpSpreadsheet  1.29.0
php 8.1.2

oleibman:
Fix #4036. The issue was originally reported as #564 (and #811) and fixed for the most part, but this is a variation that was not covered by the original. Cells with html fragments can cause `mime_content_type` to identify the file as `text\html`. Original fix was to ignore mime_content_type when file extension is 'csv' or 'tsv'. However, if the file does not have one of those extensions, it will be rejected by Csv Reader as invalid mimetype. This PR adds text\html to the list of valid mimetypes.

I imagine that this type of problem might occur for other mimetypes. If any of those are reported in future, it might be better to just add a "suppress mimetype" check option, rather than extending the list forever. Html is unusual in that its rules are so lax, which is why it seems appropriate to add it here.

Note that IOFactory may still identify a file as Html even when intended as Csv. The sample associated with this issue does not fall into this category, but one of the unit tests on this ticket does. The file will still be read correctly by Csv Reader, but IOFactory load may cause it to use Html Reader instead.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 - POWER Null/Bool Args. [PR #4031](https://github.com/PHPOffice/PhpSpreadsheet/pull/4031)
 - Do Not Output Alignment and Protection for Conditional Format. [Issue #4025](https://github.com/PHPOffice/PhpSpreadsheet/issues/4025) [PR #4027](https://github.com/PHPOffice/PhpSpreadsheet/pull/4027)
 - Xls Conditional Format Improvements. [PR #4030](https://github.com/PHPOffice/PhpSpreadsheet/pull/4030)
+- Csv Reader allow use of html mimetype. [Issue #4036](https://github.com/PHPOffice/PhpSpreadsheet/issues/4036) [PR #4049](https://github.com/PHPOffice/PhpSpreadsheet/pull/4040)
 
 ## 2024-05-11 - 2.1.0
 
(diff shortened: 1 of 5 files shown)


oleibman:
Pretty funny result with Php nightly. `mime_content_type` is changed to recognize the files as `text/csv` after all, which means that this problem would have gone away on its own eventually. Of course, that'll be several years away, so we need to proceed with this change, but I have to tweak the tests.

Issue #3206, opened 2022-11-24
Conditional formats applying to multiple cells are duplicated when reading and writing an excel file


ndench:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

Conditional formats should not be changed when reading then writing an excel file.

### What is the current behavior?

Reading an excel file that contains a conditional formatting rule which applies to multiple cells (ie. one rule: `$A$1,$C$1,$E1`) and then immediately writing the file without making any changes will result in the rule being duplicated so that it only applies to single cells (ie. 1 rule: `$A$1`, another rule `$C$1` and another `$E$1`).

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Read the excel file that has a single conditional format applying to multiple cells
$reader = new \PhpOffice\PhpSpreadsheet\Reader\Xlsx();
$excel = $reader->load('multili-format.xlsx');

// Immediately write the excel file back out without making any changes
$writer   = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($excel);
$writer->save('multi-format-output.xlsx');
```

Here is the original file that has no content, and only a a single conditional formatting rule for `A1 contains a blank value` with the format `Not format set`
[multi-format.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/10079888/multi-format.xlsx)
![multi-format](https://user-images.githubusercontent.com/2062388/203666918-6a482d67-1457-4554-91bc-b27446f8669e.png)

Here is the output file after running the above script, you can see the conditional formatting rule now has a format of `light red fill, dark red text`.
[multi-format-output.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/10079889/multi-format-output.xlsx)
![multi-format-output](https://user-images.githubusercontent.com/2062388/203666938-082913b5-d6f6-42b2-ae3b-d12344191325.png)

### What features do you think are causing the issue

- [ ] Reader
- [x] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

### Does an issue affect all spreadsheet file formats? If not, which formats are affected?

- XLSX

[...]

PR #4042, opened 2024-05-24
Conditional Range Unions and Intersections

### LINKED ISSUE #4039: Conditional styles comming separates, even though two of them are one conditional style when you open the sheet
when I tried to import a sheet using the function:
$this->worksheet->getConditionalStylesCollection();
the data inside of it was splitter by ranges
but in reality there are two ranges belong to one set of conditional styling
can you please provide a general function that fetches the conditional stylings as its defined?
[Conditional styles test.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/15401754/Conditional.styles.test.xlsx)
this is a test file to check with

oleibman:
Fix #4039. Excel sometimes stores the location for a Conditional as, say, `B1:B10 C1:C10` rather than `B1:C10`. PhpSpreadsheet does not have a problem with this, but internally stores it as separate Conditionals, one for each range. (There may be more than 2.) User would like it stored as a single Conditional. Since the range is used as the index of an array which holds the conditionals, there is no technical reason why this can't be done. And it does seem like being able to change multiple ranges all at once has some advantages, whether you're doing it in PhpSpreadsheet or in Excel itself.

Making such a change in Xlsx Reader showed no problem in Xlsx Reader and Writer tests. A number of problems did show up in CellMatcherTest, and one in WizardFactoryTest. This was good news, because it meant some of our test spreadsheets used the same type of construction as the test spreadsheet supplied with the issue. So, if additional fixes make the test problems go away, I think no new tests are required. And a smattering of source changes did indeed result in a clean test suite once more.

This is a change, but I don't really think it should break anyone. So I don't think it's necessary to add an option  to opt in to the old or new behavior. I could be wrong. I'll leave this ticket open for at least a couple of weeks to see if anyone thinks otherwise.

This is:

- [x] a bugfix (debatable)
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
There is additional discussion of this topic in issue #3206.

oleibman:
And, following on that discussion, there is more Excel weirdness. Nominally, the intersection operation for cell ranges is a space, and the union operator is a comma. Now, let's take a look at a worksheet.
[cond.union.intersect.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/15462587/cond.union.intersect.xlsx)
When we open the spreadsheet, and look at the range for the conditional style, what do we see?
```
=$A$1:$A$3,$C$1:$E$3
```
The separator is comma, and this is treated as a union, as intended. But ... what do we see in the Xml?
```xml
<conditionalFormatting sqref="A1:A3 C1:E3">
```
Here the separator is space, not comma. Excel is treating it as a union despite the apparent use of the intersection operator! And, not surprisingly, further testing seems to indicate that the use of comma as separator in the xml is treated by Excel as intersection, not union. I need to take some time to figure out how to differentiate ranges loaded from a spreadsheet vs. ranges entered via `Worksheet::setConditionalStyles`. Not to mention differentiating between the two in our own code.

So, expect some delay in merging this PR.

oleibman:
Xls Reader expects only simple ranges, not union or intersection. Therefore, Xls Writer needs to split union ranges into separate ranges before writing.

oleibman:
Updated branch to pull in some recent Conditional changes, not because PR is close to ready.

oleibman:
Taking out of draft status. I think the code has gone as far as it needs. I need to think about whether additional tests are needed. AFAICT, intersection and union are now completely supported for Xlsx/Xls Conditional; but intersection may not be fully supported in other contexts.

PR #4074, opened 2024-06-26
Update JpGraphRendererBase.php - check existing of PlotLabel

yfinkel:
I've got
Uncaught PHP Exception Symfony\Component\ErrorHandler\Error\FatalError: "Error: Uncaught Error: Call to a member function getDataValue() on bool in .../vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Chart/Renderer/JpGraphRendererBase.php:337

I found that it's necessary to check existing of PlotLabel before using its method getDataValue().

This is:

- [x ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [ x] Code style is respected
- [x ] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)

### DIFF
--- a/src/PhpSpreadsheet/Chart/Renderer/JpGraphRendererBase.php
+++ b/src/PhpSpreadsheet/Chart/Renderer/JpGraphRendererBase.php
@@ -334,7 +334,8 @@ abstract class JpGraphRendererBase implements IRenderer
                 //    Set the appropriate plot marker
                 $this->formatPointMarker($seriesPlot, $marker);
             }
-            $dataLabel = $this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index)->getDataValue();
+            if ($this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index))
+                $dataLabel = $this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index)->getDataValue();
             $seriesPlot->SetLegend($dataLabel);
 
             $seriesPlots[] = $seriesPlot;
--- a/src/PhpSpreadsheet/Chart/Renderer/JpGraphRendererBase.php
+++ b/src/PhpSpreadsheet/Chart/Renderer/JpGraphRendererBase.php
@@ -334,8 +334,9 @@ abstract class JpGraphRendererBase implements IRenderer
                 //    Set the appropriate plot marker
                 $this->formatPointMarker($seriesPlot, $marker);
             }
-            if ($this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index))
+            if ($this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index)) {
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Chart/Renderer/JpGraphRendererBase.php
```diff
@@ -334,7 +334,9 @@ private function renderPlotLine(int $groupID, bool $filled = false, bool $combin
                 //    Set the appropriate plot marker
                 $this->formatPointMarker($seriesPlot, $marker);
             }
-            $dataLabel = $this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index)->getDataValue();
+            if ($this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index)) {
+                $dataLabel = $this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index)->getDataValue();
+            }
```
oleibman, line 339: I think `seriesPlot->SetLegend($dataLabel)` needs to be inside the if statement, otherwise dataLabel will not be defined.

oleibman: Also, please add a unit test to demonstrate that there was a problem before your fix and you've fixed it.

yfinkel:
I can't, because I use a big docx file in my project, which causes a crash
in this place. I can't localize the issue in this file, because it's very
big with many formulae and charts. But this change solved the problem for
me.

--
Yury Finkel


ср, 26 июн. 2024 г. в 11:39, oleibman ***@***.***>:

oleibman:
Without an example, I am not completely unwilling to proceed, but I am a little reluctant. However, reworking your change will make me a little less reluctant.

There are 4 places in JpGraphRendererBase where dataLabel is set.
- renderPlotLine, around line 337.
- renderPlotBar, around line 411. This already tests for the situation you are trying to defend against, and I know that that situation is exercised in our unit test code.
- renderPlotScatter, around line 495.
- renderPlotRadar, around line 527.

If you were to make the code from renderPlotBar into a subroutine, and change each of these 4 instances to call the new subroutine instead of assigning a value to dataLabel directly, we would have some assurance that the other 3 are using code which is already tested. If you can confirm that the works in your case, I could proceed. As a bonus, we won't have to deal with the same situation for Radar and Scatter when someone comes across a similar situation with those in future. Something like (untested):
```php
private function assignDataLabel(DataSeries $groupId, int $index): mixed
{
    if (!$this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index)) {
        $dataLabel = '';
    } else {
        $dataLabel = $this->chart->getPlotArea()->getPlotGroupByIndex($groupID)->getPlotLabelByIndex($index)->getDataValue();
    }

    return $dataLabel;
}
```

yfinkel:
Well, I updated the code. The function name is getDataLabel(). It works well in my project. I will reappear only after a month, so do what you want with it. It's enough for me that it works for me.

oleibman:

PR #4119, opened 2024-07-29
Security: prevent XXE (XML External Entity) when loading files

PowerKiKi:
Prevent XEE by hiding custom entities by using single quote to declare a non-UTF-8 encoding.

XML standard, https://www.w3.org/TR/xml/#NT-EncodingDecl, allows single quote to declare encoding, but we did not support it. Instead, we incorrectly fell back on the default of UTF-8. That incorrectly kept the XML as non-UTF-8, and thus prevented our regexp-based custom entity detection mechanism to work.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ## TBD - 3.0.0
 
+### Security Fix
+
+- Prevent XXE when loading files
+
 ### Added
 
 - Nothing
--- a/src/PhpSpreadsheet/Reader/Security/XmlScanner.php
+++ b/src/PhpSpreadsheet/Reader/Security/XmlScanner.php
@@ -35,21 +35,33 @@ class XmlScanner
 
     private function toUtf8(string $xml): string
     {
-        $pattern = '/encoding="(.*?)"/';
-        $result = preg_match($pattern, $xml, $matches);
-        $charset = strtoupper($result ? $matches[1] : 'UTF-8');
-
+        $charset = $this->findCharSet($xml);
         if ($charset !== 'UTF-8') {
             $xml = self::forceString(mb_convert_encoding($xml, 'UTF-8', $charset));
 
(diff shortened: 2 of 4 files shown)


ionmacariatremend:
Can we add this fix for ^1 version too? For now any deploy will crash for components which are using ^1 version...

AoiKagase:
Will there ever be a backport to 2.1.0?
The PHP version that fully supports backporting on RHEL9 is 8.0, and the last version of phpspreadsheet that can be installed with composer is 2.1.0.

oleibman:
@AoiKagase Php8.0 reached end-of-life (no updates, not even for security) almost a year ago. There seems to be a way to install a supported version on RHEL9 [here](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/installing_and_using_dynamic_programming_languages/assembly_using-the-php-scripting-language_installing-and-using-dynamic-programming-languages#installing-the-php-scripting-language_assembly_using-the-php-scripting-language). If that is not possible, might PhpSpreadsheet 1.29.1 meet your needs?

AoiKagase:
Although PHP itself is no longer supported, RHEL9 has a 10-year guarantee that security updates to 8.0 will be backported by Redhat.
https://access.redhat.com/support/policy/updates/rhel-app-streams-life-cycle#rhel9_full_life_application_streams
> Application Stream | Release Date | Retirement Date | Release
> PHP 8.0 | May 2022 | May 2032 | 9.0
Since composer installs 2.1.0 by default, many companies may end up using the vulnerable version.

ChinhTQ-Vnext:
Can add fix to apply version php 7.3..https://github.com/PHPOffice/PhpSpreadsheet/pull/4119 not support for php 7.3

oleibman:
@ChinhTQ-Vnext We are unlikely to act on your request. See discussion in issue #4154, and continue the discussion there if you wish.

oleibman:
@AoiKagase I have released 2.1.1 which will handle your situation. Thank you for your explanation.

AoiKagase:
@oleibman Thank you. I will update with it.

PR #4127, opened 2024-08-06
Currency and Accounting Formats

### LINKED ISSUE #4124: Certain currency formats are handled incorrectly by formatter
This is:

- [x] a bug report
- [ ] a feature request

``1 HUF``

``#,##0 [$0000000f]``, which in turn causes auto column width to be unnecessarily wide.

```php
<?php

// Currency formats that have ISO codes and no digits will have issue.
echo \PhpOffice\PhpSpreadsheet\Style\NumberFormat\Formatter::toFormattedString(1, '#,##0_-[$HUF]');
```

- [ ] Reader
- [ ] Writer
- [x] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

#### The source of the issue is the following condition, designating the format to be a date format:

https://github.com/PHPOffice/PhpSpreadsheet/blob/1c77e0049902b245c92c233aadcc834f50d246af/src/PhpSpreadsheet/Style/NumberFormat/Formatter.php#L162

.xlsx, most likely others too.

PHP: 8.2
PhpSpreadsheet: at least up from 2.1.0

### LINKED ISSUE #4125: Currency and Accounting format wizards unable to generate formats with ISO code currencies.
This is:

- [x] a bug report
- [ ] a feature request

Apply a valid currency format.

Excel does not accept the format code, displays issue on open.

```php
<?php

$sheet->getCell([1, 1])->setValue(1234)->getStyle()->applyFromArray(
    [
        'numberFormat' => [
            'formatCode' => (new \PhpOffice\PhpSpreadsheet\Style\NumberFormat\Wizard\Accounting('EUR'))->format(),
        ],
    ]
);
```

- [ ] Reader
- [ ] Writer
- [x] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

.xlsx, most likely others too.

PHP: 8.2
PhpSpreadsheet: at least up from 2.1.0
Excel: 2021/365
[output.xlsx](https://github.com/user-attachments/files/16480619/output.xlsx)

oleibman:
Fix #4125. Currency and Accounting Wizards generate styles for ISO codes, but these are incorrect and cause a problem when Excel tries to open a spreadsheet containing these styles. Debugging that problem, other problems with Wizards came to light:
- Currency Wizard should permit four different styles for negative numbers (as Excel does) - minus sign, minus sign and red font, paretheses, and parenthese and red font. It currently uses only minus sign.
- Accounting Wizard should use parentheses for negative numbers (as Excel does). It currently uses minus sign.
- Accounting Wizard should always use SYMBOL_WITH_SPACING (as Excel does). It currently permits the use of SYMBOL_WITHOUT_SPACING. What WITH_SPACING really does is to ensure decimal-point alignment among adjacent cells in a column with the same format.
- Currency Wizard should always use SYMBOL_WITHOUT_SPACING (as Excel does). It currently permits the use of SYMBOL_WITH_SPACING.

I am correcting these problems by:
- renaming Currency Wizard to CurrencyBase
- adding a `negative` property with setter to it and its constructor.
- adding a new Currency which extends CurrencyBase, always using SYMBOL_WITHOUT_SPACING when formatting.
- having Accounting extend CurrencyBase rather than Currency, always using SYMBOL_WITH_SPACING and NEGATIVE_PARENS when formatting.
- CurrencyBase can be used if the restrictions on Currency and Accounting are not desired (e.g. the suggested accounting constant from [this unimplemented PR](https://github.com/PHPOffice/PhpSpreadsheet/pull/1576)).

Excel does some funny stuff with these formats. In particular, it might try to guess if you have a particular Accounting format in mind. So the Accounting wizard for dollar sign generates a format which (a) matches FORMAT_ACCOUNTING_USD, and (b) Excel (correctly) interprets as an Accounting format for symbol $. On the other hand, the Accounting wizard for euro sign generates a format which (a) matches FORMAT_ACCOUNTING_EUR, but (b) Excel interprets as a custom code rather than an Accounting format. This in itself is not a particularly big deal, but it has made it impossible for me to see exactly what format Excel uses for trailing currency symbols for negative numbers. I can't get them to decimal-point align with positive numbers if I put any kind of space between the trailing parenthesis and the currency symbol, so I omit that. It doesn't look terrible, and it keeps everything aligned, but it might not be what people are used to.

I've also changed the formatting to use spaces rather than non-breaking spaces. They seem to work just fine, and the constants mentioned above use them rather than nbsp.

These changes broke some Wizard Accounting tests. However, I think those tests were checking the wrong thing. They checked the format code, rather than the results of formatting data using that format code. So they are slightly altered to make those checks instead.

Fix #4124. Currency formats that contain an ISO currency code which contains one of the characters used to recognize a date format (hmsdy), e.g. [$HUF], are being formatted by PhpSpreadsheet as dates rather than currencies. Code is changed to recognize open bracket followed by dollar sign followed by 3 Latin alphabetic characters followed by close bracket as a non-date.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Not concerned about Scrutinizer "complexity" message.

PR #4142, opened 2024-08-16
Image Transparency/Opacity Including Html Reader Changes

oleibman:
See discussion #4117. This PR implements image transparency, for Xlsx Reader and Writer, Html Reader, and Html/Dompdf/Mpdf Writer. (Mpdf treats 100% opacity as if it were zero, but it's otherwise okay and that would be an unusual choice anyhow.) A new property `opacity` with getter/setter is added to BaseDrawing (and therefore Drawing and MemoryDrawing). Although the Excel UI lets you set the image *transparency*, the value stored in the Xml is actually its *opacity* (expressed as an integer between 0 and 100,000). Likewise Html/Css lets you set opacity (as a float between 0 and 1). PhpSpreadsheet lets you set it as it is stored in Xml - so a value of 40,000 would indicate 40% opacity = 60% transparency.

In the course of testing, some problems with Html Reader presented themselves. They are corrected as part of this ticket:
- Web page title will no longer be "Untitled Spreadsheet" if a title tag is present in the header section of the html.
- If a class attribute is used with the table tag representing a worksheet, showGridlines and printGridlines will be set according to whether gridlines and/or gridlinesp are among the class names.
- Reader has been performing urldecode on the src attribute of the img tag. This breaks data url's by converting plus signs to spaces. It will now perform urldecode only for non-data url's.
- If height and width are not used as attributes on the img tag, they will be extracted from the style attribute if they are present there.

This is:

- [x] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Too much has happened since this PR was started. I will probably replace it with a new version.

oleibman:
Not sure why "docs" step failed, but, since this PR doesn't update any docs, I don't think there's any cause for concern.

PR #4163, opened 2024-09-06
:wastebasket: Deprecate `IReader::SKIP_EMPTY_CELLS`

homersimpsons:
This is an alias to `IGNORE_EMPTY_CELLS` there is no point to have both.

This is:

- [-] a bugfix
- [-] a new feature
- [x] refactoring
- [-] additional unit tests

Checklist:

  - [-] New unit tests have been added
- [x] Code style is respected

`IReader::SKIP_EMPTY_CELLS` is an alias to `IGNORE_EMPTY_CELLS` there is no point to have both.

I have some questions:
1. Should I prepare the merge request for 3.0?
    1. If yes, Which branch should I target?
2. Should I deprecate + remove, or can I remove it directly in 3.0?

(Maybe this change is not wanted, but I feel like having both options is confusing)

### DIFF
--- a/docs/topics/reading-and-writing-to-file.md
+++ b/docs/topics/reading-and-writing-to-file.md
@@ -1113,19 +1113,19 @@ Flags that are available that can be passed to the Reader in this way include:
 
  - $reader::LOAD_WITH_CHARTS
  - $reader::READ_DATA_ONLY
- - $reader::IGNORE_EMPTY_CELLS 
- - $reader::SKIP_EMPTY_CELLS (synonym for IGNORE_EMPTY_CELLS)
+ - $reader::IGNORE_EMPTY_CELLS
+ - $reader::IGNORE_ROWS_WITH_NO_CELLS
 
-| Readers  | LOAD_WITH_CHARTS | READ_DATA_ONLY | IGNORE_EMPTY_CELLS |
-|----------|------------------|----------------|--------------------|
-| Xlsx     | YES              | YES            | YES                |
-| Xls      | NO               | YES            | YES                |
-| Xml      | NO               | NO             | NO                 |
-| Ods      | NO               | YES            | NO                 |
-| Gnumeric | NO               | YES            | NO                 |
-| Html     | N/A              | N/A            | N/A                |
-| Slk      | N/A              | NO             | NO                 |
-| Csv      | N/A              | NO             | NO                 |
+| Readers  | LOAD_WITH_CHARTS | READ_DATA_ONLY | IGNORE_EMPTY_CELLS | IGNORE_ROWS_WITH_NO_CELLS |
+|----------|------------------|----------------|--------------------|---------------------------|
+| Xlsx     | YES              | YES            | YES                | YES                       |
+| Xls      | NO               | YES            | YES                | NO                        |
+| Xml      | NO               | NO             | NO                 | NO                        |
+| Ods      | NO               | YES            | NO                 | NO                        |
+| Gnumeric | NO               | YES            | NO                 | NO                        |
+| Html     | N/A              | N/A            | N/A                | N/A                       |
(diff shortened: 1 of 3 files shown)


oleibman:
There is a point to have both. This would be a breaking change for people who use SKIP_EMPTY_CELLS. I would need to hear why you think the improvement offered by this change outweighs the problems that it would cause.

homersimpsons:
> There is a point to have both. This would be a breaking change for people who use SKIP_EMPTY_CELLS.

I suggest to remove it in 3.0. It should be rather straightforward to search and replace this through a codebase.

> I would need to hear why you think the improvement offered by this change outweighs the problems that it would cause.

1. The documentation sometimes refers `IGNORE_EMPTY_CELLS` and sometimes `SKIP_EMPTY_CELLS`, a user may not be aware they are the same.
2. In a codebase seeing both may lead to confusion while they are actually the same.
3. Seeing the same value defined for both made me think there was a mistake at first.
4. In this library code we were comparing twice with the same value, this is an example of the kind of confusion this creates

Of course there are no strong disadvantages, and the breaking change may not be worth. But I think having only one way to do things is less error prone.

oleibman:
I would have no problem with making it clear in the documentation that these are aliases of each other. I will note, however, that the online documentation already does this:
```
$reader::SKIP_EMPTY_CELLS (synonym for IGNORE_EMPTY_CELLS)
```
But there is no doc-block where the constants are defined which emphasizes this point; I would have no trouble if you wanted to add that. I wouldn't even have a problem if you wanted to mark one of them deprecated in its doc-block so that it can be removed in a *future* release. Likewise, if you wish to make the examples consistently use one and not the other, I would have no problem with that. And I certainly have no problem with your changing BaseReader to perform only one test (assuming it still passes the unit test suite, which I'm confident it would). But you can do all of that without causing a break, which I think is just too high a cost for too little benefit, especially considering that there is no notice beforehand.

So, please restore the deleted item in your PR, make the changes mentioned in the previous paragraph, and I will be glad to consider it then.

oleibman:
I cannot entertain deleting the redundant variable as part of imminent 3.0.0 with no notice. (Your changes did not delete it, but your explanatory text indicated that you wanted to.) However, I am okay with declaring it deprecated, and your other changes. I will, unfortunately, not be able to merge this change till after 3.0.0; my plate is overflowing till then. Please change your deprecation line to:
```php
@deprecated 3.1.0 use IGNORE_EMPTY_CELLS instead
```
I will update it if it can be installed earlier, or needs to wait longer. Also, please add a doc-block for IGNORE_EMPTY_CELLS.

I believe your php-cs-fixer problem is that the doc-block summaries need to end with a period (not my rule, I'm just here to explain). So "Flag used to load any charts that are defined (if the Reader supports Charts).", etc., should be accepted. Note that this rule applies only to summary text, and not, for example, to the `@deprecated` tag.

homersimpsons:
> I cannot entertain deleting the redundant variable as part of imminent 3.0.0 with no notice. (Your changes did not delete it, but your explanatory text indicated that you wanted to.) However, I am okay with declaring it deprecated, and your other changes. I will, unfortunately, not be able to merge this change till after 3.0.0; my plate is overflowing till then. 

Done

> Also, please add a doc-block in IReader for IGNORE_EMPTY_CELLS. (And, optionally, for IGNORE_ROWS_WITH_NO_CELLS.)

Done, I'm not entirely sure about the documentations wordings. Also, there are 3 sources of documentations: the md doc, the flag doc, the getter doc. Maybe the flag doc could refer the getters doc (or vice versa) ?

> I believe your php-cs-fixer problem is that the doc-block summaries need to end with a period (not my rule, I'm just here to explain). So "Flag used to load any charts that are defined (if the Reader supports Charts).", etc., should be accepted.

Thank you!

oleibman:

PR #4221, opened 2024-11-09
Fix incorrect versions error (ForbiddenThisUseContexts)

### LINKED ISSUE #4220: Composer versions reports error "$this can no longer be used in a plain function or method since PHP 7.1"
This is:

```
- [x] a bug report
- [ ] a feature request
```

`composer versions` should not return an error.

`composer versions` finds the following errors:

```
FILE: PhpSpreadsheet/Style/NumberFormat/Wizard/CurrencyNegative.php
-------------------------------------------------------------------------------------------------------
FOUND 3 ERRORS AFFECTING 3 LINES
-------------------------------------------------------------------------------------------------------
 14 | ERROR | "$this" can no longer be used in a plain function or method since PHP 7.1.
 22 | ERROR | "$this" can no longer be used in a plain function or method since PHP 7.1.
 30 | ERROR | "$this" can no longer be used in a plain function or method since PHP 7.1.
-------------------------------------------------------------------------------------------------------
```

run `composer versions`

- [ ] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

Bug is not related to a spreadsheet format.

Bug is introduced in commit #5e248cf. So, bug is introduced in latest release v3.3.0.

m7913d:
PHPCompatibility erroneously flags the use of $this in enumerations. Commit 5e248cf moved disabling of this error check to .github/workflows/main.yml. Hence, `composer versions` and `composer check` should also exclude this check.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Fixes #4220

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -47,7 +47,7 @@
         "check": [
             "./bin/check-phpdoc-types",
             "phpcs samples/ src/ tests/ --report=checkstyle",
-            "phpcs samples/ src/ tests/ --standard=PHPCompatibility --runtime-set testVersion 8.0- -n",
+            "phpcs samples/ src/ tests/ --standard=PHPCompatibility --runtime-set testVersion 8.0- --exclude=PHPCompatibility.Variables.ForbiddenThisUseContexts -n",
             "php-cs-fixer fix --ansi --dry-run --diff",
             "phpunit --color=always",
             "phpstan analyse --ansi --memory-limit=2048M"
@@ -61,7 +61,7 @@
             "php-cs-fixer fix"
         ],
         "versions": [
-            "phpcs samples/ src/ tests/ --standard=PHPCompatibility --runtime-set testVersion 8.0- -n"
+            "phpcs samples/ src/ tests/ --standard=PHPCompatibility --runtime-set testVersion 8.0- --exclude=PHPCompatibility.Variables.ForbiddenThisUseContexts -n"
         ]
     },
     "require": {
--- a/composer.json
+++ b/composer.json
@@ -45,12 +45,12 @@
     ],
     "scripts": {
         "check": [
-            "./bin/check-phpdoc-types",
+            "php ./bin/check-phpdoc-types",
             "phpcs samples/ src/ tests/ --report=checkstyle",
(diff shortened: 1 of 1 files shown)


oleibman:
Not an area I've ever really looked at. As long as you are changing that, can you also change:
`./bin/check-phpdoc-types`
to
`php ./bin/check-phpdoc-types`
That will allow it to run on Windows. I assume you were testing on some version of Unix/Linux, so, of course, please verify that this doesn't break things on your system.

And we should probably run phpstan before phpunit (assuming that a failure in any stops the script).

m7913d:
> Not an area I've ever really looked at. As long as you are changing that, can you also change: `./bin/check-phpdoc-types` to `php ./bin/check-phpdoc-types` That will allow it to run on Windows. I assume you were testing on some version of Unix/Linux, so, of course, please verify that this doesn't break things on your system.
> 
> And we should probably run phpstan before phpunit (assuming that a failure in any stops the script).

@oleibman Indeed tested on Ubuntu. Should I change that too in this PR? Isn't that a violation of "Please don't "bundle" several changes into a single PR; submit a PR for each discrete change/fix." of the [contributing guidelines](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/CONTRIBUTING.md)?

oleibman:
I will think about whether the contributing guidelines need to change. In the meantime, I think what happened is that you saw that the version script didn't work properly, and then noticed that the check script also didn't work properly. You sensibly did not make separate PRs for each of these problems. So, I think the problem you're solving is "composer scripts don't work". Fixing the two additional items fall under the same category, so, although you're certainly welcome to argue the point, I think that they all make a discrete change. It is not like a case where, say, you find an error in Html Reader and an error in calculation for the CONCATENATE function, and try to bundle fixes for both into a single PR; that is what I think was the intent of the guideline.

m7913d:
@oleibman Ok, I will include the requested changes in this PR. However, I will only test in on Ubuntu.

oleibman:
Thank you for your contribution, and for being flexible.

PR #4240, opened 2024-11-24
Breaking Change for DataValidation

### LINKED ISSUE #797: Memory exhausted  while load blank excel which the first column setDataValidation with a column range
I has a blank excel file which the first column setDataValidation with a column range.
then when use phpoffice/phpspreadsheet 1.5.2 to load the file, code as follows.
```
$phpReader = IOFactory::createReader('Xlsx');
$excel = $phpReader->load('D:\\develop\\test.xlsx');
```

```
Apache record error log as follows 
Fatal error: Allowed memory size of 268435456 bytes exhausted (tried to allocate 64 bytes) in phpoffice\phpspreadsheet\src\PhpSpreadsheet\Cell\Coordinate.php on line 336
```

### LINKED ISSUE #4091: Data Validation
Currently in the functionality there is only way to set validation per cell
public function setDataValidation(string $cellCoordinate, ?DataValidation $dataValidation = null): static
and the get also return the range as parts:
$this->worksheet->getDataValidationCollection();

can we let it return as defined in the excel file:
and also set as range instead of cell by cell
Screenshot 2024-07-08 at 16 12 21

### LINKED ISSUE #4206: DataValidator Does Not Handle All Types of List
This is:

```
- [x] a bug report
- [ ] a feature request
```

DataValidator::isValid calls DataValidator::isValueInList when appropriate. The expected behavior is that true/false will be returned depending on the cell's value vis-&agrave;-vis the list..

The method expects the list to be either a comma-delimited string or a cell range. With the merge of PR #4203, the list can also be the result of a spill operator, which is implemented as a call to the ANCHORARRAY function. The validator function will not recognize this, and will therefore always return "true" in this situation. It also appears that a named range might be usable as a validator list, and that also won't be recognized.

- [ ] Reader
- [ ] Writer
- [ ] Styles
- [x] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

The spreadsheet will work properly. It is only validation within PhpSpreadsheet itself that might not yield the correct answer.

All.

oleibman:
Fix #797 (marked stale in 2018, but now reopened). Fix #4091. Fix #4206. DataValidation is specified in the Xml at the sheet level rather than the cell level. PhpSpreadsheet, however, currently requires a cell and a data validation object for each cell in a range. As a consequence, it may exhaust memory allocating objects that are never really needed. If it reads a spreadsheet which applies DataValidation to an entire column, the reader creates a million cells and a million DataValidation objects. An additional problem is that, when it saves the spreadsheet, it creates a row entry for each of the million rows even if they contain no data.

The code is changed to locate DataValidation for a cell by running the DataValidation list to see if the Sqref for any entry matches the cell in question. This eliminates the need to require a DataValidation entry for each cell; indeed, it eliminates the need to create a cell just because it is subject to DataValidation. Initial tests are very encouraging. The memory-exhausting spreadsheet from 797 required over 1GB of memory. With this change in place, that is reduced to 6MB, and DataValidation works as expected for all cells in the range.

This change is a work in progress. It needs formal tests. I need to see how it works with inserting or removing rows and columns. I need to see how Excel handles conflicting entries. My early tests indicate that it at least doesn't do anything irrational (like claiming corruption) when dealing with conflicts. It appears that, if I have a DataValidation applied to column A, and another applied to cell A5, whichever appears first in the Xml is the ultimate arbiter of what rule applies to A5.

If this change is merged, it will be a breaking range. It will require a new major release (probably PhpSpreadsheet 4.0), and will not happen for at least a couple of months, and probably later.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Existing code for ReferenceHelper adjusts sqref for DataValidation when rows/columns inserted/deleted, but does not adjust formula (where list may be expressed as a range). That is a bug, and I think it needs to be fixed as part of this PR.

oleibman:
I will probably continue to make tweaks, but this PR is now in pretty good shape. I have taken it out of draft status, and the title no longer indicates work in progress. My plan is to merge it somewhere around Feb. 4 and create breaking release 4.0 about a week after that.

oleibman:

PR #4262, opened 2024-12-04
Remove unnecessary files from composer package

jlherren:
These files make no sense to be distributed as part of the composer package.

If desired I can make the same change also for 1.x, where there are two additional files to be removed.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -3,7 +3,9 @@
 /.gitattributes export-ignore
 /.github export-ignore
 /.gitignore export-ignore
-/.php_cs.dist export-ignore
+/.php-cs-fixer.dist.php export-ignore
+/.phpcs.xml.dist export-ignore
+/.readthedocs.yaml export-ignore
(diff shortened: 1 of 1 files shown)


oleibman:
Seems okay, but I need to read up on this. Should `.readthedocs.yaml` also be added to your list?

I don't know about earlier releases. I will give it some thought. What are the additional files you would like to exclude for release 1?

jlherren:
Version 1.29.5 also ships with:
- `phpstan-conditional.php`
- `phpunit10.xml.dist`

However, 3.5.0 does not ship with `.readthedocs.yaml` and I don't see why not, so perhaps there's more going on that just what's in `.gitattributes`.

jlherren:
My bad, I didn't realize that the screenshot was from 1.29.5, which in fact did not have `.readthedocs.yml` yet. I do get it on 3.5.0.

I updated the PR.

oleibman:

Issue #4282, opened 2024-12-17
"3D Range references are not yet supported" error when trying to use GROUPBY, is there a way to support it?


cuartas15:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?

That a =GROUPBY formula works

### What is the current behavior?

It doesn't work, throwing the error "3D Range references are not yet supported"

### What are the steps to reproduce?

Have a table with dates and values, in a different sheet try to group them by dates and values so it sums the total on each date, like this:
`$workSheet2->setCellValue('A1', '=GROUPBY(Report!B2:B' . (count($data) + 1) . ';Report!D2:D' . (count($data) + 1) . ';SUM)');`

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
$spreadsheet = new Spreadsheet();
$workSheet = $spreadsheet->getSheet(0);

$data = array(
    ['anything', '2024-12-10', 'anything', '5000'],
    ['anything', '2024-12-10', 'anything', '2500'],
    ['anything', '2024-12-09', 'anything', '12000'],
    ['anything', '2024-12-09', 'anything', '100'],
);

$workSheet->fromArray($data, '', 'A2');

$workSheet2 = new \PhpOffice\PhpSpreadsheet\Worksheet\Worksheet($spreadsheet, 'Dynamic');
$spreadsheet->addSheet($workSheet2);
$workSheet2->setTitle('Dynamic table');

$workSheet2->setCellValue('A1', '=GROUPBY(Report!B2:B' . (count($data) + 1) . ';Report!D2:D' . (count($data) + 1) . ';SUM)');
<?php

### What features do you think are causing the issue

- [ ] Reader
- [x] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

### Does an issue affect all spreadsheet file formats? If not, which formats are affected?

At least xlsx

### Which versions of PhpSpreadsheet and PHP are affected?

PhpSpreadsheet 1.29.6
PHP 7.4

oleibman:
[...]

PR #4283, opened 2024-12-19
Extremely Limited Support for GROUPBY Function

oleibman:
This is a partial response to issue #4282. The actual logic to implement GROUPBY is probably very complicated. And, even worse, Excel has thrown a whole new way of (internally) specifying one of the arguments into the mix. That argument is a function name, expressed not as a mapped integer (as SUBTOTAL does), nor even as a string, but as the unquoted function name prefixed by `_xleta.`. And, unlike its `_xlfn.` and `_xlws.` predecessors, it is difficult to figure out when the new prefix needs to be added, and when it needs to be ignored. I am not even going to attempt that task with this ticket.

So, what does this change do? Like earlier attempts to introduce limited functionality (such as with form controls), it is there so that using GROUPBY can be passed through - you can load a spreadsheet that contains it, and save it to a new spreadsheet, and the function and its results are preserved. Some cautionary notes. Dynamic arrays must be enabled (the function makes no sense without doing that). Changing any of the inputs used in the function may result in internal inconsistencies between PhpSpreadsheet and Excel; this is especially so if the dimensions of the returned array change as a result of changes to the input data. The programmer can avoid some of these problems by changing the formulatAttributes of the cell where the function is used; this may be difficult to do in practice. Oh, yes, using the GROUPBY cell as an argument in another formula will probably lead to problems. Finally, I confess that part of this solution looks awfully kludgey to me.

With its limitations and those cautions, is it worth proceeding with this change? My gut feel is that it is more useful to proceed than not. However, I will give others the opportunity to weigh in. I will wait at least a couple of weeks into the new year before proceeding with this.

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Actually, adding a GROUPBY function to a worksheet in PhpSpreadsheet isn't quite as worthless as I had thought. It takes a bit of work - you have to manually add the formula attributes (t must be array, but ref need not be accurate), and manually add the xleta prefix. So, for example:
```php
        $sheet->getCell('E3')->setValue('=GROUPBY(B2:B32, C2:C32, _xleta.SUM)');
        $sheet->getCell('E3')->setFormulaAttributes(['t' => 'array', 'ref=E3']);
```
PhpSpreadsheet will not be able to evaluate the cell, but, if you save the spreadsheet where you've done it, Excel will evaluate it correctly on open.

PR #4339, opened 2025-02-03
changed $trenType to $trendMethod in case TREND_BEST_FIT

quonly:
Fix TREND_BEST_FIT_NO_POLY. This resolves one of the problems in PR #4400. It does not attempt to resolve the other main problem in that issue, the fact that Polynomial Best Fit is mis-coded; it is unclear what needs to be done to fix that.

This change could be consider breaking. I think it is unlikely that any existing code will actually break. Use of Polynomial Best Fit will currently run into errors when it tries to treat a float property as an array, which it does in many places. It will be changed to throw an exception in the constructor until those problems are fixed. One place where BestFit issues `triggerError` will be replaced with an exception. Trend::calculate will now throw an exception if the trend type is unrecognized rather than returning `false`; this is not a particularly legitimate use case, and it allows us to always return `BestFit` rather than `mixed`. None of the conditions for which we will now throw exceptions were part of the unit test suite; they are now.

Some unit tests are added. Draft PR #1907 does not yet add any unit tests, nor does it address the Polynomial problem. It is not clear if it is compatible with this change.


oleibman:
Is there an issue which you are fixing? If so, please open an issue so that we can track it better. Also, please add a formal unit test  which fails with the existing code but succeeds with your change.

oleibman:

PR #4340, opened 2025-02-04
Breaking Change - 3 Defaults

### LINKED ISSUE #4092: $reader->setTestAutodetect() not usefull when php version > 8.1
This is:

```
- [x] a bug report
- [ ] a feature request
```

To not force the user to set autodetect to false when detected php version is >= 8.1
Simply keep it to false or never set it .

Forced to set false in order to not have problems with deprecation notices

Set true with php >= 8.1 version

```php
<?php

```

- [ x] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

any software

oleibman:
Fix #4092. Change default value for Csv Reader autodetect line endings. Prior behavior can be enabled via `setTestAutodetect(true)`.

Change default value for Html Writer "better boolean" logic. Prior behavior can be explicitly enabled via `setBetterBoolean(false)`.

Change default for Xlsx Writer forceFullCalc option. Prior behavior can be explicitly enabled via `setForceFullCalc(null)`.

This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
- [x] more sensible default values

Checklist:

- [x] Code style is respected


u01jmg3:
I realise this is an older PR, but I've just updated from v2 to v4 of PHPSpreadsheet.

---

> _Change default for Xlsx Writer forceFullCalc option. Prior behaviour can be explicitly enabled via setForceFullCalc(null)._

Can you clarify what behaviour change this causes? My interpretation is that `forceFullCalc` has changed from `null` to `false` as its default value, but I'm not sure of the ramifications.

We use `$writer->setPreCalculateFormulas(false);` to allow comma-separated cell references. Will the `forceFullCalc` change have any effect?

Thanks for all your hard work

oleibman:
You probably don't want forceFullCalc, so you probably shouldn't be adversely affected by this change. See PR #4271 for a description of forceFullCalc and why we planned to make this change (as we eventually did).

PR #3898, opened 2024-02-09
Eliminate Some Phpstan-ignore Annotations

oleibman:
This PR resolves about half the statements which we tell Phpstan to ignore. There will still be 23 such annotations spread over 10 source modules for various reasons (too complicated, suspect PhpSpreadsheet code, one Phpstan bug), plus some deliberate errors in the test suite.

This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
- [x] code quality 

Checklist:

- [ ] Changes are covered by unit tests
  - [x] Changes are covered by existing unit tests
  - [ ] New unit tests have been added
- [x] Code style is respected
- [x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ ] CHANGELOG.md contains a short summary of the change and a link to the pull request if applicable
- [ ] Documentation is updated as necessary

### Why this change is needed?

Provide an explanation of why this change is needed, with links to any Issues (if appropriate).
If this is a bugfix or a new feature, and there are no existing Issues, then please also create an issue that will make it easier to track progress with this PR.

PR #4351, opened 2025-02-10
Fix TEXT and TIMEVALUE Functions Master Branch

### LINKED ISSUE #4249: Unsupported operand types: string + string in date format (PHP 8.x)
This is:

```
- [X ] a bug report
- [ ] a feature request
```

detect returned string when calling fromString then skip operation
(line 132)  $value = DateTimeExcel\DateValue::fromString($value) + DateTimeExcel\TimeValue::fromString($value);

string+string is an issue in PHP 8.x
Fatal error: Uncaught TypeError: Unsupported operand types: string + string in C:\Program Files\EasyPHP-Devserver-17\eds-www\aplitec\includes\modules\phpoffice\phpspreadsheet\src\PhpSpreadsheet\Calculation\TextData\Format.php:132

<?php
	$excelfile = 'BUG_PHPSpreadsheet_PHP8_DATE.xlsx';
  	$showPHPExcel = \PhpOffice\PhpSpreadsheet\IOFactory::load($excelfile);
	
	$liassewriter = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($showPHPExcel, "Xlsx");
	$liassewriter->save("crash.xlsx");
?>

- [ ] Reader
- [ X] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

DATE Cell formatting when date cell is still a keyword and not yet a valid date!

### Which versions of PhpSpreadsheet and PHP are affected? PHP >= 8.0
[BUG_PHPSpreadsheet_PHP8_DATE.xlsx](https://github.com/user-attachments/files/17961714/BUG_PHPSpreadsheet_PHP8_DATE.xlsx)

oleibman:
Fix #4249. Technically speaking, only the 1.29 branch needs fixing, and only for TEXT. It was fixed for the other branches by PR #3898. However, in adding test cases for the fix, it became apparent that PhpSpreadsheet's parsing in TIMEVALUE (which is called from TEXT in the original issue) did not really match Excel's. There are probably still edge cases where it doesn't, but, in the absence of a spec for how it operates, this will do for now.

We do not usually backport fixes from the master branch. Because this is more of a forward port from the earlier branch, there is an equivalent PR for each active branch.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Calculation/DateTimeExcel/TimeValue.php
+++ b/src/PhpSpreadsheet/Calculation/DateTimeExcel/TimeValue.php
@@ -2,6 +2,7 @@
 
 namespace PhpOffice\PhpSpreadsheet\Calculation\DateTimeExcel;
 
+use Composer\Pcre\Preg;
 use Datetime;
 use PhpOffice\PhpSpreadsheet\Calculation\ArrayEnabled;
 use PhpOffice\PhpSpreadsheet\Calculation\Functions;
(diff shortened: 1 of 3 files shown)


oleibman:
Scrutinizer new issues are both false positive, and are now suppressed.

Issue #4048, opened 2024-05-28
add custome sheet fonctions


timouchee:
i i try to import a new usable function on the sheet a custome function .
did some function like "addcustomefunction()" existe already in this current project ??

if not can i have a explaination of how to proffesionaly update the code in order to import in the code my custom function ?

thanks

infojunkie:
I don't know about professionally, but I made a one-character modification in the source code to allow adding new custom functions: https://github.com/infojunkie/PhpSpreadsheet/commit/69ed95861ad72c4c89a5c6ee7965d1a1fedd2e7c

By returning a reference to the list of functions, you can add your own custom function in your code, like this:
```php
$functions = &get_class($calculation)::getFunctions();
$functions['CHOOSECOLS'] = [
  'category' => Category::CATEGORY_MATH_AND_TRIG,
  'functionCall' => [Custom::class, 'choosecols'],
  'argumentCount' => '2+',
];

class Custom
{
  /**
   * CHOOSECOLS.
   *
   * Returns the specified columns from an array.
   *
   * @param mixed $cells The cells being searched
   * @param int $cols List of numeric column indexes to extract
   *
   * @return array|string The resulting array, or a string containing an error
   */
  public static function choosecols(mixed $cells, int ...$cols): array|string
  {
    $columns = RowColumnInformation::COLUMNS($cells);
    if (is_string($columns)) {
      return $columns;
    }
    $result = [];
    foreach ($cols as $col) {
      if (!$col || abs($col) > $columns) {
        return ExcelError::VALUE();
      }
      $result[] = array_column($cells, $col > 0 ? $col-1 : $columns-$col);
    }
    return Matrix::transpose($result);
  }
}
```

oleibman:
Closing, this is handled by PR #4390, which is now merged.

Issue #2900, opened 2022-06-17
Add support for custom Calculation functions


looksystems:
This is:

```
- [x] a feature request
```

### What is the expected behavior?

Add ability to add own calculation functions / implementations.

### What is the current behavior?

Currently, Calculation class (and it's usage within Spreadsheet) is private and cannot be overridden or extended without changes to the source code.

In particular, Calculation::$phpSpreadsheetFunctions is private and there are no methods available to add/override function.

**Proposal:** add the following method to the Calculation class...

```php
    public static function registerFunction(string $name, array $definition)
    {
        self::$phpSpreadsheetFunctions[$name] = $definition;
    }
```

ALSO

Calculation::addDefaultArgumentValues only supports static methods (for Calculation functions):

** Proposal:** modify addDefaultArgumentValues...

```php
    private function addDefaultArgumentValues(array $functionCall, array $args, array $emptyArguments): array
    {
        if (is_object($functionCall[0])) {
            $reflector = new ReflectionMethod($functionCall[0], $functionCall[1]);
        } else {
            $reflector = new ReflectionMethod(implode('::', $functionCall));
        }

        $methodArguments = $reflector->getParameters();
```

### What are the steps to reproduce?

N/A

### What features do you think are causing the issue

- [x] Formula Calulations

### Does an issue affect all spreadsheet file formats? If not, which formats are affected?

Yes

### Which versions of PhpSpreadsheet and PHP are affected?

Latest PhpSpreadsheet and all supported versions of php.

looksystems:
PS. Use case: I would like to be able to define a function which accesses external data.

MarkBaker:
> PS. Use case: I would like to be able to define a function which accesses external data.

Useful for you perhaps, but it's a lot more significant as a feature to make it useful for others. Why do you need this to be something that can be defined as a function for the calculation engine? The purpose of teh Calculation Engine is for evaluating spreadsheet formulae, so why do you need to add new functions that don't exist in Excel itself? What functions do you have in spreadsheet formulae that don't exist in MS Excel already?

I can see value with functions between different spreadsheet applications, e.g. ODS supports a `regexp()` function that isn't available in Excel, so being able to extend the base function set dynamically to support `rexexp()` if a loaded spreadsheet was an ODS spreadsheet has some value, because then it would be possible to calculate formulae using that function without getting a "function not found" error.

[...]

PR #4390, opened 2025-03-02
Add Ability to Add Custom Functions to Calculation

oleibman:
For an overview of why this is desired (and ways that people have coped with its absence), see issue #2900 and issue #4048; also PR #4043 which will be superseded by this PR.

The list of Excel functions is moved from Calculation/Calculation to its own member. I believe that it is done in a way that will not cause big complications to two experiments from @MarkBaker (PR #2714 and PR #2734). I believe it is also done in such a way that further refactoring of Calculation can follow this model.

Custom functions can be added or removed from the function list. You cannot add a function if it already exists in the list, and you cannot remove a non-custom function from the list. They will, of course, not be understood by Excel if written to a spreadsheet; the use case is mainly using the Calculation engine outside of spreadsheet context.

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
4 new false positives from Scrutinizer, which doesn't seem to understand how "protected" works.

PR #4412, opened 2025-03-19
Conditional and table formatting support for html

### LINKED ISSUE #1375: PDF-Generation don't respect condtional formatting

mikkokoo:
This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Conditional and table formatting support for html writer were needed to provide better html output for processed files.

Fix #1375.


tests/PhpSpreadsheetTests/Writer/Html/HtmlColourScaleTest.php
```diff
@@ -0,0 +1,41 @@
...
...0000;color:#000000;font-family:\'Arial\';font-size:11pt;background-color:#57BB8A;">2</td>            <td class="column5 style1 n" style="vertical-align:bottom;border-bottom:none #000000;border-top:none #000000;border-left:none #000000;border-right:none #000000;color:#000000;font-family:\'Arial\';font-size:11pt;background-color:#E67C73;">9</td>            <td class="column6 style1 n" style="vertical-align:bottom;border-bottom:none #000000;border-top:none #000000;border-left:none #000000;border-right:none #000000;color:#000000;font-family:\'Arial\';font-size:11pt;background-color:#E67C73;">9</td>            <td class="column7 style1 n" style="vertical-align:bottom;border-bottom:none #000000;border-top:none #000000;border-left:none #000000;border-right:none #000000;color:#000000;font-family:\'Arial\';font-size:11pt;background-color:#E67C73;">9</td>            <td class="column8 style2 null"></td>            <td class="column9 style1 n" style="vertical-align:bottom;border-bottom:none #000000;border-top:none #000000;border-left:none #000000;border-right:none #000000;color:#000000;font-family:\'Arial\';font-size:11pt;background-color:#E67C73;">10</td>          </tr>    </tbody></table>';
```
oleibman, line 37: This is excellent work on your part I will want to move it forward, but some changes are needed.

This test (and others like it in your PR) is kind of meaningless. I cannot look at the expected result and tell at a glance what, if anything, is affected by your change. Let's say, for example, that you have a cell which would normally have a default background, but, due to conditional formatting, will have a yellow background. Your test should be asserting what styles are applied to the cell, and should assert which one of them gives the yellow background. You will, of course, need to do this for several cells, including some which are not conditionally styled. This is not as easy as the one massive assertion, but will make for a much better test.

src/PhpSpreadsheet/Style/ConditionalFormatting/CellMatcher.php
```diff
@@ -213,6 +214,15 @@ protected function processOperatorComparison(Conditional $conditional): bool
         return $this->evaluateExpression($expression);
     }
 
+    protected function processColorScale(Conditional $conditional): bool
```
oleibman, line 217: The parameter $conditional is unused in this method. Should it be removed, or do you need to add some code which uses it?

src/PhpSpreadsheet/Worksheet/Table/TableStyle.php
```diff
@@ -198,6 +203,34 @@ public function setShowColumnStripes(bool $showColumnStripes): self
...
+     * Set this Style's Dxfs TableStyle.
+     */
+    public function setTableDxfsStyle(TableDxfsStyle $tableStyle, array $dxfs): self
+    {
+        $this->tableStyle = $tableStyle;
+
+        if ($this->tableStyle->getHeaderRow() !== null && isset($dxfs[$this->tableStyle->getHeaderRow()])) {
+            $this->tableStyle->setHeaderRowStyle($dxfs[$this->tableStyle->getHeaderRow()]);
+        }
+        if ($this->tableStyle->getFirstRowStripe() != null && isset($dxfs[$this->tableStyle->getFirstRowStripe()])) {
```
oleibman, line 224: Shouldn't this test be `!== null`?

src/PhpSpreadsheet/Worksheet/Table/TableStyle.php
```diff
@@ -198,6 +203,34 @@ public function setShowColumnStripes(bool $showColumnStripes): self
...
+            $this->tableStyle->setFirstRowStripeStyle($dxfs[$this->tableStyle->getFirstRowStripe()]);
+        }
+        if ($this->tableStyle->getSecondRowStripe() != null && isset($dxfs[$this->tableStyle->getSecondRowStripe()])) {
```
oleibman, line 227: Shouldn't this test be `!== null`?

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -473,20 +474,31 @@ public function generateSheetData(): string
 
         // Loop all sheets
         $sheetId = 0;
+
+        $activeSheet = $this->spreadsheet->getActiveSheetIndex();
+        $selectedCells = null;
```
oleibman, line 479: I believe this statement is unneeded.

src/PhpSpreadsheet/Writer/Html.php
```diff
@@ -473,20 +474,31 @@ public function generateSheetData(): string
...
             // Write table header
             $html .= $this->generateTableHeader($sheet);
             $this->sheetCharts = [];
             $this->sheetDrawings = [];
-
+            $condStylesCollection = $sheet->getConditionalStylesCollection();
+            foreach ($condStylesCollection as $condStyles) {
+                foreach ($condStyles as $key => $cs) {
+                    if ($cs->getConditionType() === Conditional::CONDITION_COLORSCALE) {
+                        $scale = $cs->getColorScale()->setScaleArray();
```
oleibman, line 491: You do not use `$scale` except in this statement. Is it needed?

tests/PhpSpreadsheetTests/Shared/StringHelperInvalidCharTest.php
```diff
@@ -32,6 +32,11 @@ public function testInvalidChar(): void
 
         $sheet->fromArray($array);
         $row = 0;
+
+        if ($sheet->getCell("B$row")->getValue() != "{$substitution}{$substitution}{$substitution}{$substitution}{$substitution}{$substitution}") {
```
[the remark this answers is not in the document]
mikkokoo: Sorry, I forgot to mention that this test failed out-of-the-box for me as well and I couldn't come up with a fix quickly, so I skipped it. Pls, let me know how to handle it.

src/PhpSpreadsheet/Style/ConditionalFormatting/CellMatcher.php
mikkokoo: I left it in because the other processXXX functions used similar syntax.

I can remove it or write a short check `$conditional->checkColorScale()` that would check that color scale exists and is properly set up. Check would add robustness if some colors/limit values are missing. Let me know.

Is there a logging system outside the Calculation engine btw?

src/PhpSpreadsheet/Writer/Html.php
mikkokoo: no

tests/PhpSpreadsheetTests/Writer/Html/HtmlColourScaleTest.php
mikkokoo: These were meant to test the Html Writer.

I can see that color scale and table functionality should be separately tested in order for them to be used and usable for the rest of the project. With tests similar to say `CellMatcherTest.php`.

tests/PhpSpreadsheetTests/Shared/StringHelperInvalidCharTest.php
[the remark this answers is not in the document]
oleibman: These tests are important and need to not be skipped. It is surprising that they fail because they don't seem to have anything to do with your change. What failure messages did you see when they weren't skipped.

tests/PhpSpreadsheetTests/Writer/Html/HtmlColourScaleTest.php
oleibman: I don't have a problem with your making tests against the generated html. You just need to make multiple targeted tests rather than one test for everything.

oleibman: As an example of what I'm looking for, in HtmlConditionalFormattingTest, cell H2 should have a value of 90 and a yellow background. You can test this as follows (remember that the Html uses 0-origin so we want row 1 and column 7 for H2):
```php
$data = $writer->generateHtmlAll();
// extract row 1 into $matches
self::assertSame(1, preg_match('~<tr class="row1".+?</tr>~ms', $data, $matches));
// Here is what we expect to see for column 7 in that row.
// We are particularly interested that the background-color (FFEB9C) and value (90) are correct.
// The string is broken into multiple pieces for legibility.
$expected = '<td class="column7 style1 n" style="'
    . 'vertical-align:bottom;'
    . 'border-bottom:1px solid #000000 !important;'
    . 'border-top:1px solid #000000 !important;'
    . 'border-left:1px solid #000000 !important;'
    . 'border-right:1px solid #000000 !important;'
    . 'color:#9C5700;font-family:\'Arial\';font-size:11pt;'
    . 'background-color:#FFEB9C;">90</td>';
self::assertStringContainsString($expected, $matches[0]);
```

tests/PhpSpreadsheetTests/Shared/StringHelperInvalidCharTest.php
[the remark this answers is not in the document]
mikkokoo:
```
There was 1 failure:

1) PhpOffice\PhpSpreadsheetTests\Shared\StringHelperInvalidCharTest::testInvalidChar
Osmanya as surrogate pair (x)
Failed asserting that two strings are identical.
--- Expected
+++ Actual
@@ @@
-'������'
+'����'

/home/mikko/src/pss/tests/PhpSpreadsheetTests/Shared/StringHelperInvalidCharTest.php:43
```

tests/PhpSpreadsheetTests/Writer/Html/HtmlColourScaleTest.php
mikkokoo: modified and pushed new tests

tests/PhpSpreadsheetTests/Shared/StringHelperInvalidCharTest.php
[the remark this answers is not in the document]
mikkokoo: I re-enabled this test in the latest commit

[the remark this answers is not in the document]
oleibman: Good, I am glad to see that this test is unchanged and still works correctly.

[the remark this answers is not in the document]
mikkokoo: Still fails for me though

tests/PhpSpreadsheetTests/Writer/Html/HtmlConditionalFormattingTest.php
```diff
@@ -0,0 +1,103 @@
...
+        self::assertSame(1, preg_match('~<tr class="row' . $rowNumber . '".+?</tr>~ms', $this->data, $matches));
+        foreach ($expectedMatches as $i => $expected) {
+            self::assertStringContainsString($expected, $matches[0]);
+        }
+    }
+
+    public static function conditionalFormattingProvider(): array
+    {
+        return [
+            'row 0' => [0, ['<td class="column1 style1 s">Jan</td>',
```
oleibman, line 38: You are absolutely on the right track. However, the sheer number of these tests still makes it hard to figure out what is being tested. Is it possible for you to test just a few cells (e.g. one with a green background, one with a yellow background, and one with no background) rather than every cell in the spreadsheet? I have some suggestions on how you might go about doing this, but, based on your work so far, I don't think you need them (but let me know if you do).

oleibman:
Here is something like what I would like to see for HtmlTableFormatWithConditionalTest.
[HtmlTableFormatWithConditionalTest.zip](https://github.com/user-attachments/files/19413571/HtmlTableFormatWithConditionalTest.zip)

tests/PhpSpreadsheetTests/Writer/Html/HtmlConditionalFormattingTest.php
mikkokoo: Thanks, makes sense.

oleibman:
Please make a separate PR for your HashCode improvements.

src/PhpSpreadsheet/Style/ConditionalFormatting/CellMatcher.php
```diff
@@ -214,6 +215,15 @@ protected function processOperatorComparison(Conditional $conditional): bool
...
+    {
+        if ($this->wrapCellValue() !== '' && is_float($this->wrapCellValue()) && $conditional->getColorScale() !== null && $conditional->getColorScale()->colorScaleReadyForUse()) {
```
oleibman, line 220: If I understand your code correctly your first 2 conditions conditions can be combined - if it's a float, then it's not equal to null-string. And I also believe the 3rd and 4th conditions can also be combined with a null-safe operator.
```php
if (is_float($this->wrapCellValue()) && $conditional->getColorScale?->colorScaleReadyForUse())
```
However, I'm not entirely sure about is_float. It seems that wrapCellValue might also return int. If you don't want that to follow the same code path as float, then leave is_float as-is; however, if you do want int and float to follow the same path, you might use is_numeric instead.

tests/PhpSpreadsheetTests/Writer/Html/HtmlColourScaleTest.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        self::assertSame(1, preg_match('~<tr class="row' . $rowNumber . '".+?</tr>~ms', $this->data, $matches));
+        foreach ($expectedMatches as $i => $expected) {
+            self::assertStringContainsString($expected, $matches[0]);
+        }
+    }
+
+    public static function colourScaleProvider(): array
+    {
+        return [
```
oleibman, line 37: This is still an overwhelmingly large number of tests. Can you model it after the others and just test a few "interesting" cells in each row, and only for the data you're looking for (background color and value).

tests/PhpSpreadsheetTests/Writer/Html/HtmlDifferentConditionalFormattingsTest.php
```diff
@@ -0,0 +1,94 @@
...
+        if ($match !== 1) {
+            return 'unable to match column';
+        }
+
+        return $matches[0];
+    }
+
+    public function testConditionalFormattingRulesHtml(): void
+    {
+        $expectedMatches = [
```
oleibman, line 49: Again, this is just too many tests for me to absorb. Please just choose a representative sample. Maybe a couple from column 1 (all of this column seems to match whatever test it needs), the matching cell from column 2, one or two non-matching cells from column 2, and one or two cells from column 3 (these might be a little trickier because conditional formatting doesn't apply to them). It is not that the tests I'm asking you to remove aren't worthwhile - they most decidedly are. For that reason, I would suggest commenting them out rather than deleting them. I think the "missing" tests are better handled by also creating new samples, so that we can confirm visually that the html matches the original xlsx. I will add a comment to your ticket later on how that might be done.

oleibman:
As discussed above, I would like you to use your xlsx files for both unit tests and samples. Here is how you might go about it.
[htmlcondtosamples.zip](https://github.com/user-attachments/files/19458806/htmlcondtosamples.zip)

tests/PhpSpreadsheetTests/Writer/Html/HtmlColourScaleTest.php
mikkokoo: It is the test of different color scale formulas. There has to be quite a few data points to have meaningful results as the style is the only measurable data. They are all interesting. I'd say this is the bare minimum.

tests/PhpSpreadsheetTests/Writer/Html/HtmlDifferentConditionalFormattingsTest.php
mikkokoo: There is a different code path for each test and they use a different conditional case. They are lightning fast as they only check from the same loaded data. I don't understand the rationale of how removing some of them makes the regression testing better.

Making the xlsx to samples makes sense. But I'd still keep these.

oleibman:
Thank you for all the extra work. I need to do a final review, but should be able to merge this in the next day or two.

Issue #4435, opened 2025-04-02
load and save a ods or xlsx file with a cell containing formular including an external filepah on windows-system leads to a replacement [1] instead of the correct filepath


comfine:
=WENNFEHLER(SVERWEIS(D2;'file:///C:/Program Files (x86)/xampp/htdocs/cao-adresses.xlsx'#$Worksheet.A2:C99999;3;0); "")

will be converted to

=WENNFEHLER(SVERWEIS(D2;[1]worksheet!a2:C99999;3;0); "")

when loading the file with phpspreadsheet and save it again as a new file.

i attached a sample-file, that can be used for testing. cell containing the Formular is E2.

[broken.ods](https://github.com/user-attachments/files/19560051/broken.ods)

oleibman:
What release of PhpSpreadsheet and Php are you using? I get multiple errors trying to load and save it with the newest release.

As for the `[1]`, I don't think that is anything that PhpSpreadsheet would introduce on its own. However, if you load your ods spreadsheet in Excel, and save it to xlsx, then, yes, Excel will replace the file name in the formula with `[1]`, and resolve that in file externalLinks\_rels\externalLink1.xml.rels. We do not yet support the externalLinks files. So, if you loaded ods in Excel, saved it as xlsx, then used PhpSpreadsheet to load the xlsx file, you would see the replacement. Is it possible that this is how you encountered this problem?

comfine:
i am running PHP Version 8.2.12, and i think phpspreadsheet is 4.1.0 - i downloaded a fix that you provided in another issue, that i requested. but dont know how to retrieve the exact version info for that.

i dont load ods-files in excel. as in the other issue, i convert the ODS file using libreoffice command line to xlsx, than i load and save the xlsx with phpspreadsheet. i doublechecked if the the externallink gets lost in the conversion of libreoffice, but it doesnt. its stil there after the converson. only after load/save with phpspreadsheet it turns into [1].

but did i understand you correct, that it is just not supported yet?

oleibman:
Correct, it is not yet supported.

oleibman:
I use the following command, which should essentially be the same as yours:
```
"c:\Program Files\LibreOffice\program\soffice.com" --convert-to xlsx --outdir \temp c:\git\issue.4435.ods
```
When I now inspect \temp\issue.4435.xlsx, I see the following in sheet1.xml:
``` xml
<c r="E2" s="20" t="str">
<f aca="false">
[...]

PR #4436, opened 2025-04-04
Ods Reader No Datatype for Null Value

oleibman:
Issue #4435 was initially described incorrectly. While investigating the original description, I came upon this problem. Ods Reader is trying to set some cells to null without supplying a valid DataType to setValueExplicit, causing that method to throw an exception. Reader is changed to no longer call that method when value is null and DataType is null-string.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
No idea why Scrutinizer error-ed. Don't really care.

PR #4440, opened 2025-04-08
copy row

torresmald:
This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [ ] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -3870,4 +3870,36 @@ class Worksheet
 
         return true;
     }
+
+
+     /**
+      * Copy a row, including style, data and height
+      *
+      * @param int $from row that will be copied
+      * @param int $to row that will receive the copy
+      * @return void
+      */
+      public function copyRow($from, $to) {
+        $this->getRowDimension($to)->setRowHeight($this->getRowDimension($from)->getRowHeight());
+        $fromRowCellIterator = $this->getRowIterator($from,$from)->current()->getCellIterator();
+        try {
+          $fromRowCellIterator->setIterateOnlyExistingCells(true);
+          foreach ($fromRowCellIterator as $fromCellIt) {
+            $fromCoordinates = $fromCellIt->getCoordinate();
+            $toCellCoordinates = $fromCellIt->getColumn().$to;
+            // Cache data as trying to save the cells as variables doesn't work
+            $style = $this->getCell($fromCoordinates)->getStyle();
+            $value = $this->getCell($fromCoordinates)->getValue();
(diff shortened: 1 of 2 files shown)


oleibman:
You can probably clean up the phpcs and php-cs-fixer errors with:
```
composer fix
```

src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -3870,4 +3870,36 @@ public function applyStylesFromArray(string $coordinate, array $styleArray): boo
...
+
+     /**
+      * Copy a row, including style, data and height
+      *
+      * @param int $from row that will be copied
+      * @param int $to row that will receive the copy
+      * @return void
+      */
+      public function copyRow($from, $to) {
+        $this->getRowDimension($to)->setRowHeight($this->getRowDimension($from)->getRowHeight());
```
oleibman, line 3883: Let's say you want to copy from row 1, which has cells A1/C1/E1, to row 2 which has cells A2/B2/F2. I think your code will replace A2/C2/E2 but leave B2/F2 unchanged, which is probably not what you want - you need to clear B2/F2.

src/PhpSpreadsheet/Worksheet/Worksheet.php
```diff
@@ -3870,4 +3870,36 @@ public function applyStylesFromArray(string $coordinate, array $styleArray): boo
...
+        $fromRowCellIterator = $this->getRowIterator($from,$from)->current()->getCellIterator();
+        try {
+          $fromRowCellIterator->setIterateOnlyExistingCells(true);
+          foreach ($fromRowCellIterator as $fromCellIt) {
+            $fromCoordinates = $fromCellIt->getCoordinate();
+            $toCellCoordinates = $fromCellIt->getColumn().$to;
+            // Cache data as trying to save the cells as variables doesn't work
+            $style = $this->getCell($fromCoordinates)->getStyle();
+            $value = $this->getCell($fromCoordinates)->getValue();
```
oleibman, line 3892: This might not be right for formulas. Let's say cell K3 has the value 10, and A1 has the value `=1+K3`. In Excel, if I now copy row 1 to row 2, cell A2 will have the value `=1+K4`. This is probably how your method should behave. It can get tricky. Rather than trying to re-invent the wheel, your loop should probably call `copyCells` for each cell that you want to copy. That will (I hope) update the formulas and take care of the styles.

oleibman:
In addition to other comments, you will need to add unit tests to demonstrate that your function works correctly.

oleibman:
According to author, it was unintentional to make this a pull request. Closing.

PR #4441, opened 2025-04-09
Update Sum.php

PavelZul:
We have a cell in the example.xlsx sheet containing a total sum as a formula: E1 = IF(SUM(A1:B1) = SUM(C1:D1), "true", "false"), where:

A1 = 1.1,

B1 = 2.7,

C1 = 1.8,

D1 = 2.

We expect the calculation result to be E1 = "true", but instead, we get E1 = "false" because in PHP, the addition yields:

1.1 + 2.7 = 3.8000000000000003,

1.8 + 2 = 3.8.

The idea of the fork is to accept an input parameter specifying the number of decimal places and round the final sum to that precision before comparison.

Additional Notes:
This issue arises due to floating-point precision errors in programming languages like PHP.

A possible solution is to use rounding (e.g., round(SUM(A1:B1), N) == round(SUM(C1:D1), N) where N is the desired decimal precision).

This is:

- [ ] a bugfix

### DIFF
--- a/src/PhpSpreadsheet/Calculation/MathTrig/Sum.php
+++ b/src/PhpSpreadsheet/Calculation/MathTrig/Sum.php
@@ -47,12 +47,19 @@ class Sum
      */
     public static function sumErroringStrings(mixed ...$args): float|int|string|array
     {
-        $returnValue = 0;
+        $returnValue = $maxDecimals = 0;
         // Loop through the arguments
         $aArgs = Functions::flattenArrayIndexed($args);
         foreach ($aArgs as $k => $arg) {
             // Is it a numeric value?
             if (is_numeric($arg)) {
+                // We count the number of signs after aim
+                $parts = explode('.', (string)$arg);
+                if (isset($parts[1])) {
+                    $decimals = strlen(rtrim($parts[1], '0'));
+                    $maxDecimals = max($maxDecimals, $decimals);
+                }
+                
                 $returnValue += $arg;
             } elseif (is_bool($arg)) {
                 $returnValue += (int) $arg;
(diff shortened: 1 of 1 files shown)


oleibman:
Sorry, this will probably not be implemented. You are hitting too small a target in a game of whack-a-mole. Your fix works fine for:
```
SUM(A1:B1)=SUM(C1,D1)
```
It does nothing for:
```
(A1+B1)=(C1+D1)
```
And if you patch that, there are many other possibilities which will continue to fail.

There are risks in using floating-point. These are well documented in the Php manual, and manuals for most other languages. If you want to compare floats, you'll have to take precautions. For example, your cell formulas could use the following:
```
ROUND(SUM(A1:B1,5))=ROUND(SUM(C1:D1,5))
or
ABS(SUM(A1:B1)-SUM(C1:D1)) < 0.00001
```
(Where 5 and 0.00001 are, of course, totally arbitrary choices.)

oleibman:
Now that I think about it, your problem really isn't with the SUM function, it with the `=` operator. Perhaps it might be possible to introduce a 'tolerance' property to Calculation and use that when testing numerics for (in)equality. I will give that some thought.

oleibman:
There is already a tolerance (1.0E-12) built into comparison operations, so no need for a change. I will, however, add that, because of this already existing constant, I do not duplicate your results:
```php
$sheet->getCell('A1')->setValue(1.1);
$sheet->getCell('B1')->setValue(2.7);
$sheet->getCell('C1')->setValue(1.8);
$sheet->getCell('D1')->setValue(2);
$sheet->getCell('E1')->setValue('= IF(SUM(A1:B1) = SUM(C1:D1), "true", "false")');
var_dump($sheet->getCell('E1')->getCalculatedValue());
```
My result:
```
string(4) "true"
```
Do you get a different result with this code? If so, please let me know Php and PhpSpreadsheet versions.

PavelZul:
Have a great day! I'll check it on my side and send you an exact example a bit later.

oleibman:
No update in 3 months, unable to duplicate problem, closing.

PR #4474, opened 2025-05-17
WIP Bug in resizeMatricesExtend

### LINKED ISSUE #4451: `Callculation::resizeMatricesExtend()` overwrites all rows in the smaller matrix.
# The bug

Calling `Callculation::resizeMatricesExtend()` with matrices of different sizes overwrites **all** rows in the smaller matrix with the value found in the last row.  I haven't found the expected behaviour of the method (comments of the method don't specify it), but I think the idea was to extend the smaller matrix with values found in the last row, not to overwrite all of it with the value found in the last row.

Example:
```php
<?php
declare(strict_types=1);

use PhpOffice\PhpSpreadsheet\Calculation\Calculation;

// Sample matrices to test with
$matrix1 = [[1], [3]];
$matrix2 = [[5], [8], [11]];

// Use reflection to make the protected method accessible
$calculation = new Calculation();
$reflectionMethod = new ReflectionMethod(Calculation::class, 'resizeMatricesExtend');
$reflectionMethod->setAccessible(true);

// Call the method using reflection
$reflectionMethod->invokeArgs($calculation, [&$matrix1, &$matrix2, count($matrix1), 1, count($matrix2), 1]);

echo "Matrix 1 after:\n";
print_r($matrix1);
```

(Presumably) Expected output:
```php
$matrix1 = [ [1], [3], [3] ];
```
Actual output
```php
$matrix1 = [ [3], [3], [3] ];
```

Relevant code:
https://github.com/PHPOffice/PhpSpreadsheet/blob/fb2cfed070f16091e842a2b027d2d15c5a39e2f9/src/PhpSpreadsheet/Calculation/Calculation.php#L871-L876

Note that the analogous code for columns does not overwrite the previous columns, it extends them with the last value:
https://github.com/PHPOffice/PhpSpreadsheet/blob/fb2cfed070f16091e842a2b027d2d15c5a39e2f9/src/PhpSpreadsheet/Calculation/Calculation.php#L863-L870

# Proposed change
```php
            if ($matrix1Rows < $matrix2Rows) {
                $x = $matrix1[$matrix1Rows - 1];
                for ($i = matrix1Rows ; $i < $matrix2Rows; ++$i) {
                    $matrix1[$i] = $x;
                }
            }
```

Please note that analogous change should be done for case where `$matrix2Rows < $matrix1Rows`.

I can create a PR if the change is accepted.

# Why it matters

We found that some formulas that work for us in Excel result in `#CALC!` in PhpSpreadsheet. As an example, the following type of formulas result in `#CALC!`: `=FILTER(Products!M:M;(Products!K:K=D5)*(Products!L:L=E5))`.  `Products!K:K=D5` and `Products!L:L=E5` generate matrices of different sizes (for whatever reason). PhpSpreadsheet extends the smaller matrix and overwrites all of its values. In effect, `FILTER()` fails to match any rows and results in `#CALC!`

# Expected result

I don't know if the above behaviour is expected, but I suspect it's a bug. I haven't found any description of this behaviour. If it's not a bug, documenting it might be a good idea.

oleibman:
Fix #4451. The reporter's analysis is correct - the method has a bug. The solution is relatively easy. Nevertheless, proving that the fix works as desired is tricky. The submitter's Reflection test is good, but it would be better to find one within Excel. My additional tests are contrived, for reasons explained in the test member. So, I will leave this as a work in progress while I search for something better. If that search doesn't succeed, I will nevertheless merge this in about a month.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

  - [x] Changes are covered by existing unit tests (SUMPRODUCT and Gnumeric and Ods ArrayFormulaTest)
- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Calculation/Calculation.php
+++ b/src/PhpSpreadsheet/Calculation/Calculation.php
@@ -845,15 +845,15 @@ class Calculation extends CalculationLocale
         if (($matrix2Columns < $matrix1Columns) || ($matrix2Rows < $matrix1Rows)) {
             if ($matrix2Columns < $matrix1Columns) {
                 for ($i = 0; $i < $matrix2Rows; ++$i) {
-                    $x = $matrix2[$i][$matrix2Columns - 1];
+                    $x = ($matrix2Columns === 1) ? $matrix2[$i][0] : null;
                     for ($j = $matrix2Columns; $j < $matrix1Columns; ++$j) {
                         $matrix2[$i][$j] = $x;
                     }
                 }
             }
             if ($matrix2Rows < $matrix1Rows) {
-                $x = $matrix2[$matrix2Rows - 1];
-                for ($i = 0; $i < $matrix1Rows; ++$i) {
+                $x = ($matrix2Rows === 1) ? $matrix2[0] : array_fill(0, $matrix2Columns, null);
+                for ($i = $matrix2Rows; $i < $matrix1Rows; ++$i) {
                     $matrix2[$i] = $x;
                 }
             }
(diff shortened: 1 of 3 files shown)


sayur-dev:
❤️

oleibman:
Month is up. I have made no progress on a better test case, but need to move forward with the fix anyhow.

Issue #3899, opened 2024-02-11
Non-backwards compatible rounding issue in 2.0.0 - DateTime


BGehrels:
This is:

```
- [x] a bug report
- [ ] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
When i persist and reread an excel sheet, the data within the read version should be the same as the data persisted.
This has been the case with 1.x, but it is not with 2.0.0 if Date::excelToDateTimeObject / Date::dateTimeToExcel is used

### What is the current behavior?
When i use Date::dateTimeToExcel to write a DateTime in to a Cell, and then immediately reread it from the in-memory spreadsheet using Date::excelToDateTimeObject, then it works fine.

If i persist it to a file and reread it from there it will be slightly off: 

> -2020-10-21T14:55:31.000000+0000
> +2020-10-21T14:55:30.999994+0000

If you throw away milliseconds afterwards, for example during database persistence, as you end up a second off (which can land you in another hour of the day, another aggregation bucket, and so on)

### What are the steps to reproduce?
The test posted below works fine in 1.29.0x, but not in 2.0.0

Please provide a [Minimal, Complete, and Verifiable example](https://stackoverflow.com/help/mcve) of code that exhibits the issue without relying on an external Excel file or a web server:

```php
<?php

use PHPUnit\Framework\TestCase;
use DateTime;
use PhpOffice\PhpSpreadsheet\Shared\Date;
use PhpOffice\PhpSpreadsheet\Spreadsheet;
use PhpOffice\PhpSpreadsheet\Writer\Xlsx;

class ExcelDateTimeOfByOneErrorTest extends TestCase
{
    public function testThatDateTimesCanBePersistedAndReread(): void
    {
        $originalDateTime = new DateTime('2020-10-21T14:55:31');

        $dateTimeFromSpreadsheet = $this->getDateTimeFrom($this->excelSheetWithDateTime($originalDateTime));
        $dateTimeFromSpreadsheetAfterPersistAndReread = $this->getDateTimeFrom($this->persistAndReread($this->excelSheetWithDateTime($originalDateTime)));

        static::assertEquals($originalDateTime, $dateTimeFromSpreadsheet);
        static::assertEquals($originalDateTime, $dateTimeFromSpreadsheetAfterPersistAndReread);
    }

[...]

PR #4476, opened 2025-05-19
fix rounding issues with microseconds calculation

blizzz:
This is:

- [X] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [X] Code style is respected

A bug was reported with rounding issues (#3899) that is said to be connected to #3677 and I also experience a failing test downstream, when a stored time of  13:37 is interpreted as 13:36 due to the same issues.

Modifications to the rounding strategies can vastly improve the situation by successfully testing existing cases as well as satisfying new tests, based on the failure I experience, as well as the test provided in #3899 (I included @BGehrels as co-author in that respective commits).

I did not play around much with the GMP extension, but this might also be an alternative for more precise calculation? With this here we lose the .1000000 part of a second, which however already suffers to inaccuracies, hence it is acceptable.

There is one failure I receive, though I suppose this is due to how I modified the test excel sheet. Tried with both LibreOffice as well as some MS Office Online thing, but it does not make a difference. The only (knowing) change i did was adding the time `13:37` to cell `F3` and formatted it to show hours and minutes.


oleibman:
Do not replace explicitdate.xlsx. The tests involving it still need to work after your change, and, as you can see, they don't.

If you have a different spreadsheet which demonstrates a problem with the existing code and which your PR fixes, please add that spreadsheet to your PR and test using it.

oleibman:
I am unsure of the purpose of:
```php
$microseconds = (int) round($microseconds, -2);
```
If, for example, we start with a dayPart value of 0.1234567, microseconds is initially 658880. Then you apply this magic rounding to change it to 658900. Please explain why.

oleibman:
Actually, the problem is not with the microseconds calculation. The problem is that Php is not generating the full value for a float when casting to string. The calculation yields a result of `44125.62188657407387`, but, when Php casts this to a string during the write process, it casts it to `44125.621886574`, dropping the last 5 digits, representing 1E-10 and below. The unfortunate part is that 1 microsecond is represented as 1/86400/1000000, which is about 1.1574E-11, which is below the threshold that Php is dropping.

Subject to the usual warning that floating-point arithmetic is imperfect (and this is exacerbated by the very small numbers we're interested in), it is possible that the Writer can do better. I will look into it.

oleibman:
This PR is superseded by #4479. It is now converted to draft status, and will be closed if 4479 is merged.

oleibman:
PR 4479 is implemented. This can be closed.

PR #4487, opened 2025-05-26
Various Writers RichText TextElement Should Inherit Cell Style

### LINKED ISSUE #1154: Inconsistent cell styles on RichText values.
This is:
```
- [x] a bug report
- [ ] a feature request
```

Consistency among ```TextElement``` styles of the same ```RichText``` value after applying a cell style. Desired behavior would be one of the following:
  1) Overwrite any applicable ```TextElement``` and ```Run``` styles.
  2) Overwrite any applicable ```TextElement``` styles, and ignore all ```Run``` styles.
  3) Ignore all ```TextElement``` and ```Run``` styles.

Inconsistency among ```TextElement``` styles of the same ```RichText``` value after applying a cell style. Current behavior is that setting a cell style only overwrites any applicable ```TextElement``` styles found before a ```Run``` style; any ```TextElement``` styles found after a ```Run``` style are ignored.

```php
<?php
require_once 'vendor/autoload.php';

use \PhpOffice\PhpSpreadsheet\Spreadsheet;
use \PhpOffice\PhpSpreadsheet\RichText\RichText;
use \PhpOffice\PhpSpreadsheet\Writer\Xlsx;

$workbook = new Spreadsheet();

$rtf = new RichText();

/* This segment will receive applied cell styles. */
$rtf->createText('~Cell Style~');

/* This segment ignore applied cell styles. */
$rtf->createTextRun('~RTF Style~')->getFont()->setItalic(true);

/* This segment will also ignore applied cell styles. */
$rtf->createText('~No Style~');

$sheet = $workbook->getSheet(0);
$sheet->getCell('A1')->setValue($rtf);
$sheet->getStyle('A1')->getFont()->setBold(true);

(new Xlsx($workbook))->save('output.xlsx');

$workbook->disconnectWorksheets();
unset($workbook);
```

PhpSpreadsheet [1.9]
PHP [7.1.27]

oleibman:
Fix #1154, which went stale over 5 years ago, and is now reopened. RichText elements can be a Run, which sets its own style, or a TextElement, which doesn't. As the issue states, TextElement handling is inconsistent. This PR forces it to inherit the style of the cell. As implemented, this will change it to a Run when it is read in, but there should be no practical difference as far as the end-user is concerned. There is no change as far as Run is concerned. The user suggested 3 options - TextElement and Run both inherit for any unspecified style elements, TextElement inherits and Run does not, neither inherits. Option 2 makes most sense to me. Option 1 may not even be possible (we can't tell if, say, the user explicityl set Italic to false, or if that was just the default choice).

Tests are added for all of Xlsx, Xls, and Html. Html reading of RichText elements is not well-supported. The tests are fairly difficult to understand. A new sample is added to demonstrate this change.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Unsure what the runtime problem is with Xls Writer. Unable to duplicate problem on my own system. Converting to draft while I research.

PR #4499, opened 2025-06-02
refactor: cleanup code twice calling

ddevsr:
This is:

- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Just refactoring trimming twice.

### DIFF
--- a/src/PhpSpreadsheet/Spreadsheet.php
+++ b/src/PhpSpreadsheet/Spreadsheet.php
@@ -608,7 +608,7 @@ class Spreadsheet implements JsonSerializable
      */
     public function removeSheetByIndex(int $sheetIndex): void
     {
-        $numSheets = count($this->workSheetCollection);
+        $numSheets = $this->getSheetCount();
         if ($sheetIndex > $numSheets - 1) {
             throw new Exception(
                 "You tried to remove a sheet by the out of bounds index: {$sheetIndex}. The actual number of sheets is {$numSheets}."
@@ -660,10 +660,9 @@ class Spreadsheet implements JsonSerializable
      */
     public function getSheetByName(string $worksheetName): ?Worksheet
     {
-        $worksheetCount    = count($this->workSheetCollection);
         $trimWorksheetName = trim($worksheetName, "'");
 
-        for ($i = 0; $i < $worksheetCount; ++$i) {
+        for ($i = 0; $i < $this->getSheetCount(); ++$i) {
             if (strcasecmp($this->workSheetCollection[$i]->getTitle(), $trimWorksheetName) === 0) {
                 return $this->workSheetCollection[$i];
             }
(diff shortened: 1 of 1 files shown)


src/PhpSpreadsheet/Spreadsheet.php
```diff
@@ -608,7 +608,7 @@ public function addSheet(Worksheet $worksheet, ?int $sheetIndex = null, bool $re
      */
     public function removeSheetByIndex(int $sheetIndex): void
     {
-        $numSheets = count($this->workSheetCollection);
+        $numSheets = $this->getSheetCount();
```
oleibman, line 611: This change seems to add a small amount of overhead through an additional function call. What compensatory advantage do you see that would make this worthwhile?

src/PhpSpreadsheet/Spreadsheet.php
```diff
@@ -660,9 +660,10 @@ public function getAllSheets(): array
      */
     public function getSheetByName(string $worksheetName): ?Worksheet
     {
-        $worksheetCount = count($this->workSheetCollection);
-        for ($i = 0; $i < $worksheetCount; ++$i) {
-            if (strcasecmp($this->workSheetCollection[$i]->getTitle(), trim($worksheetName, "'")) === 0) {
+        $trimWorksheetName = trim($worksheetName, "'");
+
+        for ($i = 0; $i < $this->getSheetCount(); ++$i) {
```
oleibman, line 665: This absolutely adds overhead by moving a function call from outside the loop to inside. It will definitely not be approved.

src/PhpSpreadsheet/Spreadsheet.php
```diff
@@ -660,9 +660,10 @@ public function getAllSheets(): array
...
+            if (strcasecmp($this->workSheetCollection[$i]->getTitle(), $trimWorksheetName) === 0) {
```
oleibman, line 666: This one, however, does improve things a bit by moving a function call from inside the loop to outside. I would certainly be okay with including this part.

src/PhpSpreadsheet/Spreadsheet.php
```diff
@@ -790,8 +791,8 @@ public function setActiveSheetIndexByName(string $worksheetName): Worksheet
     public function getSheetNames(): array
     {
         $returnValue = [];
-        $worksheetCount = $this->getSheetCount();
-        for ($i = 0; $i < $worksheetCount; ++$i) {
+
+        for ($i = 0; $i < $this->getSheetCount(); ++$i) {
```
oleibman, line 795: Once again, you are moving a function call from outside the loop to inside. This part of the change will not be approved.

src/PhpSpreadsheet/Spreadsheet.php
oleibman: Reverted in latest commit.

src/PhpSpreadsheet/Spreadsheet.php
oleibman: Reverted in latest commit.

src/PhpSpreadsheet/Spreadsheet.php
oleibman: Reverted in latest commit.

oleibman:

PR #4503, opened 2025-06-06
AutoColor for LibreOffice Dark Mode

oleibman:
See discussion #4502. I am not sure of what needs to be involved here. This is especially so because Excel also offers AutoColor, but its dark mode looks significantly different than LibreOffice's.

So here's what I've come up with thus far.
- Add boolean autoColor with setter and getter to Font.
- If autoColor is true, ODS writer and XLSX Writer will record it in the xml, and will not write a font color. It is possible that both can co-exist; I'm just not sure how they are supposed to interact if that is the case.
- It makes most sense to me if the spreadsheet's default font specifies autoColor, but PhpSpreadsheet will not insist on that.
- Xlsx Reader will process autoColor. Like most other styles, Ods Reader is not yet set up to handle it.
- LibreOffice Calc should follow the autoColor declarations when it reads either an Xlsx or Ods spreadsheet created by PhpSpreadsheet.

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xlsx/Styles.php
+++ b/src/PhpSpreadsheet/Reader/Xlsx/Styles.php
@@ -136,6 +136,12 @@ class Styles extends BaseParserClass
             $attr = $this->getStyleAttributes($fontStyleXml->scheme);
             $fontStyle->setScheme((string) $attr['val']);
         }
+        if (isset($fontStyleXml->auto)) {
+            $attr = $this->getStyleAttributes($fontStyleXml->auto);
+            if (isset($attr['val'])) {
+                $fontStyle->setAutoColor(self::boolean((string) $attr['val']));
+            }
+        }
     }
 
     private function readNumberFormat(NumberFormat $numfmtStyle, SimpleXMLElement $numfmtStyleXml): void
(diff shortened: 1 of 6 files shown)


fwiep:
Yes! For me, this works exactly as expected.

The generated text has no preset color and is therefore rendered in perfect contrast to the cell's background. Be it dark on light or (as in my case) light on dark.

For my own reference, here's how I tested this PR manually:

1. clone the PhpSpreadsheet Github repository (`git clone "https://github.com/PHPOffice/PhpSpreadsheet.git"`)
2. change into the directory (`cd PhpSpreadsheet`)
3. install dependencies (`composer install`)
4. download PR as patch file by appending `.patch` to its URL: `https://github.com/PHPOffice/PhpSpreadsheet/pull/4503.patch`
5. apply the patch: `patch -p 1 < /path/to/4503.patch`
6. add `$spreadsheet->getActiveSheet()->getStyle('A')->getFont()->setAutoColor(true);` to `samples/Basic/01_Simple.php`
7. run sample: `php -f samples/Basic/01_Simple.php`

In the generated sample, column `A` has the font color indeed set to Automatic (or at least, is not preset to black).

Thanks! Kind regards,
FWieP

PR #4513, opened 2025-06-18
Re-implement ASC function and add unit tests

internauticos:
This commit re-implements the `ASC` spreadsheet function, which converts full-width (double-byte) characters to their corresponding half-width (single-byte) characters. This work is being recommitted to a new branch to incorporate recent updates from the master branch.

The implementation resides in the `TextData` class, and the `Calculation` class has been updated to delegate calls to `ASC` to this implementation.

Unit tests have been included in `TextDataTest.php` to cover various scenarios:
- Full-width alphanumeric characters.
- Full-width symbols.
- Mixed full-width and half-width characters.
- Empty strings.
- Boolean and numeric inputs.
- Japanese Katakana and Hiragana characters.

This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [ ] Code style is respected


oleibman:
I thought I posted this earlier, but I don't see it. So I'll try again. I assume you will be able to straighten out your change so that it only changes a few members instead of the 3,121 (!) this affects. But, before you do, I'm interested in one of your test cases.
```
['ｶﾀｶﾅ', '=ASC("カタカナ")'], // Full-width Katakana to half-width
```
Here's what I see when I try this in Excel:
image

The output seems to match the input; it hasn't been changed from full-width to half-width. There may be many possible explanations for why this doesn't agree with your test case, including font or system options. Do you have any explanation for why my Excel doesn't match your expectation?

oleibman:
Likewise, I don't think I duplicate your results for mb_convert_kana. Here's what I see on both Windows and Linux.
image
The Latin double-width characters are converted, but not the katakana characters.

oleibman:
I'm sure there's some difference in our systems that accounts for the differences in both Excel and Php. Whatever that difference is, I would consider it a show-stopper - we don't want our results to vary from system to system. Additionally, I think ASC probably handles Chinese and Korean characters which this PR would not cover. So I am going to close this PR in favor of another approach which I am going to push very soon.

oleibman:
Superseded by PR #4544. That PR is not as complete a solution as you might like, but I hope it will be useful. For the reasons explained above, we cannot move forward with this PR.

PR #4514, opened 2025-06-19
Allow Xlsx Reader to Specify ParseHuge Master

### LINKED ISSUE #4260: BC since Settings::libXmlLoaderOptions is ignored
This is:

```
- [x] a bug report
- [ ] a feature request
```

Currently, we use the `Settings::setLibXmlLoaderOptions()` function to use the constant `LIBXML_PARSEHUGE`
While we added this 4 years ago and I can hardly find which files we are receiving is using this functionality, I would like to be sure that updating the library will not cause a BC on our side, since it was removed in the #4233 
Is it something that was considered, do we have another way to allow to parse huge XML files?

Huge XLXS files can be parsed with the library

- [x] Reader
- [ ] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

Format using XML files

oleibman:
Fix #4260. A number of Security Advisories related to libxml_options were opened. In the end, we disabled the ability to specify any libxml_options. However, some users were adversely affected because they needed LIBXML_PARSEHUGE for some of their files. Having finally obtained access to a file demonstrating this problem, we can restore this ability.
- The operation is potentially dangerous, a vector for memory leaks and out-of-memory errors. It is not recommended unless absolutely needed.
- It will not be permitted as a global (static) property with the ability to adversely affect other users on the same server.
- It will instead be implemented as an instance property of Xlsx Reader (default to false), with a setter. I do not see a use case for a getter.
- People will need to set this property individually for each file which they think needs it.
- This change will be backported to all supported releases.
- The sheer size and processing time for the file involved makes it impractical to add a formal test case. It has, nevertheless, been tested satisfactorily.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,13 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com)
 and this project adheres to [Semantic Versioning](https://semver.org).
 
-## TBD - 4.4.0
+## TBD - 4.5.0
 
 ### Added
 
-- VSTACK and HSTACK. [Issue #4485](https://github.com/PHPOffice/PhpSpreadsheet/issues/4485) [PR #4492](https://github.com/PHPOffice/PhpSpreadsheet/pull/4492)
-- TOCOL and TOROW. [PR #4493](https://github.com/PHPOffice/PhpSpreadsheet/pull/4493)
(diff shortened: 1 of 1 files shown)


oleibman:
No concern whatsoever with Scrutinizer "complexity" warning.

PR #4519, opened 2025-06-20
add tests to support issue 4476

blizzz:
This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [X] additional unit tests

Checklist:

- [ ] Code style is respected

This test shall proof that a time is interpreted wrongly, at least with XLSX. Cf. https://github.com/PHPOffice/PhpSpreadsheet/pull/4479#issuecomment-2968328399

Screenshot_20250620_120113


oleibman:
The test failure here is, as we already know, a consequence of floating point rounding. However, it can be coded to avoid a problem. You want the result to agree to minutes (and the same general discussion would apply if you wanted to agree to seconds). You can achieve that by doing the following before formatting the DateTime object:
```php
Date::roundMicroseconds($dt);
```

oleibman:
No update in 10 months. PhpSpreadsheet outputs the expected spreadsheet value for `getFormattedValue` - no rounding error. There is a rounding error when converting it using `excelToDateTimeObject`, and code which corrects that rounding error is suggested above, and works as expected. I don't think there's a reason to keep this PR open.

Issue #4522, opened 2025-06-28
Data loss for large numeric value like 19966412884001452


saifulferoz:
I'm using fromArray method to write data. 
But in generated excel file getting value 19966412884001400.

Which means given value exceed excels max int value.

In this case if value exceeds Max Int value then it should be string type for $cell->setValueExplicit() method

oleibman:
You can probably get the effect you want without any changes to PhpSpreadsheet.
```php
class DefaultValueBinderWithLimit extends DefaultValueBinder
{
    const EXCEL_MAX_INT = 9007199254740992; // 2^53, max safe integer in JavaScript

    public static function dataTypeForValue(mixed $value): string
    {
        if (is_numeric($value)) {
            $value += 0;
            if (abs($value) <= self::EXCEL_MAX_INT) {
                return DataType::TYPE_NUMERIC;
            }

            return DataType::TYPE_STRING;
        }

        return parent::dataTypeForValue($value);
    }
}

// mainline code here

        $spreadsheet = new Spreadsheet();
        $spreadsheet->setValueBinder(new DefaultValueBinderWithLimit());
```

saifulferoz:
Thanks @oleibman,  for your valuable feedback and solution. I'm aware about custom value binder. My intent was this function is used for to write data in excel. Excel doesn't support PHP_INT_MAX value as Integer so we should use EXCEL max supported integer value.

According to context & scope I prefer to use Excel supported Max Integer value.

oleibman:
Excel does not, to my knowledge, have a "max supported integer value". If you can point me to something which says otherwise, please do. Your proposal replaces what we are using (Php max integer value) with something that we are not using at all (Javascript max integer value).

oleibman:
I've been doing a little more research. The code that you want to change dates back all the way to PhpExcel. I think it was introduced in 2014, and I'm not even sure that 64-bit PHP (which would affect the value of PHP_INT_MAX) was in widespread use at the time. I have found multiple references that Excel maintains 15 digits of precision, so it seems like the comparison ought to be with `999_999_999_999_999` rather than PHP_INT_MAX. This would certainly satisfy your need, but it's a breaking change, and I'm not willing to do that. So, please go with your own Binder for now, and I will plan to change DefaultValueBinder appropriately for the next breaking release.

PR #4523, opened 2025-06-28
fix: data loss for numeric value greater than excel supported max value

saifulferoz:
This is:

- [ *] a bugfix

Checklist:

  - [*] Changes are covered by existing unit tests
- [ *] Code style is respected

This change is necessary because PHP_INT_MAX represents the maximum integer value supported by PHP on the current platform, which can be larger than what Excel or JavaScript can safely handle (specifically, numbers above 2^53 lose precision in JavaScript and Excel). By switching to self::EXCEL_MAX_INT (2^53), the code ensures that any integer value larger than what Excel can safely represent is treated as a string, preventing precision loss and data corruption when exporting to Excel. This makes the behavior consistent and safe across different platforms and environments.

Reference issue #4522

### DIFF
--- a/src/PhpSpreadsheet/Cell/DefaultValueBinder.php
+++ b/src/PhpSpreadsheet/Cell/DefaultValueBinder.php
@@ -64,7 +64,7 @@ class DefaultValueBinder implements IValueBinder
             return DataType::TYPE_INLINE;
         }
         if ($value instanceof Stringable) {
-            $value = (string)$value;
+            $value = (string) $value;
         }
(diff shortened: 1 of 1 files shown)


oleibman:
I am not inclined to merge this. I see no reason why PhpSpreadsheet, which clearly has dependencies on Php and Excel, also needs a dependency on Javascript. However, I will have other suggestions to make in the linked issue.

oleibman:
Superseded by PR #4527.

PR #4524, opened 2025-07-01
Recognize "New" Mimetype for Empty File

### LINKED ISSUE #4521: mimetype mismatch (PHP 8.3+) and missing mimetype (PHP 7.4+) for "empty" files (CsvReader)
https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Reader/Csv.php#L607 only lists `inode/x-empty`

But newer PHP and libmagic versions use `application/x-empty`.

So `application/x-empty` has to be added.

See also:
* https://github.com/php/php-src/pull/2670#issuecomment-2827542497
* https://3v4l.org/l8FBt
* https://3v4l.org/8OBkG

At the last link you can also see, that previously an empty file with a crlf (`\r\n`) resulted in `application/octet-stream` and results in `text/plain` for newer PHP versions. Maybe this can be also be covered in PhpSpreadsheet (e.g. by using `trim` and checking the content before doing the mimetype check)?

The ideal way would be, to treat "empty" files (which should proably also include files with only whitespace characters) as such (`application/x-empty`) instead of `application/octet-stream` by PHP itself - or the mimetype `text/plain` for files with `\r\n` (to align the behavior to PHP 8.3+) .

oleibman:
Fix #4521. Reader/Csv (which can be invoked by IOFactory::read), checks the mimetype of files if they don't have a csv/tsv extension. It allows empty files, whose mimetype was set to `inode/x-empty` for Php5.3.11 through Php7.3.33 (except for 5.4.0). For 5.3.1-5.3.10, 5.4.0, and 7.4.0+, the mimetype is `application/x-empty`. Reader/Csv recognizes the `inode` version but not the `application` version, which this PR adds.

The person who issued the report also noted that, for a file consisting of cr-lf, Php8.1.* and Php8.2.* report the mimetype as `application/octet-stream`, whereas all other releases report it as `text/plain`. The behavior of Php8.1/2 seems to be a bug, and I cannot possibly guess all the conditions that might lead to that bug. So I will not fix that problem. Anyone who is adversely affected by it can either add a `csv` extension to the filename, or upgrade to Php8.3+, or pre-process the file (e.g. to make it truly empty) before passing it to PhpSpreadsheet. A test case to document the problem is added for documentary purposes; the test will be skipped for Php8.1/2, but run for all other releases.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Reader/Csv.php
+++ b/src/PhpSpreadsheet/Reader/Csv.php
@@ -605,6 +605,7 @@ class Csv extends BaseReader
             'text/csv',
             'text/plain',
             'inode/x-empty',
+            'application/x-empty', // has now replaced previous
             'text/html',
         ];
 
--- a/tests/PhpSpreadsheetTests/Issue4521Test.php
+++ b/tests/PhpSpreadsheetTests/Issue4521Test.php
@@ -1,0 +1,56 @@
+<?php
+
+namespace PhpOffice\PhpSpreadsheetTests;
+
+use PhpOffice\PhpSpreadsheet\IOFactory;
+use PhpOffice\PhpSpreadsheet\Shared\File;
+use PHPUnit\Framework\TestCase;
+
+class Issue4521Test extends TestCase
+{
+    private string $outfile = '';
+
+    protected int $weirdMimetypeMajor = 8;
+
+    protected int $weirdMimetypeMinor1 = 1;
+
+    protected int $weirdMimetypeMinor2 = 2;
+
+    protected function tearDown(): void
+    {
+        if ($this->outfile !== '') {
+            unlink($this->outfile);
+            $this->outfile = '';
+        }
+    }
+
+    public function testEmptyFile(): void
+    {
+        $this->outfile = File::temporaryFilename();
+        file_put_contents($this->outfile, '');
+        $spreadsheet = IOFactory::load($this->outfile);
+        $sheet = $spreadsheet->getActiveSheet();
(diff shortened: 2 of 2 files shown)


DanielRuf:
@oleibman thank you.

Regarding the releases and fixes, will this also be backported to v1 or are users encouraged to upgrade to v4? Unfortunately I found no clear overview, which major version should be used in which case. I can not even find a SECURITY.md or part in README.md that tells if and how each major version is actively supported, reached EOL or is for specific PHP versions.

At least it does not seem to be some sort of Semantic Versioning, otherwise new additions would mean a new minor version (because I can see only one addition at https://github.com/PHPOffice/PhpSpreadsheet/releases/tag/1.29.12).

oleibman:
We encourage everyone to migrate to the latest release. Realizing that this isn't entirely practical, we currently support 4 earlier branches (these are documented in CONTRIBUTING.md). We currently backport all security fixes to all 4. We do not, however, backport most other changes. It makes the maintenance burden much too high for us.

oleibman:
And, in case you are wondering about the latest minimal backports, they are precursors to a breaking security change which is going to get me in all kinds of trouble with people concerned about Semantic Versioning. I don't see an alternative other than dropping support for the backported releases altogether.

DanielRuf:
Understood, thanks for the clarification.

Issue #4530, opened 2025-07-08
Problems with INDIRECT as Part of Range


oleibman:
Consider the following code:
```php
        $sheet->getCell('A1')->setValue(5);

        $sheet->getCell('B1')->setValue(2);
        $sheet->getCell('B2')->setValue(4);
        $sheet->getCell('B3')->setValue(6);
        $sheet->getCell('B4')->setValue(8);
        $sheet->getCell('B5')->setValue(10);
        $sheet->getCell('B6')->setValue(12);
        $sheet->getCell('C1')->setValue('=SUM(B1:INDIRECT("B"&A1))');
        var_dump($sheet->getCell('C1')->getCalculatedValue());

        $sheet->getCell('D1')->setValue(3);
        $sheet->getCell('D2')->setValue(6);
        $sheet->getCell('D3')->setValue(9);
        $sheet->getCell('D4')->setValue(12);
        $sheet->getCell('E1')->setValue('=SUM(D1:INDIRECT("D"&A1))');

        var_dump($sheet->getCell('E1')->getCalculatedValue());
```

The formula in C1 resolves to `=SUM(B1:B5)`, and is correctly evaluated as `30`.

The formula in E1 should resolve to `=SUM(D1:D5)`, however it is incorrectly evaluated as `3` rather than `30`. It appears that the Calculation Engine is evaluating it as `=SUM(D1:D1)`. The difference between this and the previous example appears to be that `B5` exists in the first case, but `D5` does not in the second.

I have used the Calculation debugger to try to see what's going wrong in the second case, but haven't yet reached any satisfactory conclusion.

This issue is a secondary problem for issue #4528.

oleibman:
This may be related to issue #4403.

oleibman:
This may be related to issue #4558.

PR #4531, opened 2025-07-08
Ods Reader Process Nested table-row

### LINKED ISSUE #2507: Reader/Ods.php - missed table-row-group
I have .ods file in which several `table-row` are inside `table-row-group`: 
```
<table:table-row-group>
    <table:table-row table:style-name="ro4">
        <table:table-cell table:style-name="ce2"/>

        <table:table-cell office:value-type="string" table:style-name="ce8">
            <text:p>some_text</text:p>
        </table:table-cell>

        <table:table-cell office:value="0.225" office:value-type="float" table:style-name="ce9">
            <text:p>0,225</text:p>
```

The code in Reader/Ods.php only considers 'table-row':
https://github.com/PHPOffice/PhpSpreadsheet/blob/dbaafba6c6dcf8c75a2c460d85758514ccb8e08f/src/PhpSpreadsheet/Reader/Ods.php#L358

As a result when reading the file the data is not pulled from `table-row` inside `table-row-group`.

### LINKED ISSUE #4528: Ods Reader wrong cell value
I have some wrong cell value when  i get getCell or getCalculatedValue.

Can it be that the library have some problem when i have merge cell ?

oleibman:
Fix #4528. Fix #2507. Ods Reader expects `table:table-row` to descend directly from `table:table`. But it turns out that it can be nested under `table:table-header-rows`, `table:table-rows`, or `table:table-row-group`. Similar considerations apply to `table:table-column`. Ods Reader is changed to process the nesting tags, and therefore the nested tags underneath them. This is a major change, but it doesn't seem to have broken any existing tests or samples.

No attempt is made to add any of the nesting tags in Ods Writer.

It should be noted that 4528 has an additional problem, which has been reported as issue #4530. It is not fixed by this PR.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,9 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 ### Fixed
 
 - Ods Reader Nested table-row. [Issue #4528](https://github.com/PHPOffice/PhpSpreadsheet/issues/4528) [Issue #2507](https://github.com/PHPOffice/PhpSpreadsheet/issues/2507) [PR #4531](https://github.com/PHPOffice/PhpSpreadsheet/pull/4531)
+- Recognize application/x-empty mimetype. [Issue #4521](https://github.com/PHPOffice/PhpSpreadsheet/issues/4521) [PR #4524](https://github.com/PHPOffice/PhpSpreadsheet/pull/4524)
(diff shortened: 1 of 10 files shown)


oleibman:
Isolating the table-row logic into its own subroutine predictably, and unavoidably, creates a new F-complexity function, and Scrutinizer doesn't like that. This is not a concern.

PR #4541, opened 2025-07-16
Html Writer Conditional Formatting Inline Css

### LINKED ISSUE #4539: XLSX reader HTML writer does not support conditional formatting if using inline CSS
If you use the XLSX reader and HTML writer with the inline CSS option, conditional formatting color scale support is lost.

```php
<?php

use PhpOffice\PhpSpreadsheet\Writer\Html;

$spreadsheet = IOFactory::load('example.xlsx');

$writer = new Html($spreadsheet);
$writer->setConditionalFormatting(true);
$writer->setUseInlineCss(true);

echo 'With `setUseInlineCss(true);`';

echo $writer->generateSheetData();

$writer->setUseInlineCss(false);

echo 'With `setUseInlineCss(false);`';

echo $writer->generateSheetData();
```

Steps to reproduce:

1. Download the `.zip` attached here
2. Run `php -S 0.0.0.0:8080`

[spreadsheet-test.zip](https://github.com/user-attachments/files/21244052/spreadsheet-test.zip)

oleibman:
Fix #4539. Conditional Formatting was recently added to Html Writer. It works fine when not using inline Css. However, when using inline Css, the code inadvertently added 2 different `style` attributes (one for the unconditional style and one for the conditional style) to the same cell. This is not valid html, and results in losing the conditional styling. This PR combines the two `style` attributes into one, which will now come after the `class`, `colspan`, and `rowspan` attributes.

Aside from the new tests, this PR changes an unusually large number of existing tests. While this might normally be considered a red flag, it is not a problem here. All of the changes involve merely changing the order of attributes within html tags; none of them affect how the generated html would appear in a browser.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


ccchapman:
Confirming this fixes my issue. Thank you for fixing so quickly.

oleibman:
@ccchapman Thank you for confirming. As you can see, I have made additional changes to fix pre-existing problems that I've encountered while doing more extensive testing. I think I am done with those now. It is very unlikely that the new changes will adversely affect you, but please test again if you can.

ccchapman:
@oleibman Thank you. I just tested [3cc82a9](https://github.com/PHPOffice/PhpSpreadsheet/pull/4541/commits/3cc82a975c02dbe8bc4d07c666d5d99ade310ed0) with a more complicated real-world spreadsheet. It works perfectly.

PR #4511, opened 2025-06-18
Implement ASC function and add unit tests

internauticos:
This commit introduces the implementation for the `ASC` spreadsheet function, which converts full-width (double-byte) characters to their corresponding half-width (single-byte) characters.

The implementation has been added to the `TextData` class. The main `Calculation` class has been updated to delegate calls to `ASC` to this new implementation.

Unit tests have been added to `TextDataTest.php` to cover various scenarios, including:
- Full-width alphanumeric characters.
- Full-width symbols.
- Mixed full-width and half-width characters.
- Empty strings.
- Boolean and numeric inputs.
- Japanese Katakana and Hiragana characters.

This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [ ] Changes are covered by unit tests
  - [ ] Changes are covered by existing unit tests
  - [ ] New unit tests have been added
- [ ] Code style is respected
- [ ] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ ] CHANGELOG.md contains a short summary of the change and a link to the pull request if applicable
- [ ] Documentation is updated as necessary

### Why this change is needed?

Provide an explanation of why this change is needed, with links to any Issues (if appropriate).
If this is a bugfix or a new feature, and there are no existing Issues, then please also create an issue that will make it easier to track progress with this PR.

PR #4544, opened 2025-07-19
Allow Replace of Dummy Function with Custom Function

oleibman:
Custom functions were introduced recently, with a restriction that non-custom functions could not be replaced. However, reviewing a recent issue has led me to the conclusion that it might sometimes be impractical to implement an Excel function in PhpSpreadsheet for all users, but it could still be helpful to some users to offer an implementation anyhow.

The Excel ASC function is currently unimplemented in PhpSpreadsheet, and is a candidate for replacement using the functionality added in this PR. This PR supersedes PR #4513, which I will now close, and its earlier incarnation PR #4511. For reasons discussed in 4513, I don't see a way forward for implementing the Excel ASC function in a way that would be generally usable. However, this PR would permit a user to implement ASC in a way which would satisfy the user's local requirements. See the new `testReplaceDummyFunction`; I think, or at least hope, that, despite the deficiencies of 4511 and 4513, this might satisfy the author's requirement.

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


oleibman:
Another day ending in 'y' means another false positive for Scrutinizer. It is now suppressed.

PR #4565, opened 2025-08-01
Create Uninitialized Cell If Used in Calculation

### LINKED ISSUE #4530: Problems with INDIRECT as Part of Range

### LINKED ISSUE #4558: COUNTA counts wrong if called with multiple arguments
``COUNTA(B1,B2,B3,B4)`` should count each cell which is not ``NULL``. In this case, however, the function is passed an argument

```
array(4) {
    [0] => NULL
    [1] => NULL
    [2] => NULL
    [3[ => NULL
}
```

and when examined, each index is treated as ``not a cell value`` by ``Functions::isCellValue()`` (because it lacks two dots) and thus counted by ``COUNTA()``. While Excel evaluates the expression above to 0 (if all cells are empty), PhpSpreadsheet (4.5.0) returns 4.

I did not check, but COUNT will most likely suffer from the same problem.

[test.xlsx](https://github.com/user-attachments/files/21470079/test.xlsx)

oleibman:
Fix #4558. Fix #4530. 4558 deals with a problem with COUNTA. Excel does not count empty cells, but it does count literal nulls (usually empty parameters) in the argument list. PhpSpreadsheet has till now not created empty cells during calculation, and so winds up treating them as literals. It is changed to create the cell when appropriate.

4530 has a similar problem. When INDIRECT winds up evaluating an uninitialized cell, it treats it as a null literal, leading to incorrect results. It is now changed to create the missing cell.

A number of functions, almost all of them Financial, have a related problem. They test for a null literal when initializing some of their parameters, but they should be testing for null literal or cell containing null. They are changed to do the right thing.

SUMIF and related functions are slightly affected by this change, and are changed so that they are no longer affected.

New tests have been added. Only one existing test had to change. MergeBehaviorTest for `[12, '=5+1', '=A1/A2']`, where cell A2 was uninitialized formerly returned one row, but now returns a second all-null row because A2, by virtue of being used in a calculation, now winds up defined. I do not consider this a significant difference.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ and this project adheres to [Semantic Versioning](https://semver.org).
 
 ### Fixed
 
-- Nothing yet.
+- Create uninitialized cell if used in calculation. [Issue #4558](https://github.com/PHPOffice/PhpSpreadsheet/issues/4558) [Issue #4530](https://github.com/PHPOffice/PhpSpreadsheet/issues/4530) [PR #4565](https://github.com/PHPOffice/PhpSpreadsheet/pull/4565)
 
 ## 2025-07-23 - 4.5.0
 
--- a/src/PhpSpreadsheet/Calculation/Calculation.php
+++ b/src/PhpSpreadsheet/Calculation/Calculation.php
@@ -1999,6 +1999,9 @@ class Calculation extends CalculationLocale
                             $this->debugLog->writeDebugLog('Evaluating Cell %s in worksheet %s', $cellRef, $matches[2]);
                             if ($pCellParent !== null && $this->spreadsheet !== null) {
                                 $cellSheet = $this->spreadsheet->getSheetByName($matches[2]);
+                                if ($cellSheet && !$cellSheet->cellExists($cellRef)) {
+                                    $cellSheet->setCellValue($cellRef, null);
+                                }
                                 if ($cellSheet && $cellSheet->cellExists($cellRef)) {
                                     $cellValue = $this->extractCellRange($cellRef, $this->spreadsheet->getSheetByName($matches[2]), false);
                                     $cell->attach($pCellParent);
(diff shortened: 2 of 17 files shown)


oleibman:
No concern about Scrutinizer's "complexity" message.

PR #4618, opened 2025-08-28
Option to Create Blank Sheet If LoadSheetsOnly Doesn't Find Any

oleibman:
In PhpSpreadsheet Release 1, if the LoadSheetsOnly option was specified, and no sheets matched, a new blank sheet was created. This behavior changed in PhpSpreadsheet Release 2, so that an exception wound up being thrown instead. Although the Release 2 approach seems more sensible to me, it was actually collateral damage from a different change, and was not an intentional result.

The difference in behavior is causing a problem for Laravel-Excel. In particular, a PR which would move their supported PhpSpreadsheet release from 1 to 5, is delayed because this change in behavior breaks part of their test suite. See https://github.com/SpartnerNL/Laravel-Excel/pull/4302. We would very much like them to get off release 1. I volunteered to add a compatibility option to the Readers which would emulate the release 1 behavior. The result is this PR.

Usage:
```php
$reader->setLoadSheetsOnly([list of sheet names]);
if (method_exists($reader, 'setCreateBlankSheetIfNoneRead')) {
    $reader->setCreateBlankSheetIfNoneRead(true);
}
```
In addition to Xlsx Reader, the method is available for Xls, Ods, Xml, and Gnumeric.

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

### DIFF
--- a/src/PhpSpreadsheet/Reader/BaseReader.php
+++ b/src/PhpSpreadsheet/Reader/BaseReader.php
@@ -55,6 +55,12 @@ abstract class BaseReader implements IReader
     protected bool $allowExternalImages = false;
 
     /**
+     * Create a blank sheet if none are read,
+     * possibly due to a typo when using LoadSheetsOnly.
+     */
+    protected bool $createBlankSheetIfNoneRead = false;
+
+    /**
      * IReadFilter instance.
      */
     protected IReadFilter $readFilter;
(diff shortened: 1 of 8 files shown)


patrickbrouwers:
Looks good @oleibman ! If it's possible, could you release this earlier than around the PHP8.5 release, with this addition I think I can complete the update to 5.0 in Laravel Excel. Thanks!

oleibman:
@patrickbrouwers Thanks for confirming. I will schedule a release for next week, tentatively Wed. Sep. 3.

Issue #3982, opened 2024-04-12
rangeToArray tries to read every possible cell leading to OOM


tommasoiovane:
This is:

- [X] a bug report

### What is the expected behavior?
I would expect `readActiveSheet` to return only cells with value or something in it.

### What is the current behavior?
With certain xlsx files rangeToArray calculate $maxCol as "AMK" and $maxRow as "1048579" leading to very slow performance and OOM memory issues.

### What are the steps to reproduce?
```php
<?php

require __DIR__ . '/vendor/autoload.php';

// Create new Spreadsheet object
$spreadsheet = \PhpOffice\PhpSpreadsheet\IOFactory::load("./test_file.xlsx");
$data = $spreadsheet->getActiveSheet()->toArray(null, true, false, true);
```
This code will fail running out of memory.  I managed to read the xls file changing Worksheet.php rangeToArray like this:
```php
$maxRow = $this->getHighestDataRow();
$maxCol = $this->getHighestDataColumn();
```
Is there any reason why it should'nt be like this?

### What features do you think are causing the issue

- [X] Reader

### Which versions of PhpSpreadsheet and PHP are affected?
I tried both on 1.3 and 2.0.0

Thank you in advance
[test_file.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/14960114/test_file.xlsx)

MarkBaker:
Not a bug, and nothing to do with the Reader. The Reader reads the spreadsheet file, and the values returned for a worksheet by `getHighestColumn()` and `getHighestRow()` are calculated based on values stored in the file, and cells that are referenced in the content of the file (perhaps by style information or range settings). `getHighestDataRow()` and `getHighestDataColumn()` are calculated at the point when they are called, and based purely on cells that contain data.
The `toArray()` methods use the `getHighestColumn()` and `getHighestRow()` values, changing it would be a BC break; but `rangeToArray()` can always be used with the `getHighestDataRow()` and `getHighestDataColumn()` values... that's why we provide the method.
We don't recommend the use of `toArray()` or `rangeToArray`, because creating a large PHP array will always demand a lot of memory. Instead, we'd recommend using the row and column iterators.

oleibman:
@tommasoiovane PR #3839 (part of release 2.0.0) and PR #3906 (part of release 2.1.0) address performance problems in toArray. I think they significantly help your issue. The code below is identical to the code you supplied, except for the addition of the echo statements.
```php
[...]

PR #4642, opened 2025-09-11
Export only data cells/columns in toArray

zoltan-vespertool:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Worksheet::toArray now calls these methods:

         $maxCol = $this->getHighestDataColumn();
         $maxRow = $this->getHighestDataRow();

instead of getHighestColumn and getHighestRow. This is to make sure that rows/columns that only have some formatting are not included. Otherwise the export would contain 16384 columns if e.g. column width setting is applied.

### DIFF
--- a/src/PhpSpreadsheet/Worksheet/Worksheet.php
+++ b/src/PhpSpreadsheet/Worksheet/Worksheet.php
@@ -3286,9 +3286,9 @@ class Worksheet
         $this->garbageCollect();
         $this->calculateArrays($calculateFormulas);
 
-        //    Identify the range that we need to extract from the worksheet
-        $maxCol = $this->getHighestColumn();
-        $maxRow = $this->getHighestRow();
+        // Identify the range that we need to extract from the worksheet
+        $maxCol = $this->getHighestDataColumn();
+        $maxRow = $this->getHighestDataRow();
 
         // Return
         return $this->rangeToArray("A1:{$maxCol}{$maxRow}", $nullValue, $calculateFormulas, $formatData, $returnCellRef, $ignoreHidden, $reduceArrays);
--- a/tests/PhpSpreadsheetTests/Reader/Xlsx/XlsxTest.php
+++ b/tests/PhpSpreadsheetTests/Reader/Xlsx/XlsxTest.php
@@ -11,7 +11,6 @@ use PhpOffice\PhpSpreadsheet\Reader\Xlsx;
 use PhpOffice\PhpSpreadsheet\Shared\File;
 use PhpOffice\PhpSpreadsheet\Shared\StringHelper;
 use PhpOffice\PhpSpreadsheet\Style\Conditional;
-use PhpOffice\PhpSpreadsheet\Style\Style;
 use PhpOffice\PhpSpreadsheet\Worksheet\AutoFilter;
 use PhpOffice\PhpSpreadsheet\Worksheet\PageSetup;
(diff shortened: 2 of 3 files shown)


oleibman:
I assume you can figure out the phpcs failure on your own. Let me know if you need assistance for that.

The failing test is peculiar. The values on the spreadsheet are integers, but the test is comparing to floats. I don't see why your change should have affected anything, but the test is less than optimal anyhow. I would have no objection to your making the following changes:
```php
        $data = $spreadsheet->getActiveSheet()->toArray(formatData: false);
        $ref = [1, null, 3, null, 5, null, 7, null, 9, null];

        for ($i = 0; $i < 10; ++$i) {
            self::assertSame($ref, \array_slice($data[$i], 0, 10, true));
        }
```
That being said, there has been a lot of discussion in various issues about get... vs getHighest... I need to study those to make sure that the current code does not use getHighest for a reason

zoltan-vespertool:
Fixed the formatting.

For the test: the test failing was actually related to my change. But it is not about the float/int conversion - in fact, toArray() returns all cells as `string|null`, so the actual values returned are `'1'`, `'3'` etc. thus neither `1` as int nor `1.0` as float.

What happens there, is the rightmost column (containing '10') is not loaded because of the filter `OddColumnReadFilter`. Then all those cells (column J) become empty data-wise. On the other hand, the cells have formatting: libreoffice shows me those cells as "font=Calibri", while the rest of the cells as "font=-apple-system".

The other way to fix the test would be to use `rangeToArray` instead of `toArray`.

zoltan-vespertool:
I was thinking about this a bit more, and fixed the test in a different way.

So the scope of this test is to see if read filters work. But given that the example XLSX file used in the test has cell formatting beyond data cells (the latter are the 10x10 matrix A1:J10), it was affected by the behavior of `toArray()` considering formatting and not only data. That was what it was working around with the `array_slice()` calls.

Thus the fix is to use `rangeToArray()` in this test, to make it independent of how `toArray()` works. It then allows for the removal of the `array_slice()` call. The string/float conversions don't matter in any way, as `assertEq()` would use `==` internally, causing PHP to use `==` for the array elements.

oleibman:
Your work on this is exemplary. Unfortunately, as I mentioned was a possibility, I will not be able to merge this change in its present form. See https://github.com/PHPOffice/PhpSpreadsheet/issues/3982#issuecomment-2052201440. The fact that your change broke an unrelated test confirms that it is indeed a breaking change.

Although I cannot merge this change, the idea behind it is worthwhile. I imagine that most users would want your behavior rather than how PhpSpreadsheet handles the situation now. But a breaking change is a breaking change. It is not something we take lightly, especially since PhpSpreadsheet already provides a means to get the result you want:
```php
rangeToArray('A1:' . $sheet->getHighestDataColumn() . $sheet->getHighestDataRow);
```

However, I would not be averse to adding an extra parameter to `toArray` which tells it whether to use `getHighestDataColumn/Row` rather than `getHighestColumn/Row`. It would have to default to the current behavior, but we could change the default for the next breaking release. If you want to submit a PR which does that, I would be willing to move forward with it.

Thank you again for your work on this change. I am sorry that it has so far led to a disappointing outcome.

oleibman:
Closing. Leaving the original issue open - a PR which uses a different approach than this one may be acceptable.

Issue #4014, opened 2024-05-02
PhpSpreadsheet can't retrieve images placed inside cell


AntonioA-D-C:
```
  
    $spreadsheet=  IOFactory::load($this->file);
    $i= 0;
    dd($spreadsheet->getActiveSheet($row)->getDrawingCollection());
    foreach ($spreadsheet->getActiveSheet()->getDrawingCollection() as $drawing) {
      if ($drawing instanceof MemoryDrawing) {
          ob_start();
          call_user_func(
              $drawing->getRenderingFunction(),
              $drawing->getImageResource()
          );
          $imageContents = ob_get_contents();
          ob_end_clean();
          switch ($drawing->getMimeType()) {
              case MemoryDrawing::MIMETYPE_PNG :
                  $extension = 'png';
                  break;
              case MemoryDrawing::MIMETYPE_GIF:
                  $extension = 'gif';
                  break;
              case MemoryDrawing::MIMETYPE_JPEG :
                  $extension = 'jpg';
                  break;
          }
      } else {
          $zipReader = fopen($drawing->getPath(), 'r');
          $imageContents = '';
          while (!feof($zipReader)) {
              $imageContents .= fread($zipReader, 1024);
          }
          fclose($zipReader);
          $extension = $drawing->getExtension();
      }
    
      $myFileName = time() .++$i. '.' . $extension;
      dd($myFileName);
    }
```

Here my script won't trigger when I have images that are inside cells, it ignores images inside the cells for images outside the cells, how to fix this or how to retrieve images contained within cells

oleibman:
Confirming what you've already surmised. PhpSpreadsheet has no way to set a cell value to an image. As for retrieving such an image, I've looked at a spreadsheet which does this, and I really don't see how this is handled in the xml. Even if I do figure it out, it would probably be a pretty major change to extend the set of datatypes to which a cell value can be set.

oleibman:
Closing as duplicate of #4034.

vidya-balakrishnan:
> spreadsheet

Which spreadsheet library supports embedding an image directly into a cell as its value?

raziel057:
@oleibman I had a look on how it works and I can see the following. Considering I added 3 images. I inserted a specific image in cell H3 and the same image in cell K3 and L3, in the XML I have:

<img width="1110" height="102" alt="Image" src="https://github.com/user-attachments/assets/a4d67931-fca0-414d-9306-27e6eafa5353" />

[...]

PR #4664, opened 2025-09-26
Support for read of Image in Cell

raziel057:
Support for read of Image in Cell - Related to https://github.com/PHPOffice/PhpSpreadsheet/issues/4014

This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [x] additional unit tests

Checklist:

- [x] Code style is respected

When an image is inserted in a cell in an XLSX document, PhpSpreadsheet is currently not able to read it. This PR allow to collect those images in order to find them when calling $objWorksheet->getDrawingCollection()

The code have been tested - As shown bellow:
image

When I import the file with image inserted in the Cell, I have it now in $objWorksheet->getDrawingCollection():

image

As an improvement, maybe we should introduce a DataType::TYPE_DRAWING and adapt the code accordingly to allow the user to get the Drawing object when calling getValue() on the cell?


oleibman:
You will need to fix the phpcs, php-cs-fixer, and phpstan errors. After those, the unit test errors. After that, you need some formal unit tests. Finally, you've addressed only the read side. It isn't absolutely required, but I would be happier with a change which addressed both reading and writing (which I will admit is probably much harder).

I do not know if TYPE_DRAWING is a good idea or not. I'll think about it when this PR is in better shape.

oleibman:
It looks like you have only one test failure, but it was caught during multiple tests. When a formula has an array result which spills into cells which already have values, Excel also uses the `vm` attribute for the cell. I think it is probably okay for you test for a `vm` attribute *and* no child `f` tag when deciding if it's an image. There may be more to it, but that will at least get you over one hurdle for now.

raziel057:
I reworked the implementation as I could see that images are defined as Relationship in ``richData\_rels\richValueRel.xml.rels``.

image

I didn't had the time to work on the Write part, but I think it could be done in another PR.

raziel057:
Not sure why the coverage decreased (-0.04%) as I created a Unit Test that cover all the code that have been added. I'm not able to see the details https://coveralls.io/builds/75720737/source?filename=src%2FPhpSpreadsheet%2FReader%2FXlsx.php#L791

oleibman:
Thank you for adding the tests. Decreasing code coverage is not necessarily a reason for us to proceed with a change. However, in this case, it is a showstopper. The Coveralls report shows that most of the code you added is not being exercised by the unit test suite. In particular:
- You added lines 787-799 in Reader/Xlsx, but lines 791-795 are never executed.
- You added lines 956-977, but lines 957-975 are never executed.

So, either the bulk of your new code is unneeded, or it is inadequately tested. Please see what you can do about that. I have some ideas about how to proceed once you've attended to this.

raziel057:
Ok I forgot the Test suffix in ``DrawingInCellTest``, so it was not taken when running the test suite. It should be fine now.

raziel057:
I think the issue comes from the management of path that can be different in CI. I will have a look on it.

kboumedal:
Thank you for this PR !
Any update on merging it ?

Issue #4668, opened 2025-10-01
Undefined array key "x" PHP warning


karrakoliko:
```
N tests triggered 1 PHP warning:

1) *project_root*/vendor/phpoffice/phpspreadsheet/src/PhpSpreadsheet/Reader/Xml/Style.php:71
Undefined array key "x"
```

This php warning fails tests, and does not affect anything related to reading xls file.

Seems like there is an issue with the xls file itself, but "x" key is [hardcoded here](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Reader/Xml/Style.php#L79):

```php
                    case 'Protection':
                        $locked = $hidden = null;
                        $styleAttributesP = $styleData->attributes($namespaces['x']); // HERE IT IS
                        if (isset($styleAttributes['Protected'])) {
                            $locked = ((bool) (string) $styleAttributes['Protected']) ? Protection::PROTECTION_PROTECTED : Protection::PROTECTION_UNPROTECTED;
                        }
                        if (isset($styleAttributesP['HideFormula'])) {
                            $hidden = ((bool) (string) $styleAttributesP['HideFormula']) ? Protection::PROTECTION_PROTECTED : Protection::PROTECTION_UNPROTECTED;
                        }
```

isset/array_key_exists check is missing, so php warning is triggered.

my suggestion is to add `array_key_exists` check to avoid reading unexisting keys from unexisting array:
```
$styleAttributesP = array_key_exists('x',$namespaces) ? $styleData->attributes($namespaces['x']) : [];
```

PR #4669, opened 2025-10-01
fix: do not trigger `Undefined array key "x"`

karrakoliko:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

fixes https://github.com/PHPOffice/PhpSpreadsheet/issues/4668

### DIFF
--- a/src/PhpSpreadsheet/Reader/Xml/Style.php
+++ b/src/PhpSpreadsheet/Reader/Xml/Style.php
@@ -76,7 +76,7 @@ class Style
                         break;
                     case 'Protection':
                         $locked = $hidden = null;
-                        $styleAttributesP = $styleData->attributes($namespaces['x']);
+                        $styleAttributesP = array_key_exists('x', $namespaces) ? $styleData->attributes($namespaces['x']) : [];
                         if (isset($styleAttributes['Protected'])) {
                             $locked = ((bool) (string) $styleAttributes['Protected']) ? Protection::PROTECTION_PROTECTED : Protection::PROTECTION_UNPROTECTED;
                         }
--- a/tests/PhpSpreadsheetTests/Reader/Xls/XlsTest.php
+++ b/tests/PhpSpreadsheetTests/Reader/Xls/XlsTest.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace PhpOffice\PhpSpreadsheetTests\Reader\Xls;
 
 use PhpOffice\PhpSpreadsheet\Cell\Cell;
+use PhpOffice\PhpSpreadsheet\IOFactory;
 use PhpOffice\PhpSpreadsheet\Reader\Xls;
 use PhpOffice\PhpSpreadsheetTests\Functional\AbstractFunctional;
 
@@ -92,7 +93,7 @@ class XlsTest extends AbstractFunctional
     public function testLoadXlsBug1114(): void
     {
         $filename = 'tests/data/Reader/XLS/bug1114.xls';
-        $spreadsheet = \PhpOffice\PhpSpreadsheet\IOFactory::load($filename);
+        $spreadsheet = IOFactory::load($filename);
         $sheet = $spreadsheet->getActiveSheet();
         self::assertSame(1148140800.0, $sheet->getCell('B2')->getValue());
         $spreadsheet->disconnectWorksheets();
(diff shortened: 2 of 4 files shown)


oleibman:
Can you add a unit test which will fail without your change but succeed with it?

karrakoliko:
@oleibman done

oleibman:
Thank you for adding the test. However ... the code in question is for an Xml file, not Xls. Please rename your file accordingly and move it to tests/data/Reader/Xml, and move your test to tests/PhpSpreadsheetTests/Reader/Xml.

It's a bit of a surprise that this is a problem, because the Xml file does define a namespace 'x'. However, the `getNameSpaces` function returns the namespaces *used* in the document, not the namespaces defined in it. It would not surprise me if there are more surprises of this nature, but we needn't delay your change on that account.

karrakoliko:
@oleibman done

> getNameSpaces function returns the namespaces used in the document, not the namespaces defined in it

i can easily beleive that the file itself is semi-valid. i removed all the xml rows from it because they contain highly sensitive data.

probably it would be a better idea to anonymize data instead (`John Doe`, card number `111111***2222` etc), so it could potentitally be helpful for understanding the context (?).

in this specific case it is just about hardcoded array key without proper check, so i'm not ready to spend time on anonymization.

karrakoliko:
@oleibman done

oleibman:
Thank you. I am sorry to be late with this suggestion, but here goes anyhow. Your test does its job, but it's more complicated than it needs to be. When running unit tests, our default error handler will already convert any warning messages to exceptions. So you don't need to set up your own error handler. You will need to set up a test after you load the spreadsheet, but that should be easy, something like:
```php
$spreadsheet = $reader->load($filename);
self::assertSame('Report Date', $sheet->getCell('A1')->getValue());
```
That test will fail without your change (its warning will be converted to an exception), but will succeed with your change, which is the result we're looking for.

karrakoliko:
> When running unit tests, our default error handler will already convert any warning messages to exceptions

it can be changed any day by any reason.

test should assert that php warning was not triggered.

your test does not: it just reads specific cell value and asserts that it matches expected one, that will be passed even if php warning occurred.

so:
- expected assertion: no php warning triggered
- your test assertion: lib correctly reads value from the cell

if error handler config will be changed - test will be false positive. we dont want potentially false positive tests.

relying on phpunit runtime config parameter instead of clear assertion is a bad idea.

oleibman:
We have tests to make sure that we treat all warnings, deprecations, and notices as exceptions. That will not be changing. We could not survive new Php releases otherwise. We will not be changing that. On the other hand, Php changes its messages all the time. "Undefined array key 'x'" is by no means guaranteed to be constant.

karrakoliko:
> We have tests to make sure that we treat all warnings, deprecations, and notices as exceptions

ah, this changes a lot.

i'll change test body to yours one and add phpdoc comment about what is actually tested here.

is it ok for you?

oleibman:
Yes, that would be great. I will merge as soon as I see the change. Thank you for your flexibility.

oleibman:
@karrakoliko  This discussion has been helpful. Although this project does have tests that warnings etc. are converted to exceptions, another project that I work on did not. I will add the necessary tests there.

karrakoliko:
@oleibman done.  
Just to confirm: once this is merged into master, will the patch be included in all versions (3.x, 4.x, 5.x, etc.)?  

We’re currently using `^3.5` in production. To suppress that warning we had to add a rather ugly workaround, which we’d like to remove.

oleibman:
@karrakoliko we don't backport most changes - we just don't have the time. However, I will make an exception in this case - it is a small change, it is unlikely to depend on changes applied to the master branch since it split from the branch that you are using, and the developer (you) wants it. However, I will apply it only to the latest release on your branch, which will be 3.10.2 when I apply it. My plan will be to tag that release just before or just after Php8.5 goes live. If you don't want to try PhpSpreadsheet 5, you should at least consider upgrading to 3.10.1 in the meantime.

karrakoliko:
Thank you ♥️ @oleibman

oleibman:
Thank you for your contribution, and for your flexibility. I have also merged this into the realease390 branch.

PR #4690, opened 2025-10-21
Fix phpstan issues by making Escher generic

BackEndTea:
This is:

- [X] refactoring

Checklist:

- [X] Code style is respected

Extra checks are currently needed on the result of the `Escher::load` method, as its not clear what the return type becomes.

By making this class generic, we allow to infer the return type from the constructor argument, for static analysis tools.

Additional `get...OrThrow` methods have been added, to throw more precise errors, rather than Fail due to a method call on null.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -7,30 +7,6 @@ parameters:
 			path: src/PhpSpreadsheet/Calculation/LookupRef/Sort.php
 
 		-
-			message: '#^Cannot call method getAllSpContainers\(\) on mixed\.$#'
-			identifier: method.nonObject
-			count: 1
-			path: src/PhpSpreadsheet/Reader/Xls/LoadSpreadsheet.php
-
-		-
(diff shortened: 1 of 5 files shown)


oleibman:
Thank you for an interesting improvement. Please make it a little better by avoiding new uncovered code. Instead of:
```php
    public function getDggContainerOrThrow(): Escher\DggContainer
    {
        if ($this->dggContainer !== null) {
            return $this->dggContainer;
        }

        throw new SpreadsheetException('dggContainer is unexpectedly null');
    }
```
Please try something like:
```php
    public function getDggContainerOrThrow(): Escher\DggContainer
    {
            return $this->dggContainer ?? throw new SpreadsheetException('dggContainer is unexpectedly null');
    }
```

BackEndTea:
I considered that, but wanted to keep it in line with other `get...OrThrow` methods. I'll update these methods to the newer syntax

oleibman:
Thank you for making the change. You are correct that other routines could be similarly re-coded. It's all a question of priorities.

oleibman:

Issue #4037, opened 2024-05-22
Textboxes, shape are not visible when using mergefields in file that contains image (Excel)


MarekPolok:
This is:

```
- [x] a bug report
```

### What is the expected behavior?
Textboxes, Shapes etc should be visible after using merging of fields

### What is the current behavior?
Textboxes, Shapes are gone when there is a image in excel.
When Image is removed everything works as expected - shapes, textboxes are not removed

**Screenshots showing what is the problem:**
Orginal file before merging fields:
![Screenshot 2024-05-22 at 10 08 19](https://github.com/PHPOffice/PhpSpreadsheet/assets/23061427/ad17507b-92cd-4a59-b86f-ac84cc1bfd7a)
File after merging fields:
![Screenshot 2024-05-22 at 10 09 33](https://github.com/PHPOffice/PhpSpreadsheet/assets/23061427/a5d5d2c9-ac39-4000-a15a-070bd491194d)

**Sample file which is failing:**
[merge excel.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/15400250/merge.excel.xlsx)

### What are the steps to reproduce?

We are taking a file and using regex to match merge field inserted in excel and fill it with data from our system.
So we are taking file, process fields and save the document

```php
<?php

public static function Load(File $file): static {
	$fileFullPath = $file->GetFullPath();

	$reader = IOFactory::createReaderForFile($fileFullPath);
	$document = $reader->load($fileFullPath);

	$res = new static;
	$res->spreadsheet = $document;

	return $res;
}

public function MergeToFile(File $sourceFile, IFile $file): void {
	$excelDoc = ExcelDocument::Load($sourceFile);
	$excelDoc->CellSearchReplace(new Regex('/\[\[([\w\/_]+)]]/'), function(string $value, array $matches) {
		$key = $matches[1];
		return $this->fieldValues[$key] ?? $this->getFieldValueFromFunction($key) ?? '';
	});
	$excelDoc->SaveToFile($file);
}
```

### Which versions of PhpSpreadsheet and PHP are affected?
PhpSpreadsheet version: 2.1.0
PHP version: 8.3.2

oleibman:
Limited support for form controls was added by PR #3130. The keyword in this case is "limited". We can read them and write them (at least some of the time), but can't actually do anything in the code with them. I agree that we aren't reading your file correctly. I will see if we can do anything about that.

xiongmao-p:
[...]

Issue #4704, opened 2025-11-06
Grouped images are not preserved when reading and writing xlsx files


rpg600:
### Environment
PhpSpreadsheet version: 5.2.0 (also affects earlier versions)
PHP version: 8.x
File format: .xlsx (Office Open XML)

### Description
When loading an Excel file containing grouped images and saving it back, the images inside the group are lost.

### Screenshot
<img width="778" height="807" alt="Image" src="https://github.com/user-attachments/assets/8bbaa9ca-3792-4063-8d1a-1f047e103190" />

### test file
[group_issue.xlsx](https://github.com/user-attachments/files/23388765/group_issue.xlsx)

### Xml file comparison

[original_drawing.xml](https://github.com/user-attachments/files/23389654/original_drawing.xml)

[generated_drawing.xml](https://github.com/user-attachments/files/23389659/generated_drawing.xml)

### Xml debug
```
works : 
<xdr:twoCellAnchor>
  <xdr:pic>
    <a:blip r:embed="rId1"/>
  </xdr:pic>
</xdr:twoCellAnchor>

don't work :
<xdr:twoCellAnchor>
  <xdr:grpSp>              ← THE PROBLEM
    <xdr:pic>
      <a:blip r:embed="rId2"/>
    </xdr:pic>
    <xdr:pic>
      <a:blip r:embed="rId3"/>
    </xdr:pic>
  </xdr:grpSp>
</xdr:twoCellAnchor>
```

### What happens
```
1) Reader/Xlsx.php (~line 1489)
Parses <xdr:pic> elements directly under <xdr:twoCellAnchor>
Does NOT parse <xdr:pic> inside <xdr:grpSp>
Grouped images are never loaded into DrawingCollection

2) Writer/Xlsx/Drawing.php
Regenerates Drawing XML from DrawingCollection only
Grouped images are not in the collection
Result: Grouped images omitted from output
```

### Workaround:
```
Manually ungroup all images before processing (not always possible)
```

### Solutions :
```
Option 1: Full Group Support (complex)
Option 2: Append the unsupported xml elements (pass-through)
```

oleibman:
Agreed - we do not yet handle grouped images. It looks like it could be very difficult to add them. Would it be useful to you if the reader processed the grouped images as individual images, so that you would lose the groupings but not the images? This would still be difficult to implement, but not as difficult as full support.

rpg600:
@oleibman Yes, as long as the image position stays the same.

[...]

PR #4677, opened 2025-10-03
Support Drawing in cell Read & Write

raziel057:
Support for read & write of Image in Cell - Related to https://github.com/PHPOffice/PhpSpreadsheet/issues/4014

I must admit it was really complex to understand the relations between all xml (richData) files that have to be created / updated. I lost a lot of time comparing the structure with files edited in Excel and specifically figuring out the ``metadata.xml`` file had to be largely changed as soon as images are added in cells.

This branch is based on https://github.com/PHPOffice/PhpSpreadsheet/pull/4664

This is:

- [x] a new feature
- [x] additional unit tests

Checklist:

- [x] Changes are covered by unit tests
  - [x] New unit tests have been added
- [x] Code style is respected
- [ ] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [ ] CHANGELOG.md contains a short summary of the change and a link to the pull request if applicable
- [ ] Documentation is updated as necessary

Support for read and write of Image in Cell - Related to https://github.com/PHPOffice/PhpSpreadsheet/issues/4014

Why this change is needed?

When an image is inserted in a cell in an XLSX document, PhpSpreadsheet is currently not able to read it. This PR allow to collect those images in order to find them when calling ``$objWorksheet->getInCellDrawingCollection()``. It's also possible to get the Drawing object from from the related Cell.

The code is covered by tests and have been tested in real application - As shown bellow:
image

When I import the file with image inserted in the Cell, I have it now in $objWorksheet->getInCellDrawingCollection():

image

PR #4712, opened 2025-11-13
feat: add drawing pass-through support for unsupported elements

rpg600:
feat: add drawing pass-through support for unsupported elements

Add opt-in pass-through mechanism to preserve unsupported drawing 
elements (shapes, textboxes etc) during read/write operations.

- Add enableDrawingPassThrough property to BaseReader
- Store original drawing XML when pass-through is enabled
- Reuse stored XML in Writer instead of regenerating
- Disabled by default for backward compatibility

This allows preservation of Excel elements not yet supported by 
PhpSpreadsheet without requiring full implementation.

Currently, the pass-through mechanism works as follows:
1. **Reader**: `setEnableDrawingPassThrough(true)` stores the original drawing XML
2. **Writer**: Automatically uses the stored XML if:
   - The drawing collection is empty (no drawings added programmatically)
   - AND the unparsed data contains the original XML
   
I'm open to adding an explicit Writer flag if you believe it provides better clarity and control.

This addresses existing issue : https://github.com/PHPOffice/PhpSpreadsheet/issues/4037


oleibman:
I am still evaluating whether this is a good idea or not. Will it work if you add or remove a row or a column on the sheet? Add a comment when there weren't any before, and when there were? Remove a comment?

Your commentary says that you are fixing 3 open issues. Are there tests for each of those?

rpg600:
> Will it work if you add or remove a row or a column on the sheet?

Yes but the drawing's coordinates (col, row, offsets) will not be adjusted.

> Add a comment when there weren't any before, and when there were? Remove a comment?

Comments are completely independent from drawings.

> Your commentary says that you are fixing 3 open issues. Are there tests for each of those?

I was too quick when referencing the issues, this PR actually only addresses one of them.

I added several unit tests, including checks for unsupported elements (shapes/textboxes), behavior without pass-through, comment deletion, row/column operations, and correct flag handling between the Reader and Writer.

This feature isn’t ideal, as it’s essentially a workaround, but it will provide a temporary solution while we wait for those implementations, and it will help avoid patching the code.

oleibman:
Thank you for adding the tests.

Comments are, alas, not completely independent from drawings. It's why I asked the question. Create a new spreadsheet with a comment. Notice that there is a drawings folder with a vml file in it in the saved spreadsheet. I can't offhand tell whether this will be a problem for your change.

rpg600:
I have added **testCommentsAndPassThroughCoexist**
It shows that VML drawings and DrawingML coexist in the xl/drawings/ folder without interference. Both existing and new comments are preserved correctly when pass-through is enabled.

oleibman:
It is not absolutely required that Coveralls doesn't report a decrease. However, we strive to eliminate that situation if possible, especially in code which is new in a change. In Writer\Xlsx\Drawing.php, you have the following new statements which include some "misses" around line 617:
```php
        if (!isset($unparsedLoadedData['sheets'][$codeName])) {
            return null;
        }

        $sheetData = $unparsedLoadedData['sheets'][$codeName];
        if (!is_array($sheetData)) {
            return null;
        }

        // Only use pass-through XML if the Reader flag was explicitly enabled
        if (($sheetData['drawingPassThroughEnabled'] ?? false) !== true) {
            return null;
        }

        if (!isset($sheetData['Drawings']) || !is_array($sheetData['Drawings'])) {
            return null;
        }
``` 
I tthink if you changed that block as follows, your code would still work as planned and there would no longer be any coverage misses:
```php
        $sheetData = $unparsedLoadedData['sheets'][$codeName] ?? null;
        // Only use pass-through XML if the Reader flag was explicitly enabled
        if (!is_array($sheetData) || ($sheetData['drawingPassThroughEnabled'] ?? false) !== true || !is_array($sheetData['Drawings'] ?? null)) {
            return null;
        }
```
Please give that a shot.

oleibman:
Thank you - success on both fronts - no problems with unit tests, and no missed statements.

rpg600:
Hi @oleibman,

Thanks for your feedback.
I was wondering if you’re still interested in merging this feature. If so, do you have a timeline in mind?

oleibman:
I am probably interested in merging this. I do not have a timeline in mind, but I imagine it could happen before year-end. It is complicated, and I need to understand it better than I do at the moment, because once I merge it, I have to maintain it.

oleibman:
In issue #4037, there is a spreadsheet attached:
https://github.com/PHPOffice/PhpSpreadsheet/files/15400250/merge.excel.xlsx
It has a photograph, a textbox, a drawing of glasses, and a blue shape.

When I load and save that with the existing PhpSpreadsheet code, and open the result in Excel, I see the photograph and the glasses, but the textbox and shape are missing. The missing parts are, of course, why you are writing the PR.

When I load and save that with your new code (making sure to `setEnableDrawingPassThrough(true)`, I now see the textbox and the shape, as well as the glasses. But the photograph - Excel says the picture cannot be displayed! Do you see a different result when you try it?

rpg600:
@oleibman You're right — I observed the same behavior. 

The issue is with SVG images (the glasses in [merge.excel.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/15400250/merge.excel.xlsx)): Excel stores both the SVG and a PNG fallback with separate rIds (e.g., rId1=PNG fallback, rId2=SVG). Since PhpSpreadsheet doesn't load SVG images into the drawing collection, only the PNG fallback is written to the relations file. This causes the rIds to shift and no longer match the original pass-through drawing XML.

For example:
- Original: rId1=image1.png, rId2=image2.svg, rId3=image3.jpeg
- After save: rId1=image1.png, rId2=image3.jpeg (shifted)
- But the XML still references rId3 for image3.jpeg → broken

One solution would be to also pass through the original drawing relationships and media files. WDYT ?

oleibman:
Another solution might be to pass through the SVG file as a drawing, in the same manner as you are passing through other files. This is just a thought off the top of my head. I have no idea how practical or difficult that would be, especially compared to your proposal.

rpg600:
Thank you for your suggestion. I analyzed this approach, but it turns out SVGs in Excel are not standalone drawings, they're embedded as extensions within `<a:blip>` elements using `<asvg:svgBlip>`.

It looks like this:
```
<a:blip r:embed="rId1">
  <a:extLst>
    <a:ext uri="{...}">
      <asvg:svgBlip r:embed="rId2"/>  <!-- SVG reference inside PNG blip -->
    </a:ext>
  </a:extLst>
</a:blip>
 ```

The SVG and its PNG fallback are bound together in one drawing element. Supporting separate SVG drawings would require a major refactoring IMO.

The simpler solution is to extend pass-through to also preserve:
  1. The original drawing relationships (to keep rId alignment)
  2. The original media files (to copy SVGs from source)

 This approach:
  - Keeps changes minimal
  - Doesn't break existing drawing logic
  - Works for any unsupported media format

I've implemented this fix with tests using [merge.excel.xlsx](https://github.com/PHPOffice/PhpSpreadsheet/files/15400250/merge.excel.xlsx)

oleibman:
You are getting very close. Still, there is a problem. When I load and save merge.excel.xlsx using your PR and `setEnableDrawingPassThrough`, Excel complains when I try to open the result. If I let it do whatever it thinks is needed to clean up, the result is as desired - the photo, the glasses, the textbox, and the shape are all there. If you can figure out and avoid whatever it is Excel is objecting to, I think I will be able to merge this.

rpg600:
@oleibman Thanks for testing! I think I’ve found the issue : the svg content type was missing. 
I don’t have Excel available to test (only the online version), so could you please check if the output file now opens without warnings?
Also, I forgot to mention that grouped images are now preserved as well.

oleibman:
Excellent work! Excel now opens the copied file without a problem.

Issue #4704 deals with grouped images. Your adding them is definitely a bonus. I would like to give the author of that issue a chance to chime in before merging this. But it should happen soon.

rpg600:
Very soon! I'm actually the author of issue #4704 as well hehe.

oleibman:
My apologies for not noticing that.

oleibman:

rpg600:
@oleibman Happy New Year! Do you have an idea of when you’re planning the next release?

oleibman:
I want in-cell images (PR #4677) to be part of a new release. I think it might be ready to merge within a few days, and I'll schedule the new release for about a week after that.

PR #4716, opened 2025-11-19
Add default value to data type parameter of `Worksheet::setCellValueExplicit`, for consistency with `Cell::setValueExplicit`

RobinvanderVliet:
This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

The method `Cell::setValueExplicit` has a default value for the data type parameter, while the method `Worksheet::setCellValueExplicit` does not have this. This pull request gives both methods the same default.

### DIFF
--- a/tests/PhpSpreadsheetTests/Worksheet/WorksheetTest.php
+++ b/tests/PhpSpreadsheetTests/Worksheet/WorksheetTest.php
@@ -669,4 +669,32 @@ class WorksheetTest extends TestCase
             ],
         ];
     }
+
+    public static function setCellValueExplicitTypeArgumentProvider(): array
+    {
+        return require 'tests/data/Cell/SetValueExplicitTypeArguments.php';
+    }
+
+    #[DataProvider('setCellValueExplicitTypeArgumentProvider')]
+    public function testSetCellValueExplicitTypeArgumentHandling(
+        mixed $value,
+        ?string $dataType,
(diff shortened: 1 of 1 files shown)


oleibman:
Another one that I'll have to think about. You are absolutely correct - they _should_ be consistent. But the whole point of `setCellValueExplicit` is to supply the "explicit" part. I'm surprised that it allows a default. We can't change the Cell implementation for now, because that would be a breaking change, but I would be happier if we deprecated using a default there (though I'm not sure how to do that in PhpDocumentor so maybe it's just a pipe dream). At any rate, you should add a unit test while I'm pondering my options.

RobinvanderVliet:
I have added unit tests for both the cell and the worksheet methods. They both use the same data provider to ensure that both methods work in the same way. `DataType::TYPE_STRING` seems like a good default option for now.

oleibman:
I am inclined to not proceed with this change for reasons discussed above. I have opened PR #4717 to solve the issue you opened the way I think it ought to be. I am going to leave this one open for a few days to give you the opportunity to persuade me that this is a better approach.

Thank you for your effort on this PR. Even if it isn't merged, it hasn't all been in vain. As you can see, I have borrowed your unit tests for my PR.

oleibman:
Dropped in favor of PR #4717.

PR #4736, opened 2025-12-04
Removed duplicate try-catch statements

zhukoff74:
This is:

- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

- Instead of `try-catch`, built-in PHPUnit methods are used: `expectException()` and `expectExceptionMessageMatches()`
- Instead of `Throwable`, `Exception` is used, since this is the type that is thrown in `bootstrap.php`
- The code became shorter and clearer

### DIFF
--- a/tests/PhpSpreadsheetTests/Helper/HandlerTest.php
+++ b/tests/PhpSpreadsheetTests/Helper/HandlerTest.php
@@ -4,9 +4,9 @@ declare(strict_types=1);
 
 namespace PhpOffice\PhpSpreadsheetTests\Helper;
 
+use Exception;
 use PhpOffice\PhpSpreadsheet\Helper\Handler;
 use PHPUnit\Framework\TestCase;
-use Throwable;
 
 class HandlerTest extends TestCase
 {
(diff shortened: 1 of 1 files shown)


oleibman:
FYI, I might have used `expectExceptionMessage`, which asserts that the exception message *contains* (rather than matches) the string, rather than `expectExceptionMessageMatches` with its regexp overhead (which isn't needed in any of your tests). I don't feel strongly enough about it to ask you to change - just something to think about it in future.

oleibman:

hacfi:
@zhukoff74 @oleibman What is the Handler class good for? I just stumbled upon it because `date_default_timezone_set` in `PhpOffice\PhpSpreadsheet\Helper\Handler::notice`. I don't think anyone should use it  and it's exclusively used in its test file in `tests/PhpSpreadsheetTests/Helper/HandlerTest.php`. Should we remove it?

oleibman:
You ask a reasonable question. The answer is that we should not remove it. Explanation follows.

Phpunit used to have the ability to turn notices, warnings, and deprecation messages into exceptions. We liked that - we want to make sure our code isn't issuing any superfluous messages even when it executes correctly. Then they took that away without any real guidance on how to proceed. So we added our own handler to our test suite to do what Phpunit used to do for us. The Handler class is designed to execute code which generates warnings, etc.. We expect these to be captured by our test suite handler. It is especially helpful for handling deprecation messages when Php introduces a new release.

So why, you may ask, isn't the Handler class in the test suite rather than the main source code? It's because we want to make sure that it's properly tested. If we were to inadvertently take away the tests that use it, that would show up as a decrease in code coverage, so we would at least be aware that something unexpected had happened.

I could probably mark the class as "internal", an option I was unaware of at the time. That way, it won't be in the public documentation. People would still be able to stumble across it as you did, but perhaps that would be less likely - but perhaps not, I don't think that's how you found it. Nevertheless, I will think about doing that.

oleibman:
PR #4895

Issue #3961, opened 2024-03-26
Unable to save with setFormatCode in ODS, and getFormatCode from existing documents


MinxAa:
This is:

```

[X] a bug report

```

### What is the expected behavior?

Formatted numbers.

### What is the current behavior?

No formatting.

### What are the steps to reproduce?

```php
<?php

require __DIR__ . '/vendor/autoload.php';

use PhpOffice\PhpSpreadsheet\Spreadsheet;
use PhpOffice\PhpSpreadsheet\Writer\Ods;

$spreadsheet = new Spreadsheet();
$activeWorksheet = $spreadsheet->getActiveSheet();

$activeWorksheet->setCellValue('A1', 123456.789);
$activeWorksheet->getStyle('A1')->getNumberFormat()->setFormatCode('#,##0.00');

$writer = new Ods($spreadsheet);
$writer->save('test-fmt-01.ods');
```

Have also tried all kind of variants like

```php
$activeWorksheet->setCellValue([1, 1], 123456.789);
$activeWorksheet->setCellValueExplicit($loc, $val, DataType::TYPE_NUMERIC);

// ...

$activeWorksheet->getStyle('A1')->applyFromArray([
	'numberFormat' => [
		'formatCode' => '0,00E+00'
	]
]);

$activeWorksheet->getStyle('A1')->getNumberFormat()->setFormatCode(
	PhpOffice\PhpSpreadsheet\Style\NumberFormat::FORMAT_CURRENCY_EUR
);

// ...

$writer = \PhpOffice\PhpSpreadsheet\IOFactory::createWriter($spreadsheet, "Ods");
$writer->save("out/test01.ods");

```

+ dozens of more tweaks.

Have also tried all sorts of variants with numbers; as string, as int,as float,  ...

If I do:

```php
		var_dump( ... getFormatCode());
		 ... setFormatCode($fmt);
		var_dump(... getFormatCode());
```

The correct format is dumped after applying. But never applied to the written file.

---

Have also tried to save a file with formatting using LibreOffice then reading it with PhpSpreadsheet - but the `var_dump(... getFormatCode());` only say `string(7) "General"`. Have also tried to scan 100x100 cells but no cell has any format.

### What features do you think are causing the issue

- [X] Reader
- [X] Writer
- [X] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

[...]

PR #4799, opened 2026-02-06
Ods Writer Support Some Number Formats

### LINKED ISSUE #4798: NumberFormat Ignored When Writing Ods
Issue raised by @sebastianbergmann in issue #3961. This issue is different, and I have opened a new issue for it.

Xlsx and Xls use a string like `0.##` for their number formats. Ods takes a different approach for number styles in its xml, e.g.:
```xml
<number:number-style style:name="N4">
<number:number number:decimal-places="2" number:min-decimal-places="2" number:min-integer-digits="1" number:grouping="true"/>
</number:number-style>
```
PhpSpreadsheet is not set up to translate from one format to the other, so ignores NumberFormat when writing Ods. I think it should be relatively easy to at least get the formatted value recorded in the Xml (and therefore displayed when the spreadsheet is opened for read).

It will be more difficult to convert the style to a format that Ods likes and record that in the xml; I note that as an option here, but a PR responding to this issue may not include this part.

oleibman:
Fix #4798. Partially address #3961. Ods Reader supports very little related to styling. Ods Writer supports many styling details, but has not heretofore supported Number Formatting. 3961 addresses both issues; I created 4798 specifically for the Writer side.

Writing number formats to Excel is pretty simple - you just supply a string and that is used directly in the Xml. Ods is much more complicated - it requires Xml nodes that give a complete description of the styling. For that reason, it is difficult and painstaking to convert from the string that Excel (and PhpSpreadsheet) uses to what Ods requires.

This PR provides code to support almost all the styles defined as constants in `Style/NumberFormat`. It also allows the user to add code to handle otherwise unhandled styles. New Sample55_DefinedStyles demonstrates the use of all the constant styles, plus the addition of a couple of custom styles. I may be amenable to adding some unsupported styles to the built-in list, but the custom style option will always be around in case I am being slow or unreasonable.

This PR does not fully support Ods Reader handling of styles. However, based on the new Writer output, it will often be able to guess the true type of numeric items and assign an appropriate style for that type. So, for example, if it can identify the field as a date, it will assign a date style. It will not always match the style in the sheet being read, but it is a big advance from just formatting the data as a generic number.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

  - [x] New sample has been added
- [x] Code style is respected


oleibman:
Superseded by #4806 due to failed merge conflict resolution.

PR #4801, opened 2026-02-06
Add support for image CSS size in millimetres

### LINKED ISSUE #4800: Error loading spreadsheet with image size specified in millimetres
When an Excel spreadsheet contains an image with width or height specified in millimetres, loading the file produces an error:

```
PHP Fatal error:  Uncaught TypeError: PhpOffice\PhpSpreadsheet\Worksheet\BaseDrawing::setWidth(): Argument #1 ($width) must be of type int, string given, called in phpspreadsheet/src/PhpSpreadsheet/Reader/Xlsx.php on line 1457 and defined in phpspreadsheet/src/PhpSpreadsheet/Worksheet/BaseDrawing.php:329
Stack trace:
#0 phpspreadsheet/src/PhpSpreadsheet/Reader/Xlsx.php(1457): PhpOffice\PhpSpreadsheet\Worksheet\BaseDrawing->setWidth()
#1 phpspreadsheet/src/PhpSpreadsheet/Reader/BaseReader.php(269): PhpOffice\PhpSpreadsheet\Reader\Xlsx->loadSpreadsheetFromFile()
#2 test.php(16): PhpOffice\PhpSpreadsheet\Reader\BaseReader->load()
#3 {main}
  thrown in phpspreadsheet/src/PhpSpreadsheet/Worksheet/BaseDrawing.php on line 329
```

Example file: [test.xlsx](https://github.com/user-attachments/files/25127752/test.xlsx)

Code to reproduce:

```php
<?php

$spreadsheet = $reader->load('test.xlsx');
?>
```

(Tested with versions 5.2.0 and 5.4.0 as well as current master branch.)

Looks like `Xlsx::toCSSArray()`, which parses CSS attributes for values with units, does not support millimetres, so it leaves the "mm" units in place. When a value like that is passed to `Drawing::setWidth()` or `Drawing::setHeight()`, which expect ints, the error occurs.

This is easy enough to fix, by dividing the numeric value by 10 and using the existing `Font::centimeterSizeToPixels()` method.

However, because that method returns a float, since PHP 8.1 another error occurs:

```
PHP Deprecated:  Implicit conversion from float-string "128.5039370094" to int loses precision in phpspreadsheet/src/PhpSpreadsheet/Worksheet/BaseDrawing.php on line 329
PHP Stack trace:
PHP   1. {main}() test.php:0
PHP   2. PhpOffice\PhpSpreadsheet\Reader\BaseReader->load($filename = 'test.xlsx', $flags = *uninitialized*) test.php:16
PHP   3. PhpOffice\PhpSpreadsheet\Reader\Xlsx->loadSpreadsheetFromFile($filename = 'test.xlsx') phpspreadsheet/src/PhpSpreadsheet/Reader/BaseReader.php:269
PHP   4. PhpOffice\PhpSpreadsheet\Worksheet\BaseDrawing->setWidth($width = '128.5039370094') phpspreadsheet/src/PhpSpreadsheet/Reader/Xlsx.php:1457
```

This is because `Xlsx::toCSSArray()` takes the result of the `Font::*SizeToPixels()` methods, which could be an int or float, and converts it to string. It should probably convert to int instead, since that's what `Drawing::setWidth()` or `Drawing::setHeight()` expect, as mentioned above. The CSS value being passed to those methods should also probably be converted to float, not int, to avoid losing precision too early.

IndrekHaav:
This is:

- [X] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

This fixes a TypeError when an Excel spreadsheet contains an image with width or height specified in millimetres.

Additionally, it fixes implicit conversions from float to int (deprecated since PHP 8.1) when the image width or height contains fractions, or conversion to pixels results in fractions.

closes #4800

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ and this project adheres to [Semantic Versioning](https://semver.org). Thia is a
 - Improve performance in value binders. [PR #4780](https://github.com/PHPOffice/PhpSpreadsheet/pull/4780)
 - Handle Unions as Function Arguments. [Issue #4656](https://github.com/PHPOffice/PhpSpreadsheet/issues/4656) [Issue #316](https://github.com/PHPOffice/PhpSpreadsheet/issues/316) [Issue #503](https://github.com/PHPOffice/PhpSpreadsheet/issues/503) [PR #4657](https://github.com/PHPOffice/PhpSpreadsheet/pull/4657)
 - Unexpected Behavior of CONCATENATE. [Issue #4061](https://github.com/PHPOffice/PhpSpreadsheet/issues/4061) [PR #4797](https://github.com/PHPOffice/PhpSpreadsheet/pull/4797)
+- Image Css size in millimeters. [Issue #4800](https://github.com/PHPOffice/PhpSpreadsheet/issues/4800) [PR #4801](https://github.com/PHPOffice/PhpSpreadsheet/pull/4801)
 
 ## 2026-01-10 - 5.4.0
 


oleibman:
Please let me know if you need help with the Phpstan problems.

Your PR needs to add unit test(s) to demonstrate that your fix works correctly. Adding the test should also ensure that Coveralls is happy.

oleibman:
I think you can probably satisfy Phpstan by changing the `if` statements on lines 2226-2240 to a single `if...elseif` structure:
```php
if (... 'px') {
    ...
} elseif (... 'pt') {
    ...
} elseif (... 'in') {
    ...
// etc
```

IndrekHaav:
Thanks for the feedback. I tweaked the `if` statements and added a unit test, looks better now.

oleibman:

PR #4833, opened 2026-03-11
Optimize XLS (BIFF8) reader performance with reduced per-cell overhead

kemo:
## Summary
- Eliminates redundant `updateInCollection()` calls during cell reads: XF index assignment is reordered before `setValueExplicit()`, and `Cell::setXfIndexNoUpdate()` is used **only** where a collection-updating call (`setValueExplicit()` / `setCalculatedValue()`) is guaranteed to follow on every path. Sites without that guarantee (BOOLERR, BLANK, MULBLANK) use the regular updating `setXfIndex()`.
- Replaces the O(n²) char-by-char string concatenation in the SST reader (UTF-16 expansion across CONTINUE records) with a single `chunk_split()` call.
- Caches the sheet title per sheet and pre-computes cell coordinate strings in the per-cell reader methods, instead of calling `getTitle()` and concatenating inline for every cell.
- Caches the SST entry locally in `readLabelSst` and the formula result type byte in `readFormula` to reduce repeated lookups.

(Earlier revisions also converted the splice-offset lookup loops and added a `cachedReadFilter` property; both were dropped — the loop conversion showed no measurable benefit and was reverted in 52315526e, and direct `readFilter` property access landed separately via #4919.)

## Changes
- `src/PhpSpreadsheet/Cell/Cell.php`: new internal `setXfIndexNoUpdate()` that sets the XF index without triggering `updateInCollection()`
- `src/PhpSpreadsheet/Reader/Xls.php`: per-cell reader methods optimized as above; SST string concatenation replaced
- `src/PhpSpreadsheet/Reader/Xls/LoadSpreadsheet.php`: sets the cached sheet title at the start of each sheet iteration
- `samples/XlsReaderBenchmark.php`: reproducible benchmark (`php samples/XlsReaderBenchmark.php [iterations] [cellCounts]`)

## Benchmark results

See [the benchmark comment](https://github.com/PHPOffice/PhpSpreadsheet/pull/4833#issuecomment-4890121138) for current numbers from `samples/XlsReaderBenchmark.php`: 2-5% faster loads depending on file size, identical peak memory.

## Test plan
- [x] All XLS reader tests pass (82 tests, 1425 assertions)
- [x] Cell values, styles, and XF indices identical to master
- [x] Rich text formatting in SST strings preserved
- [x] Shared formulas resolve correctly
- [x] Blank cells retain their style information (via updating `setXfIndex()`)


oleibman:
Is there a Benchmark test you can add to this PR?

kemo:
Sure — I'll add a benchmark test for the BIFF8 reader to this PR.

oleibman:
I am trying to understand this change well enough to move forward with it. One question right off the bat. What is the advantage of "splice offset loops converted to indexed for-loops"?

oleibman:
I don't think there is a need for `cachedReadFilter`; there is already a `readFilter` property that can be used in its place. However, using the property rather than the method call is such a good idea, and applicable to other formats besides Xls, that I have created PR #4919 to implement it. If I decide to merge that PR, this one will wind up with a small number of easily resolved merge conflicts, which I will take care of when that happens. I believe that the rest of this PR will still be worthwhile, except possibly the loop replacements that I asked about above.

oleibman:
PR #4919 is now merged; merge conflicts in this PR are resolved.

src/PhpSpreadsheet/Reader/Xls.php
```diff
@@ -3038,10 +3039,10 @@ protected function readFormula(): void
                 $value = self::extractNumber(substr($recordData, 6, 8));
             }
 
-            $cell = $this->phpSheet->getCell($columnString . ($row + 1));
+            $cell = $this->phpSheet->getCell($cellCoordinate);
             if (!$this->readDataOnly && isset($this->mapCellXfIndex[$xfIndex])) {
                 // add cell style
-                $cell->setXfIndex($this->mapCellXfIndex[$xfIndex]);
+                $cell->setXfIndexNoUpdate($this->mapCellXfIndex[$xfIndex]);
```
oleibman, line 3047: This call is not immediately followed by `setValueExplicit`. How do we know that is safe to skip the update here?

src/PhpSpreadsheet/Reader/Xls.php
```diff
@@ -3160,7 +3162,11 @@ protected function readBoolErr(): void
             // offset: 7; size: 1; 0=boolean; 1=error
             $isError = ord($recordData[7]);
 
-            $cell = $this->phpSheet->getCell($columnString . ($row + 1));
+            $cell = $this->phpSheet->getCell($cellCoordinate);
+            if (!$this->readDataOnly && isset($this->mapCellXfIndex[$xfIndex])) {
+                // add cell style
+                $cell->setXfIndexNoUpdate($this->mapCellXfIndex[$xfIndex]);
```
oleibman: This call is not immediately followed by `setValueExplicit`. How do we know that is safe to skip the update here?

src/PhpSpreadsheet/Reader/Xls.php
```diff
@@ -3210,14 +3211,15 @@ protected function readMulBlank(): void
...
             for ($i = 0; $i < $length / 2 - 3; ++$i) {
                 $columnString = Coordinate::stringFromColumnIndex($fc + $i + 1);
 
                 // Read cell?
-                if ($this->readFilter->readCell($columnString, $row + 1, $this->phpSheet->getTitle())) {
+                if ($this->readFilter->readCell($columnString, $rowIndex, $this->phpSheetTitle)) {
                     $xfIndex = self::getUInt2d($recordData, 4 + 2 * $i);
                     if (isset($this->mapCellXfIndex[$xfIndex])) {
-                        $this->phpSheet->getCell($columnString . ($row + 1))->setXfIndex($this->mapCellXfIndex[$xfIndex]);
+                        $this->phpSheet->getCell($columnString . $rowIndex)->setXfIndexNoUpdate($this->mapCellXfIndex[$xfIndex]);
```
oleibman: This call is not immediately followed by `setValueExplicit`. How do we know that is safe to skip the update here?

src/PhpSpreadsheet/Reader/Xls.php
```diff
@@ -3296,14 +3298,16 @@ protected function readBlank(): void
...
         // Read cell?
-        if ($this->readFilter->readCell($columnString, $row + 1, $this->phpSheet->getTitle())) {
+        if ($this->readFilter->readCell($columnString, $rowIndex, $this->phpSheetTitle)) {
             // offset: 4; size: 2; XF index
             $xfIndex = self::getUInt2d($recordData, 4);
 
             // add style information
             if (!$this->readDataOnly && $this->readEmptyCells && isset($this->mapCellXfIndex[$xfIndex])) {
-                $this->phpSheet->getCell($columnString . ($row + 1))->setXfIndex($this->mapCellXfIndex[$xfIndex]);
+                $this->phpSheet->getCell($columnString . $rowIndex)->setXfIndexNoUpdate($this->mapCellXfIndex[$xfIndex]);
```
oleibman: This call is not immediately followed by `setValueExplicit`. How do we know that is safe to skip the update here?

kemo:
As promised, here are benchmark numbers. I added `samples/XlsReaderBenchmark.php`, which generates .xls files with mixed numeric/string/formula content and reports median load times over several iterations (CLI-parameterizable: `php samples/XlsReaderBenchmark.php [iterations] [cellCounts]`).

Results on PHP 8.5.2 (xdebug off, best median of several alternating master/branch runs to filter out background-load noise; each file has 10 columns per row — 4 numeric, 4 string, 2 formula cells, plus header styling and number formats):

| Cells   | File size | master (ms) | branch (ms) | Delta  | Peak memory |
|---------|-----------|-------------|-------------|--------|-------------|
| 10,000  | 251 KB    | 107.5       | 102.6       | -4.6%  | 20.0 MB     |
| 50,000  | 1,236 KB  | 530.8       | 510.2       | -3.9%  | 49.0 MB     |
| 100,000 | 2,466 KB  | 1082.8      | 1058.9      | -2.2%  | 86.9 MB     |

Peak memory is identical between master and branch at every size. A modest but consistent 2-5% win from removing the per-cell `getTitle()`/`updateInCollection()` overhead.

src/PhpSpreadsheet/Reader/Xls.php
kemo: I missed these comments until now, will look into it more.

src/PhpSpreadsheet/Reader/Xls.php
kemo: This one is safe on all paths: everything after this point in `readFormula()` funnels into `$cell->setCalculatedValue($value, ...)` at the end of the block (including the BIFF8 shared-formula path that skips `setValueExplicit`), and `setCalculatedValue()` unconditionally calls `updateInCollection()`. I've added a code comment in 5db3842d9 stating that invariant so it doesn't rely on tribal knowledge.

src/PhpSpreadsheet/Reader/Xls.php
kemo: Good catch — this one was not safe. If the type byte at offset 7 is neither 0 nor 1 (corrupt file), neither `case` runs `setValueExplicit`, and the style change would then only persist if the cell happened to be newly created (`Cells::add()` marks the current cell dirty). The original code updated unconditionally after the switch. Reverted to the updating `setXfIndex()` in 5db3842d9.

src/PhpSpreadsheet/Reader/Xls.php
kemo: You're right to question this one — blank cells never receive a value write at all. It happens to work because these cells are newly created, and `Cells::add()` marks the current cell dirty so `storeCurrentCell()` persists it on the next cell switch. But that's an implementation subtlety rather than a guarantee (a pre-existing cell re-fetched via `Cells::get()` is not dirty, so with a serializing cache backend the style write could be silently lost). Not worth the risk on a cold path — reverted to the updating `setXfIndex()` in 5db3842d9.

src/PhpSpreadsheet/Reader/Xls.php
kemo: You're right to question this one — blank cells never receive a value write at all. It happens to work because these cells are newly created, and `Cells::add()` marks the current cell dirty so `storeCurrentCell()` persists it on the next cell switch. But that's an implementation subtlety rather than a guarantee (a pre-existing cell re-fetched via `Cells::get()` is not dirty, so with a serializing cache backend the style write could be silently lost). Not worth the risk on a cold path — reverted to the updating `setXfIndex()` in 5db3842d9.

kemo:
Closing the loop on your splice-offsets question: there was no real advantage, and you've since reverted that part yourself (52315526e) — the conversion predated proper measurement and didn't survive it. Apologies for the stale PR description claiming it; I've updated the description to match what the branch actually contains now:

- one `getCell()`-and-update per cell instead of repeated lookups (`setXfIndexNoUpdate()` is now only used where a collection-updating call is guaranteed to follow — see the resolved inline threads and 5db3842d9);
- the SST char-padding loops replaced with a single `chunk_split()` (that one is measurable: it turns a per-character PHP-level concatenation loop into one C-level call for CONTINUE-spanning strings);
- cached sheet title / precomputed coordinate strings in the per-cell hot path.

Benchmark numbers for the current state of the branch are in the comment above.

oleibman:

PR #4834, opened 2026-03-11
Add parallel Xlsx writing via pcntl_fork

kemo:
## Summary

- Adds opt-in parallel processing infrastructure (`src/PhpSpreadsheet/Parallel/`) using `pcntl_fork()` with automatic sequential fallback
- Parallelizes per-sheet XML generation in `Writer\Xlsx::save()` — the single most CPU-intensive phase of writing
- **Zero overhead when disabled** (default) — no forking, no temp files, identical code path to before
- Safely degrades: 2-core machines auto-detect 1 worker → sequential path, no fork attempted

### Architecture

| Component | Purpose |
|-----------|---------|
| `ParallelExecutor` | Public API: `map(tasks, worker)` with auto-backend detection, memory safety checks |
| `PcntlBackend` | `pcntl_fork()` + temp-file IPC, batched execution, timeout + zombie reaping |
| `SequentialBackend` | No-op fallback, always available |
| `CpuDetector` | Multi-strategy CPU count detection (pcntl_cpu_count → /proc/cpuinfo → sysctl → nproc → fallback 2), cached |

### Writer instance API

Parallel settings live on the `Writer\Xlsx` instance (not global statics):

```php
$writer = new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet);
$writer->setParallelEnabled(true);
$writer->setMaxWorkers(4); // Optional: null = auto-detect based on CPU count
$writer->save('output.xlsx');
```

### Worker count formula

`min(cpuCount - 1, taskCount, 8)` — reserves 1 core for parent, caps at 8 to prevent IPC overhead.

### Safety

- **Child process isolation**: `$isChild` flag prevents children from running parent cleanup (could delete unread temp files)
- **Memory limit check**: Estimates 30% dirty-page overhead per fork, auto-reduces worker count if tight
- **Timeout**: Configurable per-task timeout (default 60s), parent sends SIGTERM to hung children
- **Zombie prevention**: `pcntl_waitpid()` in `finally` block reaps all children, even on exceptions
- **posix_kill guarded**: `function_exists()` check, numeric signal value (15) avoids undefined constant

### Benchmarks (Apple M2 Pro, 11 cores)

| Workload | Sequential | Parallel | Speedup |
|----------|-----------|----------|---------|
| 5 sheets × 5K rows (250K cells) | 2.64s | 1.82s | **1.45x** |
| 10 sheets × 5K rows (500K cells) | 6.66s | 4.07s | **1.63x** |

Speedup is bounded by ZIP compression (~45% of `save()` time) which runs sequentially after XML generation (~48%).

## Test plan

- [x] 44 tests, 579 assertions — all passing
- [x] PHPStan level 10 clean (0 new errors)
- [x] Platform-specific and defensive error paths annotated with `@codeCoverageIgnore`
- [x] `ParallelXlsxWriterTest`: validates round-trip read-back of parallel-written files
- [x] `ParallelXlsxWriterTest`: sequential vs parallel produce identical cell data
- [x] `ParallelXlsxWriterTest`: single-sheet parallel-enabled falls back to sequential
- [x] Timeout test: verifies hung children are killed and exception propagated
- [x] Error propagation test: child exceptions surface in parent
- [x] Memory limit parsing: covers `-1`, `M`, `G` suffix formats
- [x] Xlsx writer parallel setter/getter coverage


oleibman:
I have for some time been actively engaged in eliminating static settings in favor of instance instance properties, to make testing easier and to avoid problems in shared environments. The effort is far from finished, but I'd still like to avoid static when possible. Here, you could make your new switches an instance variable of the writer rather than a global static setting. Would you be able to make that change?

kemo:
parallelEnabled and maxWorkers are instance properties on the Xlsx writer, not static settings. The writer is configured per-instance via setParallelEnabled() and setMaxWorkers().

ddevsr:
@kemo You can use `fidry/cpu-core-counter` instead for complexity detection CPU like virtualhost. That library used by PHPStan, Rector, PHP-CS-Fixer for parallel.

oleibman:
`pcntl_fork` is not available in Windows. Since this is opt-in, that's not necessarily a show-stopper, although it might be. Regardless, you certainly need to add code to ignore the opt-in when `pcntl_fork` is not available.

kemo:
I've pushed a round of hardening based on the review feedback plus an internal audit of the fork lifecycle:

**Addressing review comments:**
- @ddevsr — good call: `CpuDetector` now delegates to `fidry/cpu-core-counter` instead of the hand-rolled detection strategies (the shell-outs to `nproc`/`sysctl` are gone).
- @oleibman — on Windows/missing pcntl: the opt-in was already ignored when `pcntl_fork` is unavailable (it falls back to the sequential writer), but that path was never actually exercised in CI since the Ubuntu runners always have pcntl. I've added a dedicated CI job that runs the Parallel test suite with `disable_functions=pcntl_fork` to prove the graceful fallback, and tightened `isAvailable()` further: forking is now also refused outside the CLI SAPI (php-fpm/Apache would be unsafe).

**Fork-lifecycle hardening (new commits):**
- A crashed/OOM-killed child previously produced a silently corrupt xlsx (its empty result file was written verbatim as sheet XML). Results now cross the process boundary in a validated envelope; a child that dies before writing a complete result fails the save with an exception that reports how the child died. There's a regression test that SIGKILLs a child mid-task.
- Children now terminate via SIGKILL-on-self instead of `exit(0)`, so they never run destructors/shutdown functions/output buffers inherited from the parent (e.g. an application's shared DB connection can no longer be torn down by a forked child).
- The previous hard-coded 60s timeout is gone: the default is now no time limit (large sheets are the whole point of this feature), configurable via `$writer->setParallelTimeout($seconds)`. Timed-out children are SIGTERM'd, escalated to SIGKILL after a grace period, and always reaped — no more orphaned workers after a timeout.
- Worker exceptions now propagate with their original class and stack trace instead of a bare message.

**Also added:** documentation in `docs/topics/reading-and-writing-to-file.md`, a CHANGELOG entry, a merge with current master, and tests bringing the `Parallel` namespace to 100% line coverage (which also fixes the Coveralls check).

ddevsr:
@kemo Since this optional, move `fidry/cpu-core-counter` in `required` to `suggest`

oleibman:
@kemo You have given a thumbs-up to the comment by @ddevsr to move the fidry package from required to optional (and, of course, to require-dev), which suggests that you are amenable to the idea. I think that is my preference as well. Can you make this change? I think it would require you to check the availability of fidry at the same time as you check the availability of pcntl_fork.

PR #4863, opened 2026-04-28
Add border color, fill opacity, and shape type to Comments (#1790)

### LINKED ISSUE #1790: Add Comment Box Style
can add comment box style()->shapeType!
like Drawing!

AnnaGaile:
Resolved #1790 by adding shapeType and styling support for Comments. Users can now programmatically set the box shape, background color, border color, and transparency. This replaces the previously hardcoded rectangles with fully customizable comment objects. Unit tests are included to verify the VML output for all new properties.

### DIFF
--- a/src/PhpSpreadsheet/Comment.php
+++ b/src/PhpSpreadsheet/Comment.php
@@ -54,6 +54,21 @@ class Comment implements IComparable, Stringable
     private Color $fillColor;
 
     /**
+     * Comment border color.
+     */
+    private Color $borderColor;
+
+    /**
+     * Comment fill opacity.
+     */
+    private float $fillOpacity = 1.0;
+
+    /**
+     * Comment shapeType.
+     */
+    private int $shapeType = 202;
+
+    /**
      * Alignment.
      */
     private string $alignment;
@@ -80,6 +95,7 @@ class Comment implements IComparable, Stringable
         $this->author = 'Author';
         $this->text = new RichText();
         $this->fillColor = new Color('FFFFFFE1');
+        $this->borderColor = new Color(Color::COLOR_BLACK);
         $this->alignment = Alignment::HORIZONTAL_GENERAL;
         $this->backgroundImage = new Drawing();
     }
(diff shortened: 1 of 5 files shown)


oleibman:
This looks very promising. However, it also seems incomplete - it takes care of writing, but not reading. Can you add code to Xlsx Reader so that it will recognize the attributes that the writer is now generating?

oleibman:
We are getting close. However, I do not really understand the tests. They prove that you are reading and writing the attributes correctly, but I do not understand what you intend the notes to be.

testCommentStylingOutput - this comment does not have any text. And, even when I add text, the resulting note seems to be a blue diagonal line through the comment text. (I apologize for not being able to figure out how to capture the output in a screenshot.) Is this what you intended? I see you've selected a blue border, and that matches the diagonal line. I don't see the fill color anywhere. I don't know what shape 203 is supposed to be.

testCommentStylingOutputRoundTrip - this appears to be the same as the previous case, except this time there is some comment text. So ... essentially the same questions.

multipleCommentStyling first comment. This is a good test. I see the red background with a lot of opacity and the green border. It uses what I think is the default shape (202).

multipleCommentStyling second comment. I see a transparent background, not a (somewhat opaque) blue one. Strangely enough, when I attempt to edit the note, the blue background does appear - do you know why it is behaving in this now-you-see-me-now-you-don't manner? I don't see a yellow, or indeed any, border, even when editing the note. I don't see a shape (it looks like a plain box when I edit it), and I don't know what 204 is supposed to be anyhow.

Can you make it clearer what the notes in these tests are supposed to be? In addition, can you add a Sample demonstrating variants of what you can do with the new styling you are adding? This will go a long way to helping me understand what you are trying to accomplish, and give the user base a good introduction to how to make use of your new feature. Please include some non-boxy shapes - I'm not really familiar with this feature, but I see you can use shapes like triangles or smiley faces - your sample doesn't have to include those, but it should use some "unexpected" designs.

I will note that there is no current support in PhpSpreadsheet for "shapes" in other contexts (our assessment so far is that adding it would be difficult), and this may be the reason why your test cases appear as they do. If that is the case, perhaps we need to restrict your change to border color and fill opacity, unless, of course, you are able to unlock the mystery of how to specify shapes specifically for comments (or everywhere, if you can figure that out).

oleibman:
No response in several weeks. Converting to draft for now.

PR #4884, opened 2026-05-23
Allow whitelisting stream wrappers in File::prohibitWrappers() (#4881)

### LINKED ISSUE #4881: prohibitWrappers() breaks testing with mikey179/vfsstream — no way to whitelist custom stream wrapper
[File::prohibitWrappers()](https://github.com/PHPOffice/PhpSpreadsheet/blob/master/src/PhpSpreadsheet/Shared/File.php#L142) unconditionally rejects any URL scheme with more than one character. This breaks unit tests that use [mikey179/vfsstream](https://github.com/bovigo/vfsStream) — a widely-used PHP library for mocking the filesystem in tests.

vfsStream registers a `vfs://` stream wrapper and provides `vfs://root/...` paths that are fully compatible with standard PHP file functions (`file_put_contents`, `fopen`, `file_get_contents`, `is_file`, etc.). PhpSpreadsheet itself internally uses those functions, so a `vfs://` path would work but `prohibitWrappers()` aborts before even trying.

**Environment**                                                                                                                                                                                                                                                                                     
- phpoffice/phpspreadsheet: 5.7.0                                                                                                                                                                                                                                                               
- mikey179/vfsstream: v1.6.12                                                                                                                                                                                                                                                                   
- PHP: 8.4.8

**Expected behavior**
Tests that use vfsStream to avoid real disk I/O should work. The library should either:
1. Provide a whitelist API so callers can opt-in to specific safe stream wrappers:
    - `File::allowStreamWrappers(['vfs']); // e.g. for test environments`
2. Or check whether the scheme is a registered PHP stream wrapper (via `stream_get_wrappers()`) and allow those, while still blocking dangerous protocols like `phar://`, `http://`, `expect://`


The comment in the source already hints at this:
> A whitelist of protocols may be added if needed in future.

YoussefMansour9:
Fixes #4881

The `prohibitWrappers()` method was added as part of a security advisory but is too strict: it blocks all stream wrappers with multi-character schemes, including legitimate ones like `vfs://` used by the vfsStream testing library.

This adds `File::allowStreamWrappers(["vfs"])` to selectively whitelist specific wrappers. The existing default behavior (blocking all multi-character schemes) is preserved.

### DIFF
--- a/src/PhpSpreadsheet/Shared/File.php
+++ b/src/PhpSpreadsheet/Shared/File.php
@@ -134,17 +134,34 @@ class File
     }
 
     /**
-     * All filenames starting with protocol (e.g. phar://) are prohibited.
+     * @var list<string>
+     */
+    private static array $allowedWrappers = [];
+
+    /**
+     * Register stream wrappers that should be permitted by prohibitWrappers.
+     * By default, all stream wrappers with multi-character schemes (e.g. phar://) are blocked.
+     * Call this method to allow specific wrappers for your use case.
+     *
+     * @param list<string> $wrappers
+     */
+    public static function allowStreamWrappers(array $wrappers): void
+    {
+        self::$allowedWrappers = array_values(array_unique([...self::$allowedWrappers, ...$wrappers]));
+    }
+
+    /**
+     * All filenames starting with protocol (e.g. phar://) are prohibited,
+     * unless the wrapper has been explicitly allowed via allowStreamWrappers().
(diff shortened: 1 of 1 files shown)


oleibman:
Thank you for submitting this. I encourage you to continue to look for outstanding issues which you can solve. Unfortunately, both of the issues which you have chosen to address today come with complications - one has security implications, and one involves a potential breaking change. For those reasons, I need to reject them both. I truly hope this will not be overly discouraging for you.

PR #4885, opened 2026-05-23
Add listWorksheetInfo() to IReader interface (#4883)

### LINKED ISSUE #4883: listWorksheetInfo() not in IReader
listWorksheetInfo() is not declared in the IReader interface, this seems an oversight, but is it intentional? Is it recommended to depend on BaseReader, or is there another interface that could be relied on for this?

YoussefMansour9:
Fixes #4883

The listWorksheetInfo() method exists in BaseReader and all concrete reader classes (Csv, Xls, Xml, Ods, Gnumeric, Html), but was missing from the IReader interface. Users type-hinting on IReader cannot call it without a cast.

### DIFF
--- a/src/PhpSpreadsheet/Reader/IReader.php
+++ b/src/PhpSpreadsheet/Reader/IReader.php
@@ -51,6 +51,13 @@ interface IReader
     public function canRead(string $filename): bool;
 
     /**
+     * Return worksheet info (Name, Last Column Letter, Last Column Index, Total Rows, Total Columns).
+     *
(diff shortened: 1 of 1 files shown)


oleibman:
Thank you for submitting this. I encourage you to continue to look for outstanding issues which you can solve. Unfortunately, both of the issues which you have chosen to address today come with complications - one has security implications, and one involves a potential breaking change. For those reasons, I need to reject them both. I truly hope this will not be overly discouraging for you.

PR #4891, opened 2026-05-31
Bug in FIXED Function

### LINKED ISSUE #4889: `FIXED` function is not properly formatting `Decimals` when `No_commas` is true
The [`FIXED` function](https://github.com/PHPOffice/PhpSpreadsheet/blob/5.7.0/src/PhpSpreadsheet/Calculation/TextData/Format.php#L69-L109) is not properly taking into account the `$decimals` parameter if `$noCommas` is `true`.

In Excel (and Google Sheets),
 `=FIXED(1507.5, 2, TRUE)`
results in
 `1507.50`
(no thousands separator, but two digits to the right of the decimal point).

In PHPSpreadsheet, that formula results in
 `1507.5`
with the wrong number of decimals.

The reason seems to be that the `number_format()` function is only called if `$noCommas === false`, when it should likely be called in both cases, just with different parameters for the thousands separator.

oleibman:
Fix #4889. Function was not outputting correct number of decimals when thousands separator was omitted.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected


src/PhpSpreadsheet/Calculation/TextData/Format.php
```diff
@@ -103,6 +103,13 @@ public static function FIXEDFORMAT(mixed $value, mixed $decimals = 2, mixed $noC
                 StringHelper::getDecimalSeparator(),
                 StringHelper::getThousandsSeparator()
             );
+        } else {
```
TobiasBg: How about shortening the full `if` block to
```php
$valueResult = number_format(
  $valueResult,
  $decimals,
  StringHelper::getDecimalSeparator(),
  ($noCommas === false) ? StringHelper::getThousandsSeparator() : ''
);
```
?

PR #4893, opened 2026-06-03
Phpstan Safe Keys

oleibman:
Phpstan 2.2 adds type-safe protection for array keys, which Php may convert from string to int. I suppose it can be useful to us for arrays indexed by column or cell address. The initial conversion for PhpSpreadsheet to make use of it is a bit of a pain, but it's done. It requires the use of more `@var` annotations than I would like. Also several cases where, even though `$x` is known to be a string, `$y="$x"` works out much better than `$y=$x`. And the occasional less precise use of `array<whatever>` instead of `array<string, whatever>` when we truly don't care if the key might be converted to int.

Almost all of this change is done in comments (doc blocks and annotations), but there are some small code changes.

This is:

- [ ] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests
- [x] using new feature in 3rd party software

Checklist:

- [x] Code style is respected

### DIFF
--- a/infra/DocumentGenerator.php
+++ b/infra/DocumentGenerator.php
@@ -17,7 +17,16 @@ class DocumentGenerator
     ];
 
     /**
-     * @param array<string, array{category: string, functionCall: string|string[], argumentCount: string, passCellReference?: bool, passByReference?: bool[], custom?: bool}> $phpSpreadsheetFunctions
+     * @param array<
+     *     array{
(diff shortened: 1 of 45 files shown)


oleibman:
PhpDocumentor does not understand `non-decimal-int-string`. I am going to treat that as a show-stopper. We can try this again some other time.

PR #4899, opened 2026-06-06
Document exception in IOFactory.php

AJenbo:

It lets down stream be more strict about there exception handeling and enables tools to notify them if the behavior changes in the future.


oleibman:
Thank you for your contribution. I will be making a formal release which will include this change this weekend.

PR #4946, opened 2026-07-28
Expand GAMMAINV bracket so tail quantiles are not clamped

gaoflow:
`GammaBase::calculateInverse` fixed its search bracket at `alpha*beta*5`, so any GAMMA.INV / GAMMAINV quantile beyond that ceiling was clamped to it:

```
GAMMAINV(0.9999, 1, 1)  ->  5        (expected ~9.2103)
```

The round-trip then breaks: `GAMMADIST(5, 1, 1, true)` is `0.9932`, not `0.9999`.

This grows the upper bound geometrically until it brackets the root. If the CDF stops increasing before that (the internal series approximation is past its usable range) it keeps the original bound instead of expanding into it, which also stops a probability the series cannot reach from running the bound off to a nonsensical value.

Tests cover the round-trip invariant and reference quantiles (from mpmath) for several extreme-tail cases, and that an unreachable tail probability stays bounded. Existing GAMMAINV/GAMMADIST fixtures and the full Statistical suite are unchanged.

### DIFF
--- a/tests/PhpSpreadsheetTests/Calculation/Functions/Statistical/GammaInvTest.php
+++ b/tests/PhpSpreadsheetTests/Calculation/Functions/Statistical/GammaInvTest.php
@@ -5,27 +5,29 @@ declare(strict_types=1);
 namespace PhpOffice\PhpSpreadsheetTests\Calculation\Functions\Statistical;
 
 use PhpOffice\PhpSpreadsheet\Calculation\Calculation;
-use PhpOffice\PhpSpreadsheet\Calculation\Statistical\Distributions\Gamma;
 
 class GammaInvTest extends AllSetupTeardown
 {
+    private function gammaInvFormulaResult(float $probability, float $alpha, float $beta): mixed
+    {
+        $sheet = $this->getSheet();
+        $sheet->getCell('A1')->setValue($probability);
+        $sheet->getCell('A2')->setValue($alpha);
+        $sheet->getCell('A3')->setValue($beta);
+        $sheet->getCell('B1')->setValue('=GAMMA.INV(A1, A2, A3)');
+
+        return $sheet->getCell('B1')->getCalculatedValue();
+    }
+
     /**
-     * Extreme upper-tail quantiles whose true root exceeds the old fixed
-     * alpha*beta*5 bracket ceiling, which used to clamp the result to it.
-     * Expected values from mpmath (findroot on the regularized gammainc).
+     * Upper-tail quantiles whose true root exceeds the old fixed alpha*beta*5
+     * bracket ceiling that used to clamp the result to it. Reference values from
+     * mpmath findroot on the regularized gammainc.
      */
     #[\PHPUnit\Framework\Attributes\DataProvider('providerGammaInvExtremeTail')]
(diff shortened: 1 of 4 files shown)


tests/PhpSpreadsheetTests/Calculation/Functions/Statistical/GammaInvTest.php
```diff
@@ -5,9 +5,48 @@
...
 {
+    /**
+     * Extreme upper-tail quantiles whose true root exceeds the old fixed
+     * alpha*beta*5 bracket ceiling, which used to clamp the result to it.
+     * Expected values from mpmath (findroot on the regularized gammainc).
+     */
+    #[\PHPUnit\Framework\Attributes\DataProvider('providerGammaInvExtremeTail')]
+    public function testGammaInvExtremeTail(float $expected, float $probability, float $alpha, float $beta): void
+    {
+        $x = Gamma::inverse($probability, $alpha, $beta);
```
oleibman: Please restructure this test so that it evaluates the function in the context of a spreadsheet (i.e. set cell values and a formula then determine the calculated result of the cell with the formula) rather than via a direct call to `Gamma`.

tests/PhpSpreadsheetTests/Calculation/Functions/Statistical/GammaInvTest.php
```diff
@@ -5,9 +5,48 @@
...
+            'p=0.9999 alpha=1 beta=1' => [9.210340371976, 0.9999, 1.0, 1.0],
+            'p=0.99995 alpha=1 beta=1' => [9.903487552536, 0.99995, 1.0, 1.0],
+            'p=0.9999 alpha=0.5 beta=2' => [15.136705226623, 0.9999, 0.5, 2.0],
+            'p=0.9999 alpha=1 beta=2' => [18.420680743952, 0.9999, 1.0, 2.0],
+        ];
+    }
+
+    public function testGammaInvUnreachableTailStaysBounded(): void
+    {
+        // A probability the forward series cannot reach must not send the
```
oleibman: This test is a little confusing. A result less than a given value is unexpected; we would normally expect equality. I understand that there may be excellent reasons for that not to be the case here. Can you just explain what is going on here? Also, if this is an expected result, please expand the doc-block for the `Gamma::inverse` function to indicated that this will sometimes happen.

oleibman:
Thank you - this looks promising. I have added some review comments.

binggao1230:
Done. Both extreme-tail tests now evaluate through a cell formula (`=GAMMA.INV(A1, A2, A3)`) rather than calling `Gamma::inverse` directly.

On the "less than": that case (`GAMMA.INV(0.9999999, 1, 1)`) is far enough into the tail that the forward-CDF series plateaus below the target probability, so the root can't be bracketed. Instead of letting the new expansion run away, it stops at the original `alpha*beta*5` ceiling and returns 5.0 there — the assertion just guards that it stays bounded. The other four cases now assert equality against the mpmath reference, and I added a note about the bounded fallback to the `Gamma::inverse` doc-block.

oleibman:

PR #4953, opened 2026-08-04
Fix gamma-family overflow for large shape parameters

gaoflow:
Follow-up promised in #4945, and built on top of it (the earlier commits here are #4945's; I'll rebase once it lands, leaving only the last commit).

**`GAMMA.INV` breaks for large alpha even with an accurate CDF.** The Newton-step pdf computes `Gamma(a)`, `b ** a` and `value ** (a - 1)` in linear domain, which overflow while the density itself is a small representable number. For alpha in ~[143, 171.62] the pdf is INF, so `dx = error / INF = 0` reads as convergence and the first bisection midpoint comes back silently — `GAMMA.INV(0.5, 143, 1)` returns 358.0 (true value 142.667). Above 171.62, `Gamma(a)` is INF and the result is `#NUM!`. The same linear-domain pattern breaks the `GAMMA.DIST` / `CHISQ.DIST` / `F.DIST` densities (`CHISQ.DIST(500, 500, FALSE)` and `F.DIST(1, 300, 300, FALSE)` are `#NUM!` today, both fine in Excel) and `GAMMALN` above 171.62, which went through `Gamma(x)` although `logGamma` sits unused in the same class. These now evaluate in log domain via `logGamma`; value=0 edge behaviour is preserved.

**The incomplete-gamma iteration cap starves for huge shape.** Near x ≈ a both the series and the continued fraction need O(sqrt(a)) terms, so with the fixed 256 cap `GAMMA.INV(0.5, 10000, 1)` returned 10001.0 (true 9999.667), and `CHISQ.INV` drifted the same way at high df. The cap now scales as max(256, 10·sqrt(a)).

Remaining `gammaValue` call sites audited: `GAMMA()`/`FACT()` overflows are genuine (Excel also gives `#NUM!` there) and Beta already uses `logBeta` — left unchanged.

Test reference values are mpmath (50-digit) cross-checked against scipy; results agree to ~1e-11 or better. A 618-point old-vs-new sweep over the non-overflowing region matches to ≤1.3e-10 — the residual is `gammaValue`'s own Lanczos error, which the log-domain path removes; the two GammaLnTest array expectations were updated to the exact values for that reason. Statistical suite: 983 tests green; phpcs / php-cs-fixer / phpstan clean.

This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests


oleibman:
You said you would rebase after #4945 lands, which is now the case. Please do so.

binggao1230:
Rebased onto master (`52da37f`), statistical suite still 983/983.

oleibman:

PR #4954, opened 2026-08-05
Fix BETAINV false convergence when the Beta CDF underflows

gaoflow:
`Beta::calculateInverse()` (BETAINV / BETA.INV) has a convergence test that can only ever fire on the wrong case:

```php
$result = self::distribution($guess, $alpha, $beta);
if (($result === $probability) || ($result === 0.0)) {
    $b = $a;   // collapse the bracket, we are done
}
```

`inverse()` rejects `probability <= 0` before `calculateInverse()` is reached, so inside the loop `$probability` is strictly positive. That makes `$result === 0.0` unable to mean "the guess is the root". It can only mean the CDF underflowed to zero at a guess far below the root, which is an ordinary "guess too low" and is what the existing `else { $a = $guess; }` branch is for. Instead the bracket is collapsed and the search returns whichever midpoint it happened to be holding.

The bisection starts at 1.0 and 0.5, so for a peaked Beta the third probe (0.25) is already dozens of standard deviations out and underflows. The search then ends on iteration three:

```
i=1 guess=1.00  cdf=1.0                 a=0.00 b=2.00
i=2 guess=0.50  cdf=0.5000000000031132  a=0.00 b=1.00
i=3 guess=0.25  cdf=0.0                 a=0.00 b=0.50   -> b = a, loop ends, 0.25 returned
```

Symptoms, no oracle required for the first two:

* `BETAINV(0.5, 5000, 5000)` returns `0.25`. `Beta(a, a)` is symmetric about 0.5, so its median is exactly 0.5.
* `BETAINV(0.1, 5000, 5000)` returns `0.25` as well — the same value for a different quantile, which on its own shows the loop ended on an artefact of the search rather than on an answer.
* `BETAINV(0.5, 20000, 3)` returns `0.5`; the true value is `0.99986631`.

Onset is not exotic. The Beta(alpha, 1) CDF is `x ** alpha`, so `BETAINV(p, alpha, 1)` has the closed form `p ** (1 / alpha)`, and the first wrong answer appears at `alpha = 1080`:

```
BETAINV(0.5, 1070, 1) = 0.999352408675   (closed form 0.999352408674507)
BETAINV(0.5, 1080, 1) = 0.5              (closed form 0.999358402966493)
```

### Extent

I ran the whole `(alpha, beta, probability)` surface — 9 alphas x 10 betas x 9 probabilities = 810 points, shapes from 0.5 to 20000 — against `scipy.special.betaincinv` (self-checked by round-tripping through `betainc`, and cross-checked against a 40-digit mpmath bisection at moderate shapes). 173 of the 810 points diverged by more than 1e-6 relative, spread over 19 of the 90 shape pairs. After the fix 10 points remain divergent, and all ten are a separate issue: `round($rMin + $guess * ($rMax - $rMin), 12)` on the return quantises quantiles smaller than 1e-12, so `BETAINV(0.001, 0.5, 20000)` yields `3.9e-11` against a true `3.9270419608e-11`. That is orthogonal to this change and I have left it alone.

### The degenerate-shape check

Removing the test also removes the only thing that stopped the search when `incompleteBeta` declines to evaluate at all. Its docblock says it returns 0 once `p + q` exceeds `LOG_GAMMA_X_MAX_VALUE` (2.55e305) "to avoid errors and over/underflow" — so the CDF is identically zero there and `$result === 0.0` was ending the loop on every iteration. Without a check, the search would walk the bracket to 1.0 and report that; `inverse()` now rejects those shapes up front instead, since a CDF that is zero everywhere has no quantile to find. `BETAINV(0.5, 2e305, 1e305)` returns `#NUM!` where master returned `0.5`.

I checked the neighbouring degenerate regions rather than assuming: for `alpha + beta` between roughly 1e18 and 2.55e305 the CDF is NAN, every comparison against it is false, and the search behaves identically before and after this change (`BETAINV(0.5, 1e20, 1e20)` stays `0.5`, which is right). The one place the behaviour improves for free is skewed huge shapes where the CDF underflows without hitting the docblock limit: `BETAINV(0.5, 1e20, 3)` was `0.5` and is now `1.0`, matching `betaincinv`. Both are pinned by tests.

### The sibling root-finders

The same "guard that can only fire on the case it was not written for" shape appears three more times in `Statistical/Distributions/`, and I deliberately did not touch any of them:

`Beta::calculateInverse()`, `GammaBase::calculateInverse()` and `NewtonRaphson::execute()` all end with `if ($i === self::MAX_ITERATIONS) { return NA; }`. Both loop idioms in use (`++$i <= MAX` and `$i++ < MAX`) leave `$i` at 257 when the iteration budget runs out, because the increment still runs on the check that fails. So the guard never fires on exhaustion; the only way to reach it is to converge exactly on iteration 256.

For Beta that is inert. `$b - $a` halves unconditionally every iteration, so the loop always ends by iteration 52 (measured maximum 52 over a wide sweep) and 256 is unreachable in either direction. Correcting the comparison changes nothing observable, and phpstan flags the corrected form as always-false, so it is not in this PR.

For `GammaBase` the inversion is load-bearing today, which is why it needs its own change rather than a one-character fix. Instrumenting the loop shows it genuinely exhausts all 256 iterations on ordinary input — `GAMMA.INV(0.5, 100, 1)`, `GAMMA.INV(0.9, 5, 1)`, `GAMMA.INV(0.75, 5, 1)` and five other points in a modest sweep — because `abs($dx) > Functions::PRECISION` is an absolute tolerance on an unbounded Newton step, and the iteration limit-cycles in the last ulp. All eight of those return values are correct to within 3e-15 of `scipy.stats.gamma.ppf`. If the guard were simply corrected to fire on exhaustion, those calls would start returning `#N/A`. The convergence test has to be fixed first, and `GammaBase` already has #4953 open against it, so I have kept this PR to `Beta.php` and can follow up separately if you want that one addressed.

### Tests

`tests/data/Calculation/Statistical/BETAINV.php` gains thirteen rows: symmetric large shapes including the scaled `[1, 3]` path, the `beta = 1` closed form, the onset at `alpha = 1100`, both skew directions, and the three degenerate-shape cases above. `BetaInvTest` gains three parametrised checks over a shape grid that need no external reference — the symmetric median is exactly 0.5, `BETAINV(p, alpha, 1)` matches `p ** (1 / alpha)`, and `BETADIST(BETAINV(p, a, b), a, b)` returns `p` while `BETAINV(p, a, b) + BETAINV(1 - p, b, a)` returns 1 (that mirror identity came out exact across the grid I measured). 105 new cases, 56 of which fail on master.

Mutation-checked: restoring the original disjunct, sending the underflow to `$b = $guess` instead of `$a = $guess`, and dropping or widening the degenerate-shape threshold each redden the suite (55, 55, 1 and 1 failures). Two mutants survive and are equivalent by construction — writing `$a = $b` instead of `$b = $a` on the exact-hit branch, and dropping the exact-hit branch entirely, which is a shortcut rather than a decision.

Full suite is 22052 tests with the same 8 failures and 1 error as master (Reader/Writer Html and intl locales, unrelated to this); phpstan, phpcs and php-cs-fixer are clean against the baseline.


oleibman:

PR #4957, opened 2026-08-10
Optimize large Xlsx populate and write hot paths

JulienChavee:
This is:

- [ ] a bugfix
- [ ] a new feature
- [x] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

# Why this change is needed?

Large Xlsx generation spends a lot of time in cell population and sheet writing. A Blackfire profile of a ~167k-cell export showed heavy cost in `Cell::setValueExplicit` → `Worksheet::getStyle()` (quote-prefix handling on every value set) and repeated `getCell()` calls in `Writer\Xlsx\Worksheet::writeSheetData` (ignored-errors checks + write).

## Changes

1. **`Cell::setValueExplicit`** — read the cell’s existing quote-prefix from its xf via `getCellXfByIndex()`; call the expensive `getStyle()` / `setQuotePrefix()` path only when quote-prefix must change. Selection / active-sheet restore behavior is unchanged when that path runs.
2. **`Writer\Xlsx\Worksheet::writeSheetData`** — fetch each cell once, read ignored-errors from that instance, and pass the `Cell` into `writeCell()`. The existing comma-separated `$cellsByRow` index is kept (lower memory than an array-of-columns).

## Performance (library-only microbenchmark, PHP 8.3, opcache CLI off)

Dense string populate + Xlsx save (`setPreCalculateFormulas(false)`):

| Cells | Before | After | Delta |
|------:|-------:|------:|------:|
| 40 000 | ~764 ms | ~491 ms | **~36% faster** |
| 170 000 | ~3296 ms | ~2098 ms | **~36% faster** |

Isolated write-loop lookup pattern on 170 000 cells (7× `getCell`/cell → 1×): ~236 ms → ~98 ms.

## Tests

- New coverage for quote-prefix apply / clear / no-op paths in `setValueExplicit` (including selection / active sheet preservation when prefix changes).
- Existing Xlsx ignored-errors round-trip tests still pass.
- Optional benchmark: `vendor/bin/phpunit --testsuite Benchmark --filter LargeXlsxWriteBenchmark --stderr`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,7 +22,7 @@ Some earlier branches remain supported and security fixes are applied to them; i
 
 ### Changed
 
-- Nothing yet.
+- Performance: avoid `Worksheet::getStyle()` on every `Cell::setValueExplicit()` unless quote-prefix must change; reuse a single `getCell()` per cell in Xlsx `writeSheetData` (ignored errors + write). On a dense 40k-cell populate+save microbenchmark this cut wall time by ~36%.
 
 ### Moved
 
--- a/src/PhpSpreadsheet/Cell/Cell.php
+++ b/src/PhpSpreadsheet/Cell/Cell.php
@@ -370,16 +370,16 @@ class Cell implements Stringable
         $worksheet = $this->getWorksheet();
         $spreadsheet = $worksheet->getParent();
         if (isset($spreadsheet) && $spreadsheet->getIndex($worksheet, true) >= 0) {
-            $originalSelected = $worksheet->getSelectedCells();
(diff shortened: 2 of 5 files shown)


oleibman:
The `ignoreErrors` part of your change was merged with another PR within the last few days. Can you please fetch a fresh copy, then apply just your `quotePrefix` change along with benchmark data showing its effects?

JulienChavee:
Yes, just saw that.
I will fetch latest changes and re-apply the quotePrefix, then re-run the benchmark

PR #4958, opened 2026-08-10
Avoid getStyle on every setValueExplicit quote-prefix check

JulienChavee:
This is:

- [ ] a bug fix
- [ ] a new feature
- [x] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Large Xlsx generation spends a lot of time in cell population. A Blackfire profile of a ~167k-cell export showed heavy cost in `Cell::setValueExplicit` → `Worksheet::getStyle()` because quote-prefix was checked via the style supervisor on every value set, even when it did not change.

### Changes

- **`Cell::setValueExplicit`** — read the cell’s existing quote-prefix from its xf via `getCellXfByIndex()`; call the expensive `getStyle()` / `setQuotePrefix()` path only when quote-prefix must change. Selection / active-sheet restore behavior is unchanged when that path runs.

### Performance (against current master, PHP 8.3, opcache CLI off)

Dense string populate + Xlsx save (`setPreCalculateFormulas(false)`), 3-run average:

| Cells | Before | After | Delta |
|------:|-------:|------:|------:|
| 40 000 | ~703 ms | ~484 ms | **~31% faster** |
| 170 000 | ~3114 ms | ~2089 ms | **~33% faster** |

### Tests

- New coverage for quote-prefix apply / clear / no-op paths in `setValueExplicit` (including selection / active sheet preservation when prefix changes)
- Optional benchmark: `vendor/bin/phpunit --testsuite Benchmark --filter LargeXlsxWriteBenchmark --stderr`

### DIFF
--- a/tests/PhpSpreadsheetTests/Cell/SetValueExplicitQuotePrefixPerfTest.php
+++ b/tests/PhpSpreadsheetTests/Cell/SetValueExplicitQuotePrefixPerfTest.php
@@ -56,10 +56,13 @@ class SetValueExplicitQuotePrefixPerfTest extends TestCase
         $spreadsheet->disconnectWorksheets();
     }
 
-    public function testNumericValueDoesNotTouchQuotePrefix(): void
+    public function testQuotePrefixClearedWhenReplacingWithNumericValue(): void
     {
         $spreadsheet = new Spreadsheet();
         $sheet = $spreadsheet->getActiveSheet();
+        $sheet->getCell('A1')->setValueExplicit('=prefixed', DataType::TYPE_STRING);
+        self::assertTrue($spreadsheet->getCellXfByIndex($sheet->getCell('A1')->getXfIndex())->getQuotePrefix());
+
         $sheet->getCell('A1')->setValueExplicit(42, DataType::TYPE_NUMERIC);
 
         self::assertFalse($spreadsheet->getCellXfByIndex($sheet->getCell('A1')->getXfIndex())->getQuotePrefix());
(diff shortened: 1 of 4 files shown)


tests/PhpSpreadsheetTests/Cell/SetValueExplicitQuotePrefixPerfTest.php
```diff
@@ -0,0 +1,70 @@
...
+        $sheet->getCell('A1')->setValueExplicit('=prefixed', DataType::TYPE_STRING);
+        self::assertTrue($spreadsheet->getCellXfByIndex($sheet->getCell('A1')->getXfIndex())->getQuotePrefix());
+
+        $sheet->getCell('A1')->setValueExplicit('plain', DataType::TYPE_STRING);
+        self::assertFalse($spreadsheet->getCellXfByIndex($sheet->getCell('A1')->getXfIndex())->getQuotePrefix());
+
+        $spreadsheet->disconnectWorksheets();
+    }
+
+    public function testNumericValueDoesNotTouchQuotePrefix(): void
```
oleibman: I don't think this test is adequate, or is correctly named. If cell A1 was set to a value with style quotePrefix true, and you were then to set it to a numeric value, quotePrefix should indeed change to false. Which I think it does, but your test doesn't demonstrate that.

JulienChavee: Thanks for the feedback! Updated it, sounds the assertion sounds better?

oleibman:

PR #4962, opened 2026-08-15
Ods Writer: do not convert inside string literals

### LINKED ISSUE #4454: Ods Writer Adjusts Cell References Even Within Quoted Strings
```php
$sheet1->setCellValue('E1', '="THIS IS E1"');
```
Writer converts formula to `="THIS IS [.E1]"`. The substitution would be correct for, say, `=G1`. This seems like it could be very tricky to fix.

saifulferoz:
Fixes #4454

`convertFormula()` ran the cell reference, defined name and function name
conversions over the whole formula string, with nothing to tell it where a
quoted string starts. Anything inside one that looked like a reference was
rewritten:

```php
$sheet->setCellValue('E1', '="THIS IS E1"');
// written as of:="THIS IS [.E1]"
```

The `str_replace(',', ';', ...)` in `convertCellReferences()` has the same
blind spot, so the text of a string was corrupted too:

```php
$sheet->setCellValue('B1', '=IF(A1>1,"yes, really","no")');
// written as of:=IF([.A1]>1;"yes; really";"no")
```

This splits the formula on string literals and converts only the segments
outside of them, reusing `Calculation::CALCULATION_REGEXP_STRING` so that a
doubled quote keeps escaping a quote (`="a""b E1"` stays intact).

Splitting is safe for the conversion that walks offsets: a range cannot be
interrupted by a string literal — `A1:` and `:B2` can never end up in
different segments — so every reference still sees the `:` on either side of
it that `convertCellReferences()` checks for.

Three of the four added tests fail on master and pass here; the fourth covers
a range and passes either way, to catch a regression in the splitting.

Sheet-qualified references such as `='My Sheet'!A1` are unaffected, since
those are single-quoted and never match a string literal.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ Some earlier branches remain supported and security fixes are applied to them; i
 
 ### Fixed
 
+- Ods Writer no longer converts cell references or commas inside string literals, so `="THIS IS E1"` is written unchanged instead of as `="THIS IS [.E1]"`. [Issue #4454](https://github.com/PHPOffice/PhpSpreadsheet/issues/4454)
 - BETAINV/BETA.INV no longer abandons its search when the Beta CDF underflows to zero (wrong results for alpha above about 1080). [PR #4954](https://github.com/PHPOffice/PhpSpreadsheet/pull/4954)
 - GAMMA.INV, the GAMMA.DIST/CHISQ.DIST/F.DIST densities, and GAMMALN no longer fail or return wrong results for large shape parameters / degrees of freedom. [PR #4953](https://github.com/PHPOffice/PhpSpreadsheet/pull/4953)
 - GAMMAINV/GAMMA.INV no longer clamps upper-tail quantiles beyond alpha*beta*5. [PR #4946](https://github.com/PHPOffice/PhpSpreadsheet/pull/4946)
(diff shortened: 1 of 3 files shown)


oleibman:
This is very nice work. And, although your tests do prove that your code works, they are difficult to understand as-is. The test does not need to depend on the internal format of the Ods xml. What we really want to see is that writing and then reading the spreadsheet results in correct values for `getValue` and `getCalculatedValue`. For an example of how we do this, see `DurationTest` in the same directory (tests `getFormattedValue` rather than `getCalculatedValue`, but the principle is the same). (Admittedly, we probably have tests that look like yours does now, but ...)

saifulferoz:
Thanks — that is a much better way to write it, and the failure output is far more useful too. Reverting the fix now gives:

```
-'="THIS IS E1"'
+'="THIS IS [.E1]"'
```

which is the thing a user would actually see, rather than an xml fragment they have to decode.

Rewritten to extend `AbstractFunctional` and assert `getValue()` and `getCalculatedValue()` after `writeAndReload()`. The four xml tests collapse into one round trip covering the same cases: a string that looks like a reference, a string next to a reference that does need converting, a comma inside a string, a doubled quote escaping a quote, and a range as the control.

PR #4966, opened 2026-08-18
Add streaming (constant-memory) Xlsx writer

kemo:
This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

PhpSpreadsheet holds every cell as an object (~1KB+ each), so writing large files exhausts memory and users switch to OpenSpout for big exports. The PSR-16 cell cache reduces resident memory but keeps O(n) storage and is slow; `setUseDiskCaching()` only helps the writer-side XML buffers. There was no way to write a large Xlsx file with memory independent of row count.

This PR adds an **append-only streaming Xlsx writer** with constant memory in row count, as an additive feature:

```php
use PhpOffice\PhpSpreadsheet\Writer\Xlsx\Streaming\StreamingWriter;

$writer = new StreamingWriter('big.xlsx');
$bold = $writer->registerStyle(['font' => ['bold' => true]]);
$sheet = $writer->startSheet('Data');
$sheet->appendRow(['Name', 'Amount', 'Date'], $bold);
$sheet->appendRow(['Widget', 12.5, new DateTimeImmutable('2026-01-01')]);
$writer->close();
```

### How it works

- `appendRow()` serializes each row immediately to a `php://temp/maxmemory` stream via `XMLWriter`; strings are written as inline strings, so no shared-strings table accumulates.
- All workbook-level parts (ContentTypes, rels, docProps, theme, styles, workbook) are produced by the **existing** `Writer\Xlsx` part writers running against an internal shell `Spreadsheet` that carries only sheet names and registered styles — no duplicated XML logic, and **no changes to any existing class**.
- `close()` streams each finished sheet into the archive with `ZipStream::addFileFromStream()`.

### Benchmark (200,000 rows × 8 mixed columns, medians of 3 isolated processes)

ZipStream 3.1.1, Apple Silicon (Darwin 23.1.0 arm64).

**PHP 8.2.27**

| Engine | Wall time | Peak memory | File size |
| --- | --- | --- | --- |
| Standard `Spreadsheet` + `Writer\Xlsx` | 31.43 s | 1,091.06 MB | 9,173,721 bytes |
| `StreamingWriter` | **5.09 s** | **42.27 MB** | 8,118,639 bytes |
| Ratio | ~6.2x faster | ~25.8x less memory | ~1.13x smaller |

**PHP 8.5.10**

| Engine | Wall time | Peak memory | File size |
| --- | --- | --- | --- |
| Standard `Spreadsheet` + `Writer\Xlsx` | 33.83 s | 1,093.06 MB | 9,173,722 bytes |
| `StreamingWriter` | **5.28 s** | **42.27 MB** | 8,118,639 bytes |
| Ratio | ~6.4x faster | ~25.9x less memory | ~1.13x smaller |

Peak memory is flat past ~100k rows: 100k and 200k rows both peak at ~42 MB inside the writer (most of that is ZipStream's fixed 16 MB read block, not row data). The benchmark script is committed under `tests/Benchmark/` (plain PHP, outside the phpunit test path) with the recorded results in its README, so the numbers are reproducible: `php -d memory_limit=4G tests/Benchmark/bench.php <standard|streaming> [rows]`.

### Scope

Supported: scalars, formulas (written without cached values, `fullCalcOnLoad` set), `DateTimeInterface` (default date format unless styled), per-row and per-cell styles via `registerStyle()`/`StreamedCell`, column widths, freeze pane, autofilter over the written range, multiple sequential sheets.

Not supported (documented): random access, reading back, merged cells, charts, drawings, comments, conditional formatting, hyperlinks, rich text. One caveat is documented explicitly: PhpSpreadsheet's own reader returns inline-string cells as `RichText` objects; Excel and other consumers read them as plain strings.

### Testing

- 30 unit tests / 80 assertions: round-trips through the real `Reader\Xlsx` for every value type, style, and sheet feature; lifecycle guard tests (append after close, stale sheets, invalid widths, unsupported forced types, failure cleanup); a corrupted-sheet guard (a failed `appendRow()` poisons the sheet so `close()` throws instead of writing malformed XML); and a memory test asserting the peak at 200k rows does not exceed the peak at 100k rows.
- Full suite, PHPStan (level 10, no new errors), PHP-CS-Fixer, and PHPCS are clean.
- New docs page: `docs/topics/streaming-writer.md`.


src/PhpSpreadsheet/Writer/Xlsx/Streaming/StreamingWriter.php
```diff
@@ -0,0 +1,205 @@
...
+    public function isStyleIdRegistered(int $styleId): bool
+    {
+        return $styleId >= 0 && $styleId < count($this->shell->getCellXfCollection());
+    }
+
+    public function getDefaultDateStyleId(): int
+    {
+        if ($this->defaultDateStyleId === null) {
+            $this->defaultDateStyleId = $this->registerStyle([
+                'numberFormat' => ['formatCode' => NumberFormat::FORMAT_DATE_DATETIME],
```
oleibman: I will not accept an ambiguous default for date/time number format. Please use `FORMAT_DATE_DATETIME_BETTER` or, if you don't like that, any other format which includes a 4-digit year and no ambiguity between month and day.

docs/topics/streaming-writer.md
```diff
@@ -0,0 +1,233 @@
...
+Because the workbook contains a formula, `close()` marks the workbook for
+full calculation on load, so Excel (or another spreadsheet application)
+computes `=SUM(C2:C2)` when it opens the file. The written cell itself has
+no cached value. If no sheet in the workbook contains a formula, `close()`
+does not force a recalculation on load.
+
+Any `DateTimeInterface` value that has no explicit style gets a default
+date number format automatically.
+
+`freezePane('A1')` is a no-op; freezing at the top-left cell freezes
```
oleibman: I do and do not understand this comment. I agree that it freezes nothing, but what is your point? Is it that you expect freeze to always be in effect, but this is a way to circumvent that?

docs/topics/streaming-writer.md
```diff
@@ -0,0 +1,233 @@
...
+// Start a second sheet; the first sheet is finished and can no longer
+// be appended to once startSheet() runs again.
+$sheet2 = $writer->startSheet('Notes');
+$sheet2->appendRow(['Generated by the streaming writer.']);
+
+$writer->close();
+```
+
+Because the workbook contains a formula, `close()` marks the workbook for
+full calculation on load, so Excel (or another spreadsheet application)
```
oleibman, line 95: I do not see where `close` marks the workbook for full calculation. Please enlighten me.

src/PhpSpreadsheet/Writer/Xlsx/Streaming/StreamingWriter.php
```diff
@@ -0,0 +1,205 @@
...
+    private bool $hasFormulas = false;
+
+    private ?int $defaultDateStyleId = null;
+
+    public function __construct(string $filename)
+    {
+        try {
+            $fileHandle = fopen($filename, 'wb');
+        } catch (Exception) {
+            throw new WriterException("Could not open file $filename for writing.");
```
oleibman: You could just set $fileHandle to false here, eliminating the need to code the `throw` twice, while simultaneously dealing with your only uncovered statement.

docs/topics/streaming-writer.md
kemo: Comment* updated 729d1b03fc28746bc0329c8763075ccee2c5aca0

docs/topics/streaming-writer.md
kemo: I reworded the doc to say why the note is there.

src/PhpSpreadsheet/Writer/Xlsx/Streaming/StreamingWriter.php
kemo: Updated to use `FORMAT_DATE_DATETIME_BETTER` ( 2badc79ac9b9341c6bc146b3b2a5faf1b4a23c2c )

docs/topics/streaming-writer.md
oleibman: I still don't see where that attribute is set in your code. Please explain.

More to the point, 'forceFullCalc` is a very dangerous option. Please see PR #4271 for details on its unexpected side effects, and why we would not want to make it an "automatic" choice.

PR #4968, opened 2026-08-19
Xlsx writer: a header-row table takes the empty row below it

dkulyk:
This is:

- [x] a bugfix
- [ ] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

A table whose range covers its header row alone is written out as-is today, and Excel refuses the
workbook: "We found a problem with some content in ...". Repairing drops the table, so the markup is
lost without a word — and with it the only thing that tells a screen reader which row is the header.

```php
$sheet->fromArray([['Name', 'Owner']], null, 'A1');   // headings, no data rows yet
$sheet->addTable((new Table('A1:B1', 'Risks'))->setShowHeaderRow(true));
// -> written as ref="A1:B1", and Excel reports the file as unreadable
```

Measured in Excel 16.112.1 on macOS: a workbook with `ref="A1:B1"` and `headerRowCount="1"` errors on
open; the same workbook with `ref="A1:B2"` opens clean, even though `sheetData` holds no cell for row 2.

Excel itself never writes the first shape. Asked to insert a table over a single row of headings, it
writes `ref="A1:B2"` while `sheetData` still holds row 1 only and `dimension` stays `A1:B1` — the
table reaches over an empty row, and that is what makes the file readable.

So the Xlsx writer now does the same: when a table shows a header row and its range is one row tall,
the written `ref` (and the autofilter's) extends over the row below. Nothing is added to `sheetData`,
and the `Table` object the caller holds is untouched — only what is written changes.

The row below is taken only when it holds nothing. A table silently swallowing a row that belongs to
something else would change what the sheet says, so that case throws instead, naming the row.

This is a writer-side fix rather than a validation in `Worksheet\Table`, because the constructor
validates the range before `setShowHeaderRow()` can be called, and a single-row table without a
header row is legitimate — only the writer knows both facts at once.

Four unit tests in `tests/PhpSpreadsheetTests/Writer/Xlsx/TableHeaderRowTest.php` cover the four
cases: the empty row below is taken, a non-empty row below is refused, an ordinary table with data
is untouched, and a single-row table without a header row is written as given.

### DIFF
--- a/tests/PhpSpreadsheetTests/Writer/Xlsx/TableHeaderRowTest.php
+++ b/tests/PhpSpreadsheetTests/Writer/Xlsx/TableHeaderRowTest.php
@@ -1,0 +1,96 @@
+<?php
+
+declare(strict_types=1);
+
+namespace PhpOffice\PhpSpreadsheetTests\Writer\Xlsx;
+
+use PhpOffice\PhpSpreadsheet\Exception as PhpSpreadsheetException;
+use PhpOffice\PhpSpreadsheet\Spreadsheet;
+use PhpOffice\PhpSpreadsheet\Worksheet\Table;
+use PhpOffice\PhpSpreadsheetTests\Functional\AbstractFunctional;
+
+/**
+ * Excel reports a workbook whose table covers its header row alone as unreadable, and repairs it
+ * by dropping the table. Excel itself never writes one: asked to make a table over a single row of
+ * headings it writes the table over the row below as well, leaving that row without a cell. The
+ * writer does the same — unless that row already holds something.
+ */
+class TableHeaderRowTest extends AbstractFunctional
+{
+    private ?Spreadsheet $spreadsheet = null;
+
+    protected function tearDown(): void
+    {
+        if ($this->spreadsheet !== null) {
+            $this->spreadsheet->disconnectWorksheets();
+            $this->spreadsheet = null;
+        }
+    }
+
+    public function testHeaderRowTakesTheEmptyRowBelowIt(): void
+    {
+        $this->spreadsheet = new Spreadsheet();
+        $sheet = $this->spreadsheet->getActiveSheet();
+        $sheet->fromArray([['Year', 'Country']], null, 'A1');
+        $sheet->addTable(new Table('A1:B1', 'SalesData'));
+
+        $reloaded = $this->writeAndReload($this->spreadsheet, 'Xlsx');
+        $worksheet = $reloaded->getActiveSheet();
+        $table = $worksheet->getTableByName('SalesData');
+
+        // exactly what Excel writes for a table made over a single row of headings
+        self::assertInstanceOf(Table::class, $table);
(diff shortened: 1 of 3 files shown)


tests/PhpSpreadsheetTests/Writer/Xlsx/TableHeaderRowTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $sheet->getCell('A2')->setValue('Total');
+        $sheet->addTable(new Table('A1:B1', 'SalesData'));
+
+        $this->expectException(PhpSpreadsheetException::class);
+        $this->expectExceptionMessage('needs at least 2 rows');
+
+        $this->writeAndReload($this->spreadsheet, 'Xlsx')->disconnectWorksheets();
+    }
+
+    public function testHeaderRowWithOneRowOfDataIsWritten(): void
```
oleibman, line 63: This test does not appear to have changed behavior with your change. That does not mean it's not a valuable test; I just want to confirm that this is the intended result.

tests/PhpSpreadsheetTests/Writer/Xlsx/TableHeaderRowTest.php
```diff
@@ -0,0 +1,96 @@
...
+        $table = $reloaded->getActiveSheet()->getTableByName('SalesData');
+
+        self::assertInstanceOf(Table::class, $table);
+        self::assertSame('A1:B2', $table->getRange());
+        self::assertTrue($table->getShowHeaderRow());
+
+        $reloaded->disconnectWorksheets();
+    }
+
+    public function testASingleRowTableWithoutAHeaderRowIsWritten(): void
```
oleibman, line 80: This test does not appear to have changed behavior with your change. That does not mean it's not a valuable test; I just want to confirm that this is the intended result.

tests/PhpSpreadsheetTests/Writer/Xlsx/TableHeaderRowTest.php
dkulyk: Yes, intended. Both tests sit on the edges of the change rather than inside it: the first has two rows, so `$rowCount > 1` and the writer leaves it alone; the second has `showHeaderRow = false`, so the condition doesn't apply either. They're there to fail if the condition is ever widened and starts reaching for the row below where it shouldn't.

tests/PhpSpreadsheetTests/Writer/Xlsx/TableHeaderRowTest.php
dkulyk: Yes, intended. Both tests sit on the edges of the change rather than inside it: the first has two rows, so `$rowCount > 1` and the writer leaves it alone; the second has `showHeaderRow = false`, so the condition doesn't apply either. They're there to fail if the condition is ever widened and starts reaching for the row below where it shouldn't.

oleibman:

PR #4974, opened 2026-08-24
Add Agile XLSX password-to-open encryption support

n3crosis:
This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Changes are covered by unit tests
  - [ ] Changes are covered by existing unit tests
  - [x] New unit tests have been added
- [x] Code style is respected
- [x] Commit message explains **why** the change is made (see https://github.com/erlang/otp/wiki/Writing-good-commit-messages)
- [x] CHANGELOG.md contains a short summary of the change and a link to the pull request if applicable
- [x] Documentation is updated as necessary

### Why this change is needed?

Adds programmatic XLSX password-to-open encryption and decryption using the
Office Agile encryption format. This fulfils the request in historical issue
[#3878](https://github.com/PHPOffice/PhpSpreadsheet/issues/3878), which was
closed as not planned.

The feature is deliberately limited to XLSX password-to-open encryption.
Password-to-modify, certificate-only encryption, custom providers, legacy XLS,
ODS encryption, and unsupported cipher/hash modes remain out of scope.

### What changed?

- `Reader\Xlsx` detects and decrypts supported encrypted workbooks after
  `setEncryptionPassword()` is called. `listWorksheetNames()` and
  `listWorksheetInfo()` support them as well as `load()`.
- `Writer\Xlsx` encrypts workbooks after `setEncryptionPassword()` is called.
  It defaults to AES-256/CBC, SHA-512, and 100,000 spin iterations;
  AES-128/AES-192 and SHA-1/SHA-2 profiles can be selected with
  `setEncryptionProfile()`.
- Adds `Shared\Xlsx\AgileEncryption` for Agile metadata parsing, password
  verification, integrity checks, encryption/decryption, and CFB container
  serialization.
- Extends the existing OLE reader with named-stream access and correct
  mini-stream traversal. No new OLE/CFB dependency is introduced.
- Writes ordered, red-black CFB directory trees so consumers performing
  directory name lookups can locate the Data Spaces streams.
- Declares the required `ext-openssl` PHP extension.

### Specification basis

The implementation follows the specified Agile password-encryption and Data
Spaces container structures for the supported AES/CBC profiles. In particular,
it serializes the encrypted package as an OLE Compound File with the required
Data Spaces streams and an ordered CFB directory tree; it does not claim to
implement every encryption mode described by the specifications.

[...]

Issue #3878, opened 2024-01-26
How to apply file encryption password?


rodrigoguariento:
This is:

```
- [ ] a bug report
- [X] a feature request
- [ ] **not** a usage question (ask them on https://stackoverflow.com/questions/tagged/phpspreadsheet or https://gitter.im/PHPOffice/PhpSpreadsheet)
```

### What is the expected behavior?
Apply file encryption password programmatically, namely, the password is required to open the file in Excel.

### What is the current behavior?
I haven't found documentation explaining this: https://phpspreadsheet.readthedocs.io/en/latest/topics/recipes/#setting-security-on-a-spreadsheet

### What are the steps to reproduce?

N/A

### What features do you think are causing the issue

- [ ] Reader
- [X] Writer
- [ ] Styles
- [ ] Data Validations
- [ ] Formula Calculations
- [ ] Charts
- [ ] AutoFilter
- [ ] Form Elements

### Does an issue affect all spreadsheet file formats? If not, which formats are affected?
XLSX

### Which versions of PhpSpreadsheet and PHP are affected?
phpoffice/phpspreadsheet: ^1.29

MarkBaker:
This feature isn't supported, and unless you can find details online of how Microsoft implements their file encryption then it's not practical.

rodrigoguariento:
Alternatives? Suggestion? Someone?

oleibman:
PR #4975 implements encryption support. Changing close reason from "not planned" to "completed".

PR #4975, opened 2026-08-25
Add Agile XLSX file encryption support

n3crosis:
This is:

- [ ] a bugfix
- [x] a new feature
- [ ] refactoring
- [ ] additional unit tests

Checklist:

- [x] Code style is respected

Adds programmatic XLSX password-to-open encryption and decryption using the
Office Agile encryption format. This fulfils the request in historical issue
[#3878](https://github.com/PHPOffice/PhpSpreadsheet/issues/3878), which was
closed as not planned.

This PR replaces [#4974](https://github.com/PHPOffice/PhpSpreadsheet/pull/4974),
which I accidentally closed while rewriting Git history and force-pushing the
branch. I apologize for the disruption.

The feature is deliberately limited to XLSX password-to-open encryption using
the Agile profile. Standard Encryption (`EncryptionInfo` version 4.2),
password-to-modify, certificate-only encryption, custom providers, legacy XLS,
ODS encryption, and unsupported cipher/hash modes remain out of scope. Standard
Encryption workbooks are rejected with an unsupported-profile error rather than
being interpreted as Agile encryption.

### What changed?

- `Reader\Xlsx` detects and decrypts supported encrypted workbooks after
  `setEncryptionPassword()` is called. `listWorksheetNames()` and
  `listWorksheetInfo()` support them as well as `load()`.
- `Writer\Xlsx` encrypts workbooks after `setEncryptionPassword()` is called.
  It defaults to AES-256/CBC, SHA-512, and 100,000 spin iterations;
  AES-128/AES-192 and SHA1/SHA-2 profiles can be selected with
  `setEncryptionProfile()`.
- Adds `Shared\Xlsx\AgileEncryption` for Agile metadata parsing, password
  verification, integrity checks, encryption/decryption, and CFB container
  serialization.
- Extends the existing OLE reader with named small-stream access: it follows
  the root mini-stream allocation chain, uses mini-sector offsets, and rejects
  malformed or cyclic chains. This is required to read `EncryptionInfo` from
  encrypted XLSX files without a new OLE/CFB dependency.
- Writes ordered, red-black CFB directory trees so consumers performing
  directory name lookups can locate the Data Spaces streams.
- Declares the required `ext-openssl` PHP extension.

### Specification basis

The implementation follows the specified Agile password-encryption and Data
Spaces container structures for the supported AES/CBC profiles. In particular,
it serializes the encrypted package as an OLE Compound File with the required
Data Spaces streams and an ordered CFB directory tree; it does not claim to
implement every encryption mode described by the specifications.

- [ECMA-376: Office Open XML File Formats](https://ecma-international.org/publications-and-standards/standards/ecma-376/)
  — OPC package structure and encrypted-package requirements.
- [MS-OFFCRYPTO: Agile Encryption](https://learn.microsoft.com/en-us/openspecs/office_file_formats/ms-offcrypto/cab78f5c-9c17-495e-bea9-032c63f02ad8)
  — password derivation, AES/CBC encryption, verifier, and integrity data.
- [MS-CFB: Compound File Binary File Format](https://learn.microsoft.com/en-us/openspecs/windows_protocols/ms-cfb/53989ce4-7b05-4f8d-829b-d08d6148375b)
  — directory entries, red-black name trees, FAT/DIFAT, and mini-streams.

### Documentation

- Adds `docs/topics/encryption.md`, including reader/writer examples, supported
  profiles, limitations, and format/specification references.
- Updates the encryption feature cross-reference and the existing protection
  recipe.

### Tests and validation

- Sanitized Excel-generated Agile-encrypted XLSX reader fixture (password:
  `open`); its creator and last-modified-by properties are blank.
- Wrong-password, tampered-package, and unsupported-profile rejection tests.
- Reader metadata API coverage for encrypted workbooks.
- Writer profile, encrypted round-trip, CFB directory-tree, FAT, and DIFAT
  coverage.
- Existing OLE coverage for named encrypted streams.

Validation run:

```text
PHPUnit: 32 tests, 119 assertions — passed
Focused PHPStan — passed
PHP-CS-Fixer (sequential dry run) — passed
git diff --check — passed
```

### Manual Microsoft Excel validation

Run this from the repository root, then open
`agile-encrypted-maintainer-test.xlsx` in Microsoft Excel using the password
`writer-password`:

```sh
php -r '
require "vendor/autoload.php";

$spreadsheet->getActiveSheet()->setCellValue("A1", "PhpSpreadsheet Agile encryption test");
$spreadsheet->getActiveSheet()->setCellValue("B1", 42);

(new \PhpOffice\PhpSpreadsheet\Writer\Xlsx($spreadsheet))
    ->setEncryptionPassword("writer-password")
    ->save("agile-encrypted-maintainer-test.xlsx");
'
```


oleibman:
PR #4982 should fix the problems associated with running Phpstan on an earlier Php version that you reported in the previous incarnation of this ticket.

src/PhpSpreadsheet/Shared/Xlsx/AgileEncryption.php
```diff
@@ -0,0 +1,740 @@
...
+        return pack('V2', $size % 4294967296, intdiv($size, 4294967296));
+    }
+
+    private static function unpackSize(string $data): int
+    {
+        $unpackedSize = unpack('Vlow/Vhigh', $data);
+        if ($unpackedSize === false || !isset($unpackedSize['low'], $unpackedSize['high']) || !is_int($unpackedSize['low']) || !is_int($unpackedSize['high'])) {
+            throw new Exception('Malformed encrypted XLSX package.');
+        }
+        $size = $unpackedSize['low'] + $unpackedSize['high'] * 4294967296;
```
dkulyk: `$unpackedSize['high'] * 4294967296` overflows to float once `high >= 0x80000000`, and the guard below then compares two floats. For a size field of exactly 2^63 (`00 00 00 00 00 00 00 80`) the comparison is false, because `(float) PHP_INT_MAX` is 2^63 as well, so the guard passes and the function returns a float from an `: int` signature:

```
TypeError: unpackSize(): Return value must be of type int, float returned
```

A malformed file on the decryption path then raises a `TypeError` instead of `Exception`. `if (!is_int($size))` covers every overflow case including the boundary.

dkulyk (commented):
Disclosure first: I maintain `dkulyk/openxml`, which implements the same MS-OFFCRYPTO profiles, so I have an interest here. I am not proposing it -- keeping this in-tree with no new dependency is the right call for PhpSpreadsheet. Two findings from reading the crypto path, one of them a bug.

The rest holds up: `hash_equals` on both the password verifier and the HMAC, integrity checked before the package is decrypted rather than after, per-segment IVs from the block number, `OPENSSL_ZERO_PADDING`, strict length validation of every `EncryptionInfo` field, and a bounded `spinCount`. Standard Encryption is rejected with an unsupported-profile error rather than misread as Agile, and the limitations listed in the description match what the code actually does.

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -109,6 +120,60 @@ public function canRead(string $filename): bool
...
+        try {
+            $ole = new OLE();
+            $ole->read($filename);
+            $encryptionInfo = $ole->getDataByName('EncryptionInfo');
+            $encryptedPackage = $ole->getDataByName('EncryptedPackage');
+        } catch (Throwable) {
+            return null;
+        }
+
+        $plaintext = AgileEncryption::decrypt(AgileEncryption::parse($encryptionInfo), $encryptedPackage, $this->encryptionPassword);
```
dkulyk: The write path has `encryptFile()` and streams. This one holds the whole ciphertext returned by `getDataByName('EncryptedPackage')`, then the whole plaintext returned by `decrypt()`, before writing it out -- peak memory around two to three times the workbook. `decrypt()` already loops over 4096-byte segments, so a `decryptToFile()` counterpart writing each segment as it goes would be a small change rather than a rewrite.

src/PhpSpreadsheet/Shared/Xlsx/AgileEncryption.php
n3crosis: Fixed in 44a82a7ab: unpackSize() now rejects unrepresentable values before arithmetic on both 32-bit and 64-bit platforms, with regression coverage for each boundary.

src/PhpSpreadsheet/Reader/Xlsx.php
n3crosis: Fixed in 44a82a7ab (with additional coverage in 08e59dcc6): EncryptedPackage is copied and decrypted file-to-file, avoiding full ciphertext/plaintext retention in memory; both large and mini-stream paths are covered.

oleibman:
In https://github.com/PHPOffice/PhpSpreadsheet/pull/4974#issuecomment-5407306664, you indicated that you had added some tests to approve coverage. At least some of those appear to be missing now, e.g. passing a null-string password to decrypt. As I mentioned then, I don't need you to figure out a coverage test in all, or even most, situations (e.g. fread failing after fopen succeeds), but, if you could take care of the low-hanging fruit ...

n3crosis:
Thanks for pointing this out. I’ve added coverage for what I believe are most of the useful and practical remaining cases, including missing or incorrect passwords, Unicode passwords, additional encryption profiles, malformed keys, invalid OLE block chains, failed-save recovery, and temporary-file cleanup.

The remaining uncovered paths appear to require fairly contrived conditions or extensive mocking, such as low-level read failures after a file has already been opened successfully. Given your earlier guidance, I think those would add more complexity than value.

Please let me know if there are any other specific low-hanging cases you think I’ve missed.

oleibman:
@n3crosis Thank you, I agree that what you have left uncovered would require very contrived conditions, and are not a cause for concern.

@dkulyk Thank you for your thorough analysis. Please confirm that the points you have raised are addressed by the latest changes.

src/PhpSpreadsheet/Shared/Xlsx/AgileEncryption.php
```diff
@@ -0,0 +1,838 @@
...
+    public static function decryptFile(array $info, string $inputFilename, string $outputFilename, string $password): void
+    {
+        if ($password === '') {
+            throw new Exception('XLSX encryption password required.');
+        }
+        $input = @fopen($inputFilename, 'rb');
+        if ($input === false) {
+            throw new Exception('Could not open XLSX package for decryption.');
+        }
+        $output = @fopen($outputFilename, 'wb');
```
dkulyk: `decryptFile()` opens the destination with `wb` before verifying the password and the encrypted-package HMAC. Although the Xlsx reader currently passes a fresh temporary filename, this public method will truncate an existing destination even when authentication subsequently fails because of an incorrect password or corrupted input.

Could the destination be opened only after authentication succeeds, or could decryption write to a temporary file and replace the destination after the complete operation succeeds? A regression test with a pre-existing destination and an invalid password would make the intended behavior explicit.

dkulyk (commented):
Reviewed the file-backed decryption path; one destination-safety concern remains.

src/PhpSpreadsheet/Reader/Xlsx.php
```diff
@@ -109,6 +120,77 @@ public function canRead(string $filename): bool
...
+        }
+    }
+
+    private function hasEncryptedPackage(string $filename): bool
+    {
+        try {
+            $ole = new OLE();
+            $ole->read($filename);
+            $ole->getDataByName('EncryptionInfo');
+            $ole->getDataByName('EncryptedPackage');
```
dkulyk: `canRead()` reaches this method, which calls `getDataByName('EncryptedPackage')` and therefore loads the complete encrypted workbook payload into memory merely to check whether the stream exists. This negates the new file-backed processing path and can cause excessive memory consumption during format detection for large or untrusted CFB files.

Could OLE expose a lightweight `hasDataByName()` or equivalent directory-entry lookup so this check does not read the stream contents?

src/PhpSpreadsheet/Shared/Xlsx/AgileEncryption.php
```diff
@@ -0,0 +1,838 @@
...
+        $encryptedKeyBlockSize = self::decimalAttribute($encryptedKey, 'blockSize');
+        $encryptedKeyBits = self::decimalAttribute($encryptedKey, 'keyBits');
+        $encryptedKeyHashSize = self::decimalAttribute($encryptedKey, 'hashSize');
+        if (
+            (string) $keyData['cipherAlgorithm'] !== 'AES' || (string) $keyData['cipherChaining'] !== 'ChainingModeCBC'
+            || !self::isSupportedProfile($keyBits, $hashAlgorithm, $hashSize) || $keyDataSaltSize !== 16 || $keyDataBlockSize !== 16
+            || (string) $encryptedKey['cipherAlgorithm'] !== 'AES'
+            || (string) $encryptedKey['cipherChaining'] !== 'ChainingModeCBC'
+            || (string) $encryptedKey['hashAlgorithm'] !== $hashAlgorithm || $encryptedKeySaltSize !== 16 || $encryptedKeyBlockSize !== 16 || $encryptedKeyBits !== $keyBits
+            || $encryptedKeyHashSize !== $hashSize || $spinCount > 10000000
```
dkulyk: `spinCount` is controlled by the input file, and accepting up to 10,000,000 iterations can consume substantial CPU in PHP when processing untrusted uploads. A fixed upper bound prevents an unlimited attack, but it does not let server applications choose an appropriate work budget.

Could the reader expose a configurable maximum spin count, with the current value retained as a compatibility ceiling? This would let applications apply a stricter limit without rejecting encryption support entirely.

dkulyk (commented):
Two additional hardening observations for encrypted-file detection and untrusted password-hashing parameters.

src/PhpSpreadsheet/Shared/Xlsx/AgileEncryption.php
dkulyk: This is optional hardening rather than a merge blocker. The existing 10,000,000 ceiling already prevents an unbounded value; a configurable application-level limit could also be added in a follow-up if keeping this PR focused is preferable.

dkulyk:
One broader architectural question for the maintainers — @oleibman and @Progi1984, I would appreciate your views on this. This is not a blocker for the current PR.

OLE/CFB handling already exists in several variations across PHPExcel, PhpSpreadsheet, and PHPWord. If equivalent encryption support is added independently to PHPWord and PHPPresentation, we may end up duplicating the CFB container reader/writer, stream handling, validation, and security fixes, with each implementation maintained separately.

Would it make sense to treat the implementation introduced here as a proving ground, while considering convergence on a shared CFB component before equivalent encryption support is added to other PHPOffice packages?

This PR does not need to address that architectural question. It may simply be worth considering a shared direction before separate encryption/container implementations become established across the projects.

src/PhpSpreadsheet/Shared/Xlsx/AgileEncryption.php
n3crosis: Thanks for catching this. Fixed in 111c08884: decryptFile() now opens the destination only after both password  verification and encrypted-package HMAC validation succeed, so authentication failures no longer truncate an  existing file. I also added regression coverage for incorrect passwords and tampered packages preserving pre-existing destination content.

src/PhpSpreadsheet/Reader/Xlsx.php
n3crosis: Fixed in b3d9a8acc. I added OLE::hasDataByName() and updated canRead() to check the OLE directory entries without reading either stream payload. The actual load path continues to stream EncryptedPackage through a temporary file using copyDataByName(). Positive and negative stream-existence tests were added.

src/PhpSpreadsheet/Shared/Xlsx/AgileEncryption.php
n3crosis: Fixed in 4b1344335. The Xlsx reader now exposes setMaxEncryptionSpinCount() so applications can choose a stricter work limit, while retaining 10,000,000 as the default and absolute compatibility ceiling. The public parser also clamps caller-provided limits to that ceiling, so direct API use cannot bypass it. Tests cover lower-limit rejection, exact-boundary acceptance, invalid configured limits, and direct-parser bypass attempts.

dkulyk:
I reviewed the latest changes. All my findings have been addressed, and I have no remaining concerns or blockers. Thank you.

oleibman:
@n3crosis This is getting very close now. Can you please change Writer AgileEncryptionTest to replace `throw new RuntimeException` with `self::assertNotFalse`; the net effect is the same, but it seems more consistent with how Phpunit is supposed to work.

oleibman:
@dkulyk I don't object to a shared version, but there is some history here. There is a package `phpoffice/common` which I believe was supposed to provide common services across the PhpOffice suite. However, it had fallen by the wayside before I became involved - neither PhpSpreadsheet nor PhpWord lists it as a dependency, so it seems that it wasn't serving its purpose. PhpPresentation still uses it, but I am completely unfamiliar with that project. PhpExcel is abandoned, replaced by PhpSpreadsheet, so it wouldn't be involved even if we did move forward with what you are suggesting.

dkulyk:
I have a few ideas for this, but the approach is a different from using the common package. I'll try to run it on PHPPresentation and get everything aligned with @Progi1984  to start.

n3crosis:
I’ve updated Writer AgileEncryptionTest to use PHPUnit assertions instead of throwing RuntimeException, while preserving the existing validation. All checks passes now.

oleibman:
@n3crosis Thank you for your contribution. @dkulyk Thank you for your review.

PR #4997, opened 2026-09-18
Fix VALUE() treating product text as a date

### LINKED ISSUE #4996: VALUE() parses product text like "5V 2.1A - EU Wall Adaptor" as a date
## Which version of PhpSpreadsheet are you using?

Reproduced on 1.21 and against current `master` (`DateValue::fromString()` / `Format::VALUE()`).

## What is the expected vs actual behaviour?

Excel:

```excel
=VALUE("5V 2.1A - EU Wall Adaptor")  →  #VALUE!
=IFERROR(VALUE("5V 2.1A - EU Wall Adaptor"), 1)  →  1
```

PhpSpreadsheet:

```php
Format::VALUE('5V 2.1A - EU Wall Adaptor');
// actual: 37381.0  (Excel serial for 2002-05-05)
// expected: #VALUE!
```

This shows up in real workbooks that do `VALUE(LEFT(spec, FIND(" pcs", spec & " pcs")-1))` on a product specification. When the spec is not `"N pcs ..."`, Excel falls back through `IFERROR(..., 1)`. PhpSpreadsheet instead treats the spec as a date, so unit price becomes `amount / 37381` and quantity becomes `qty * 37381`.

A nearby description without a decimal point is already rejected:

```text
"5V 1A - EU Wall Adaptor for VC4/..."  →  #VALUE!   (correct)
"5V 2.1A - EU Wall Adaptor"            →  37381     (wrong)
```

## What are the steps to reproduce?

```php
use PhpOffice\PhpSpreadsheet\Calculation\DateTimeExcel\DateValue;
use PhpOffice\PhpSpreadsheet\Calculation\TextData\Format;

var_dump(Format::VALUE('5V 2.1A - EU Wall Adaptor'));
var_dump(DateValue::fromString('5V 2.1A - EU Wall Adaptor'));
```

`DateValue` replaces `.` and `-` with spaces, then `date_parse('5V-2-1A')` returns `error_count === 0` (year 2002, month 5, day 5). Excel `VALUE()` does not accept that string.

## Suggested fix

Reject strings that still contain leftover letters after stripping English month/weekday names, `am`/`pm`, digits, ordinals, and date/time separators, before calling `DateValue::fromString()` from `VALUE()`. Real dates such as `2026-09-18`, `18 Sep 2026`, `Sep 18, 2026`, and `6:45 PM` still parse.

gechuanpan:
## Summary

`VALUE()` currently forwards arbitrary non-numeric text to `DateValue::fromString()`. PHP's date parser then accepts strings such as `5V 2.1A - EU Wall Adaptor` (`date_parse('5V-2-1A')`), while Excel returns `#VALUE!`.

Guard `VALUE()` with a syntax check: after removing English month/weekday names, `am`/`pm`, digits, ordinals (`st`/`nd`/`rd`/`th`), and date/time punctuation, any leftover characters mean the text is not a date/time.

This keeps the existing `str_contains($value, ':')` branch for times. It does not change `DATEVALUE()` itself.

## Test plan

- [x] `VALUE("5V 2.1A - EU Wall Adaptor")` returns `#VALUE!`
- [x] `VALUE("13 Monkeys")` still returns `#VALUE!`
- [x] `VALUE("1-Jan-2014")` still returns `41640`
- [x] `VALUE("18 Sep 2026")` / `VALUE("Sep 18, 2026")` still parse
- [x] `VALUE("6:45 PM")` / `VALUE("12:34:56")` still parse
- [x] `composer check`

Fixes #4996


src/PhpSpreadsheet/Calculation/TextData/Format.php
```diff
@@ -229,6 +234,33 @@ public static function VALUE(mixed $value = '')
...
+    private static function isDateTimeString(string $value): bool
+    {
+        $text = strtolower(trim($value, "\" \t\n\r\0\x0B"));
+        if ($text === '' || strpbrk($text, '0123456789') === false) {
+            return false;
+        }
+
+        // Longer names first so "september" is not reduced to "ember".
+        $text = str_replace([
+            'september', 'october', 'november', 'december', 'january', 'february',
```
oleibman, line 250: Please rearrange these in a logical order (full month names, in calendar order, day names in calendar order, anything left over) so that I can review it properly.

src/PhpSpreadsheet/Calculation/TextData/Format.php
oleibman, line 229: Probably because of your change, this statement is no longer covered. Can you add a test case which covers it? Alternatively, it looks like the if statement preceding this is superfluous, and you can delete the if-block and just change the next statement to return `$dateValue` rather than `ExcelError::VALUE()`.

src/PhpSpreadsheet/Calculation/TextData/Format.php
gechuanpan: Done. The list is now full month names in calendar order, then full weekday names Sunday–Saturday, then leftover abbreviations (still longest-first so sept/tues/thurs are not shortened too early) plus utc/gmt/am/pm.

src/PhpSpreadsheet/Calculation/TextData/Format.php
gechuanpan: I took the second option: after isDateTimeString(), DateValue already returns #VALUE! on failure, so the extra if and trailing return ExcelError::VALUE() were redundant. It now always restores the date type and return $dateValue.

oleibman (approved):

oleibman:

